PR #96, opened 2026-05-10
feat(frontend): bridge generated-token logprobs through vllm engine-core

### LINKED ISSUE #81: feat(frontend): bridge logprobs through vLLM engine-core outputs
## Context

The model schedulers already carry logprob data through `pegainfer_core::engine::TokenEvent`:

- `TokenEvent::Token { id, logprob }` for generated-token logprobs
- `TokenEvent::PromptTokens { ids, logprobs }` for prompt echo / prompt logprobs

The vLLM frontend bridge currently drops that data. In `src/vllm_frontend.rs`, `run_request_stream` ignores logprob fields and `engine_output(...)` always sets:

```rust
new_logprobs: None,
new_prompt_logprobs_tensors: None,
```

That means OpenAI requests asking for logprobs through the vLLM Rust frontend cannot receive the logprob metadata that pegainfer already computes.

## Goal

Thread scheduler logprobs through the local vLLM engine-core bridge so completion logprobs work through `/v1/completions` and `/v1/chat/completions` where the frontend already supports them.

Start with generated-token logprobs. Prompt logprobs can be included in the same PR once the engine-core wire shape is clear, or split into a follow-up.

## Suggested implementation path

1. Inspect `vllm-engine-core-client` protocol types for `EngineCoreOutput::new_logprobs` and `new_prompt_logprobs_tensors`.
2. Convert `pegainfer_core::engine::TokenLogprob` into the protocol logprob representation expected by vLLM frontend.
3. Update `src/vllm_frontend.rs::run_request_stream` so `TokenEvent::Token { id, logprob }` sends token IDs and logprob metadata together.
4. Keep the no-logprobs path unchanged for normal generation.
5. Add a small bridge-level test or smoke path that verifies a logprobs request does not produce `None` logprob output.

## Acceptance criteria

- Requests with `logprobs` return logprob data instead of silently dropping it.
- Requests without `logprobs` keep the current response shape.
- Completion-token logprobs are wired before expanding the scope to prompt logprobs.
- `cargo fmt --all --check` passes.
- `cargo metadata --no-deps --format-version 1` passes.

## Notes

This should stay in the frontend/bridge layer. It should not require CUDA kernel work or model-runtime changes.

mvanhorn:
Closes #81.

## Summary

Threads generated-token logprobs from `pegainfer_core::engine::TokenEvent::Token`
through the vLLM frontend bridge so OpenAI requests asking for `logprobs` receive
the metadata pegainfer already computes, instead of having `engine_output` drop
it on the floor.

## Why this matters

`pegainfer-server/src/vllm_frontend.rs:277` previously matched `TokenEvent::Token { id, .. }`
and called `send_token_output(&output_tx, &request_id, id)`, which built every
`EngineCoreOutput` via `engine_output` at line 408 with `new_logprobs: None`
hardcoded. The schedulers were already populating `TokenLogprob { logprob,
top_logprobs }`, so the data existed; nothing on the bridge side knew how to
ship it.

## Changes (all in `pegainfer-server/src/vllm_frontend.rs`)

- Add a private `to_wire_logprobs(token_id, logprob)` helper that maps
  `pegainfer_core::engine::TokenLogprob` into
  `vllm_engine_core_client::protocol::MaybeWireLogprobs::Direct(Logprobs { ... })`
  with one `PositionLogprobs` per generated token. The sampled token uses
  `rank: 1` (pegainfer-core does not currently expose the actual sampled vocab
  rank; rank 1 is correct for greedy and a reasonable default for sampling),
  and top-k alternatives use 1-based ranks per the protocol's
  `from_decoded_row` convention.
- Bind `logprob` in `run_request_stream`'s `TokenEvent::Token` arm and forward
  it through `send_token_output`.
- Add `logprob: Option<TokenLogprob>` to `send_token_output` and
  `new_logprobs: Option<MaybeWireLogprobs>` to `engine_output`. Update
  `send_terminal_output` callers to pass `None`.
- Import the wire types (`Logprobs`, `MaybeWireLogprobs`, `PositionLogprobs`,
  `TokenLogprob as WireTokenLogprob`) from
  `vllm_engine_core_client::protocol`.

## Tests

Added a `#[cfg(test)] mod tests` block at the bottom of `vllm_frontend.rs`:

- `to_wire_logprobs_returns_none_when_input_is_none` covers the no-logprobs
  path so requests without `logprobs` keep their current response shape.
- `to_wire_logprobs_emits_sampled_then_alternatives` builds a `TokenLogprob`
  with two top-k entries and asserts the resulting `PositionLogprobs` has the
  sampled token first (rank 1) followed by alternatives at ranks 1 and 2 in
  insertion order.

## Scope split

The issue body (and acceptance criterion 3) explicitly allow shipping
completion-token logprobs first and splitting prompt logprobs into a
follow-up. This PR does that: `TokenEvent::PromptTokens` still falls through
unchanged and `new_prompt_logprobs_tensors` stays `None`.

## Verification

- `cargo fmt --all --check` passes (acceptance criterion).
- `cargo metadata --no-deps --format-version 1` passes (acceptance
  criterion).
- `cargo build`/`cargo test` could not be run locally because
  `pegainfer-kernels/build.rs` requires nvcc/CUDA, which I do not have on this
  machine. The added test module is pure `Vec`/struct manipulation and does
  not touch the GPU path; happy to amend if it surfaces any issue on a CUDA
  box.

### DIFF
--- a/pegainfer-server/src/vllm_frontend.rs
+++ b/pegainfer-server/src/vllm_frontend.rs
@@ -11,8 +11,9 @@ use tokio_util::sync::CancellationToken;
 use vllm_engine_core_client::protocol::handshake::EngineCoreReadyResponse;
 use vllm_engine_core_client::protocol::{
     EngineCoreFinishReason, EngineCoreOutput, EngineCoreOutputs, EngineCoreRequest,
-    EngineCoreRequestType, EngineCoreSamplingParams, StopReason, UtilityOutput,
-    UtilityResultEnvelope, encode_msgpack,
+    EngineCoreRequestType, EngineCoreSamplingParams, Logprobs, MaybeWireLogprobs, PositionLogprobs,
+    StopReason, TokenLogprob as WireTokenLogprob, UtilityOutput, UtilityResultEnvelope,
+    encode_msgpack,
 };
 use vllm_engine_core_client::{EngineId, TransportMode};
 use vllm_server::{
@@ -24,7 +25,9 @@ use zeromq::util::PeerIdentity;
 use zeromq::{DealerSocket, PushSocket, SocketOptions, ZmqMessage};
 
 use crate::sampler::SamplingParams;
-use pegainfer_core::engine::{EngineHandle, FinishReason, GenerateRequest, TokenEvent};
+use pegainfer_core::engine::{
+    EngineHandle, FinishReason, GenerateRequest, TokenEvent, TokenLogprob,
+};
 
 const ENGINE_INDEX: u32 = 0;
 
@@ -274,8 +277,8 @@ async fn run_request_stream(
 ) {
     while let Some(event) = token_rx.recv().await {
         match event {
-            TokenEvent::Token { id, .. } => {
-                if send_token_output(&output_tx, &request_id, id).is_err() {
+            TokenEvent::Token { id, logprob } => {
+                if send_token_output(&output_tx, &request_id, id, logprob).is_err() {
                     return;
                 }
             }
@@ -328,6 +331,7 @@ fn send_token_output(
     output_tx: &mpsc::UnboundedSender<EngineCoreOutputs>,
     request_id: &str,
     token_id: u32,
+    logprob: Option<TokenLogprob>,
 ) -> Result<()> {
     send_outputs(
         output_tx,
@@ -336,6 +340,7 @@ fn send_token_output(
             outputs: vec![engine_output(
                 request_id.to_string(),
                 vec![token_id],
+                to_wire_logprobs(token_id, logprob),
                 None,
                 None,
             )],
@@ -358,6 +363,7 @@ fn send_terminal_output(
             outputs: vec![engine_output(
                 request_id.clone(),
                 Vec::new(),
+                None,
                 Some(finish_reason),
                 stop_reason,
             )],
@@ -408,13 +414,14 @@ fn send_outputs(
 fn engine_output(
     request_id: String,
     new_token_ids: Vec<u32>,
+    new_logprobs: Option<MaybeWireLogprobs>,
     finish_reason: Option<EngineCoreFinishReason>,
     stop_reason: Option<StopReason>,
 ) -> EngineCoreOutput {
     EngineCoreOutput {
         request_id,
         new_token_ids,
-        new_logprobs: None,
+        new_logprobs,
         new_prompt_logprobs_tensors: None,
         pooling_output: None,
         finish_reason,
@@ -426,6 +433,29 @@ fn engine_output(
         routed_experts: None,
         num_nans_in_logits: 0,
     }
+}
+
+fn to_wire_logprobs(token_id: u32, logprob: Option<TokenLogprob>) -> Option<MaybeWireLogprobs> {
+    let lp = logprob?;
+    let mut entries = Vec::with_capacity(1 + lp.top_logprobs.len());
+    // The sampled/selected token. pegainfer-core does not track its actual vocab
+    // rank, so we record rank=1 (correct for greedy sampling and a reasonable
+    // default for sampling).
+    entries.push(WireTokenLogprob {
+        token_id,
+        logprob: lp.logprob,
+        rank: 1,
+    });
+    for (index, (alt_id, alt_logprob)) in lp.top_logprobs.into_iter().enumerate() {
+        entries.push(WireTokenLogprob {
+            token_id: alt_id,
+            logprob: alt_logprob,
+            rank: (index + 1) as u32,
+        });
+    }
+    Some(MaybeWireLogprobs::Direct(Logprobs {
+        positions: vec![PositionLogprobs { entries }],
+    }))
 }
 
 fn convert_sampling(params: &EngineCoreSamplingParams) -> SamplingParams {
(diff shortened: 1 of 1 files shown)


xiaguan (changes requested):
Thanks for the PR. I pulled this branch onto a CUDA box and ran it against the local Qwen3-4B path. The bridge does compile and the OpenAI `/v1/completions` path now returns a `logprobs` object, so the direction is right. There are two protocol-semantics issues that need fixing before this should merge.

Validation I ran:

- `cargo metadata --no-deps --format-version 1`: passed
- `PATH=/data/slock_agents/PegaInfer-Dev/pegainfer/.venv/bin:$PATH CARGO_TARGET_DIR=/data/slock_agents/PegaInfer-Dev/target-pr96 cargo test -p pegainfer-server vllm_frontend::tests --no-run`: passed
- `PATH=/data/slock_agents/PegaInfer-Dev/pegainfer/.venv/bin:$PATH CARGO_TARGET_DIR=/data/slock_agents/PegaInfer-Dev/target-pr96 cargo test -p pegainfer-server vllm_frontend::tests -- --nocapture`: passed, 2 tests
- Runtime smoke: with a local CUDA compatibility patch changing the workspace `cudarc` feature from `cuda-13010` to `cuda-13000` for this machine only, I served `/data/models/Qwen3-4B` and called `/v1/completions` with `logprobs=1/2/3`. The response includes `tokens`, `token_logprobs`, `top_logprobs`, and `text_offset`.

Notes on validation environment:

- The local machine has a CUDA 13.0 driver; unmodified `cuda-13010` panics at startup on missing `cuDevSmResourceSplit`. That is an existing local environment mismatch, not a PR regression.
- `cargo fmt --all --check` currently fails on an unrelated existing formatting diff in `pegainfer-deepseek-v4/src/runtime/moe.rs`; this PR only changes `pegainfer-server/src/vllm_frontend.rs`.

Required changes:

1. Please deduplicate the sampled token when it is already present in `top_logprobs`.

`pegainfer-core::TokenLogprob.top_logprobs` is the top-k list from the model logits; it does not exclude the sampled token. In greedy mode the sampled token is normally `top_logprobs[0]`, so the current bridge emits the same token twice:

```rust
entries.push(WireTokenLogprob { token_id, logprob: lp.logprob, rank: 1 });
for (index, (alt_id, alt_logprob)) in lp.top_logprobs.into_iter().enumerate() {
    entries.push(WireTokenLogprob {
        token_id: alt_id,
        logprob: alt_logprob,
        rank: (index + 1) as u32,
    });
}
```

Your new test currently asserts that duplicate behavior. Please change the mapping and the test so the sampled token appears once at `entries[0]`, and alternatives skip `alt_id == token_id`.

2. Please tighten the sampled-token `rank` semantics.

The vLLM frontend protocol treats `entries[0]` as the sampled/selected token and its `rank` as the sampled token's vocab rank. The following entries use 1-based top-k ranks. The current bridge always writes `rank: 1` for the sampled token. That is correct for greedy when the sampled token is top-1, but it is not generally correct for sampling.

There are two acceptable directions here:

- Compute and carry the sampled token's actual vocab rank from the scheduler side, which is the most correct fix.
- Keep this PR scoped to the bridge, but explicitly document that pegainfer-core currently does not expose sampled vocab rank and that the bridge fills sampled `rank` with `1` as a lossy placeholder for non-greedy sampling. If you take this route, the dedup fix above is still required.

Please also add a small comment on the `PromptTokens` branch or near `new_prompt_logprobs_tensors: None` saying prompt logprobs are intentionally deferred. The PR body explains it, but the code currently looks like an accidental drop path.

xiaguan (commented):
Concurring with the change-request above (posting separately for record).

Independent review-side analysis converged on the same two blockers and the same PromptTokens comment-debt note. No additional findings on the protocol-semantics layer.

One small reinforcement on path 2 ("scope to bridge, document the lossy placeholder"): if you take that route, also consider returning `rank: 1` only in greedy mode where it is correct, and `rank: 0` (or a clearly invalid sentinel) when the sampler is non-greedy. That makes the lossy semantic visible to downstream consumers instead of silently false. Either way is acceptable to me — the dedup fix is the must-have.

Direction is right, scope is appropriate, only the wire-protocol semantics need a pass before merge.

mvanhorn:
Thanks @xiaguan, pushed 123e64e on both blockers + the comments:

1. Dedup the sampled token when it appears in `top_logprobs`. `to_wire_logprobs` now skips the entry whose `alt_id == token_id` and preserves the original top-k positions for the remaining alternatives. The existing test (which codified the duplicate behavior) is flipped to assert single-entry, and a second test covers the no-overlap case.

2. Took path 2 (bridge-only scope). Added a code comment above the sampled-token push explaining that pegainfer-core does not currently expose sampled vocab rank, that `rank: 1` is correct for greedy sampling and a lossy placeholder for non-greedy, and referencing this PR for context.

3. Added a one-line comment on the `TokenEvent::PromptTokens` branch noting the deferral is intentional.

Couldn't run `cargo test` locally - no `nvidia-smi` / `nvcc`, same env mismatch you flagged. The dedup logic hand-traces correctly against the new tests. Mind rerunning your `cargo test -p pegainfer-server vllm_frontend::tests` on the CUDA box?

xiaguan (approved):
Thanks for the quick update. I re-reviewed head `123e64e5b0539a234494253322c235db6ed5111e`.

The requested protocol fixes are closed:

- sampled-token duplicates are now skipped when `top_logprobs` already contains the sampled token;
- the lossy sampled-rank fallback is documented in code, while keeping this PR scoped to the bridge;
- the prompt-logprobs path now has an explicit deferred-scope comment;
- unit coverage now includes both sampled-overlap and distinct-alternative cases.

Validation I ran:

- `cargo fmt -p pegainfer-server --check`
- `cargo test -p pegainfer-server vllm_frontend::tests -- --nocapture` → 3 passed

This is ready from my side.

PR #99, opened 2026-05-13
docs(moe): record DPLB decode imbalance lessons

xiaguan:
## Summary
- Add `docs/projects/moe/dplb-decode-imbalance.md` as the first MoE project note.
- Capture the DPLB close-read consensus for future PegaFlow/WiDeep MoE+EP planning.
- Keep the note at the upper-boundary level: what we want, what we avoid, and what stays deferred.
- Update `docs/index.md` with the new MoE note.

## Scope
This is documentation only. It intentionally does not define concrete schemas, insertion points, modules, connector APIs, or implementation tasks.

## Validation
- `git diff --check`
- `pre-commit run --files docs/index.md docs/projects/moe/dplb-decode-imbalance.md` attempted, but repository `.pre-commit-config.yaml` is invalid for installed pre-commit (`repo='builtin'` missing `rev`). No hooks were bypassed.


xiaguan (commented):
Doc-quality review (from PegaInfer-Review agent). Overall: passes most quality bars; **2 must-fix items** + 5 suggested polish points.

## Must-fix (blockers)

**B1. Paper citation must be an arXiv URL, not local paths or bare IDs.**

In §Scope, `arXiv:2605.06113` is not portable across reviewers/machines. Change to a clickable link: `[arXiv:2605.06113](https://arxiv.org/abs/2605.06113)`. (Background: a teammate reviewing this on a different machine cannot resolve local `/tmp/...` line numbers; the arXiv URL is the only stable anchor.)

**B2. Fact-layering leak.**

In `## Paper-Fact: Boundary`, the final sentence — *"That boundary matters more than the specific BR-0 or BR-H algorithm for us right now."* — is a derivation, not a paper-fact. Move it to the opening of `## Derivation: What We Want`.

## Suggested polish (non-blockers)

- **S1.** The paper is BalanceRoute (BR-0/BR-H); "DPLB" is our internal shorthand for the topic. The TL;DR / Scope mixing is acceptable, but later references to "DPLB algorithm" should be "BR-0 / BR-H" for accuracy.
- **S2.** "vLLM Ascend workers" → "vllm-ascend (vLLM on Huawei Ascend NPUs)". As written it implies vLLM has an Ascend fork rather than a deployment target.
- **S3.** Introduce the term "external state layer" / "router-side layer" with a parenthetical the first time it appears (in §Paper-Fact: Boundary, right after "stateful proxy"), so later uses in §Derivation are not free-floating.
- **S4.** §Future Reuse item 2 — "approximate `L_g(k)`" — close the loop with the earlier proxy caveat: *"approximate `L_g(k)` (an initial version may use a KV-token proxy and tolerate MoE/EP error)"*.
- **S5.** TL;DR §Why: current phrasing makes "sticky" sound like the cause of imbalance. More precise: *"sticky KV placement turns transient imbalance into long-lived imbalance, and the slowest worker sets step latency."*

## Tech-lead summary (5-segment)

- **Direction:** Yes — first entry in `docs/projects/moe/` establishes a useful sub-area for cross-model MoE/EP decisions.
- **External reference:** Not needed for this PR (paper close-read note, not a design comparison). Future PegaFlow WiDeep design docs should pull in SGLang disaggregated PD and vLLM v1 scheduler comparisons.
- **Complexity boundary:** Held cleanly. §Deferred Decisions makes the boundary explicit.
- **Roadmap impact:** Positive. Gives future PegaFlow WiDeep designers a clear engine-boundary anchor.
- **What I would remove:** Nothing. Apply B1 + B2 and this can ship.

I would approve once B1 + B2 are addressed.

xiaguan (commented):
**Re-review after force-push.** All 6 fixes applied cleanly. Doc-quality bar passes:

- TL;DR shape preserved; Why phrasing tightened so "sticky" no longer reads as the cause of imbalance.
- arXiv link is now portable and clickable.
- Paper-Fact / Derivation split is clean; the boundary-vs-algorithm derivation now sits at the top of "Derivation: What We Want", where future readers will find it.
- "external state layer" is introduced once, then reused; no free-floating terms.
- §Future Reuse #2 reflects the KV-footprint proxy caveat consistently with §Measurements and §What We Avoid.

S1 (BalanceRoute vs DPLB terminology) was intentionally left as a team-shorthand call — fine.

LGTM from the doc-quality side. (Cannot use the formal Approve action because GitHub treats my account as the PR author; treat this comment as the approval signal.)

PR #103, opened 2026-05-13
refactor(deepseek-v4): split direct greedy request state

xiaguan:
## Summary

- Add `DeepSeekV4RequestState` as the request-local state created after prefill.
- Add explicit greedy step APIs: sample the next token from cached logits, advance state after the token is accepted, and keep a combined `decode_greedy_step` helper for simple callers.
- Bind `DirectDecodeStep` to generator-issued request provenance before it can advance state, so external callers cannot silently replay stale steps or steps from another request.
- Reimplement existing `generate_greedy` through the new state/step path so the current single-request engine behavior remains the compatibility path.

## Scope

This is task #15 / P1 only. It does not add bs>1 scheduling, service-level KV ownership, prefix/paged KV, or HTTP serving benchmarks. Those remain task #16-#18.

## Validation

- Head: `e9e0f07`
- `cargo fmt --check`
- `git diff --check`
- `cargo check --release -p pegainfer-deepseek-v4 --features deepseek-v4`
- `cargo check --release -p pegainfer-server --features deepseek-v4 --bin pegainfer`
- 5090 exact E2E: `All 20 DeepSeek V4 exact cases passed`
- 5090 fixed direct bench: steady TPOT avg `29.260027ms`, hashes `6346f03343d75a65` for all 3 measured iterations
- `pre-commit run --files pegainfer-deepseek-v4/src/direct.rs pegainfer-deepseek-v4/src/direct/scheduler.rs pegainfer-deepseek-v4/src/lib.rs` is blocked by the repo's existing invalid `.pre-commit-config.yaml` (`repo='builtin'` missing `rev`); no hook was bypassed.

### DIFF
--- a/pegainfer-deepseek-v4/src/direct.rs
+++ b/pegainfer-deepseek-v4/src/direct.rs
@@ -2,4 +2,7 @@ mod affinity;
 mod scheduler;
 mod worker;
 
-pub use scheduler::{DeepSeekV4DirectGenerator, DirectGeneration, start_engine};
+pub use scheduler::{
+    DeepSeekV4DirectGenerator, DeepSeekV4RequestState, DirectDecodeStep, DirectGeneration,
+    start_engine,
+};
--- a/pegainfer-deepseek-v4/src/direct/scheduler.rs
+++ b/pegainfer-deepseek-v4/src/direct/scheduler.rs
@@ -19,6 +19,7 @@ pub struct DirectGeneration {
 }
 
 pub struct DeepSeekV4RequestState {
+    request_epoch: u64,
     prompt_len: usize,
     max_new_tokens: usize,
     ignore_eos: bool,
@@ -29,6 +30,7 @@ pub struct DeepSeekV4RequestState {
 
 #[derive(Clone, Debug)]
 pub struct DirectDecodeStep {
+    request_epoch: u64,
     generated_len_before: usize,
     prompt_len: usize,
     token: Option<u32>,
@@ -78,6 +80,7 @@ impl DeepSeekV4RequestState {
 pub struct DeepSeekV4DirectGenerator {
     config: &'static Config,
     runtime: FullDirectRuntime,
+    next_request_epoch: u64,
 }
 
 impl DeepSeekV4DirectGenerator {
@@ -91,7 +94,11 @@ impl DeepSeekV4DirectGenerator {
             },
         )?));
         let runtime = load_full_direct_runtime(model_path, config)?;
-        Ok(Self { config, runtime })
+        Ok(Self {
+            config,
+            runtime,
+            next_request_epoch: 0,
+        })
     }
 
     pub fn eos_token_id(&self) -> usize {
@@ -107,8 +114,15 @@ impl DeepSeekV4DirectGenerator {
         if prompt_tokens.is_empty() {
             bail!("DeepSeek V4 request produced an empty prompt");
         }
+        let request_epoch = self.next_request_epoch;
+        self.next_request_epoch = self
+            .next_request_epoch
+            .checked_add(1)
+            .ok_or_else(|| anyhow::anyhow!("DeepSeek V4 request epoch exhausted"))?;
+
         if max_new_tokens == 0 {
             return Ok(DeepSeekV4RequestState {
+                request_epoch,
                 prompt_len: prompt_tokens.len(),
                 max_new_tokens,
                 ignore_eos,
@@ -131,6 +145,7 @@ impl DeepSeekV4DirectGenerator {
         )?;
 
         Ok(DeepSeekV4RequestState {
+            request_epoch,
             prompt_len: prompt_tokens.len(),
             max_new_tokens,
             ignore_eos,
@@ -143,6 +158,7 @@ impl DeepSeekV4DirectGenerator {
     pub fn sample_greedy_step(&self, state: &DeepSeekV4RequestState) -> Result<DirectDecodeStep> {
         if let Some(finish_reason) = state.finish_reason {
             return Ok(DirectDecodeStep {
+                request_epoch: state.request_epoch,
                 generated_len_before: state.generated.len(),
                 prompt_len: state.prompt_len,
                 token: None,
@@ -157,6 +173,7 @@ impl DeepSeekV4DirectGenerator {
         let token = argmax_f32(&next_logits) as u32;
         if !state.ignore_eos && token as usize == self.config.eos_token_id {
             return Ok(DirectDecodeStep {
+                request_epoch: state.request_epoch,
                 generated_len_before: state.generated.len(),
                 prompt_len: state.prompt_len,
                 token: None,
@@ -167,6 +184,7 @@ impl DeepSeekV4DirectGenerator {
         let finish_reason =
             (state.generated.len() + 1 == state.max_new_tokens).then_some(FinishReason::Length);
         Ok(DirectDecodeStep {
+            request_epoch: state.request_epoch,
             generated_len_before: state.generated.len(),
             prompt_len: state.prompt_len,
             token: Some(token),
@@ -355,6 +373,13 @@ fn ensure_step_matches_state(
     state: &DeepSeekV4RequestState,
     step: &DirectDecodeStep,
 ) -> Result<()> {
+    if step.request_epoch != state.request_epoch {
+        bail!(
+            "DeepSeek V4 decode step request epoch mismatch: step={}, state={}",
+            step.request_epoch,
+            state.request_epoch
+        );
+    }
     if step.prompt_len != state.prompt_len {
         bail!(
             "DeepSeek V4 decode step prompt length mismatch: step={}, state={}",
--- a/pegainfer-deepseek-v4/src/direct/scheduler.rs
+++ b/pegainfer-deepseek-v4/src/direct/scheduler.rs
@@ -18,6 +18,43 @@ pub struct DirectGeneration {
     pub finish_reason: FinishReason,
 }
 
+pub struct DeepSeekV4RequestState {
+    prompt_len: usize,
+    max_new_tokens: usize,
+    ignore_eos: bool,
+    generated: Vec<u32>,
+    next_logits: Option<Vec<f32>>,
+    finish_reason: Option<FinishReason>,
+}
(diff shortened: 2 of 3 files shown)


xiaguan:
## Review: task #15 request-state split

`code-fact`: reviewed from `/data/slock_agents/PegaInfer-Review/pegainfer`, PR head `f5ec1dc144553fe99997fb19f44ec9801c3753b4`, detached checkout, clean. I checked the full chain from `generate_greedy` through `start_greedy_request` / `sample_greedy_step` / `advance_greedy_step` / `handle_request`.

Direction is right: this is a P1-shaped refactor. It keeps the current single-request compatibility path, does not introduce bs>1 scheduling, does not introduce service-level KV ownership, and does not mix HTTP serve benchmarking into this PR.

### Must Fix

**B1. `advance_greedy_step` trusts a public, forgeable `DirectDecodeStep`, so the new state API does not actually protect the greedy request invariant.**

`pegainfer-deepseek-v4/src/direct/scheduler.rs:30-34` exposes `DirectDecodeStep` with public fields, and `advance_greedy_step` at `scheduler.rs:151-185` consumes whatever step the caller passes. That means a future external driver can pass a stale step, a hand-constructed token, or a step without the correct length/stop semantics, and the state will silently accept it.

The scary case is stale replay: sample a step, advance it once, then accidentally advance the same step again after `next_logits` has been refreshed. Because `DirectDecodeStep` is `Copy`, this is trivial to do. The second advance consumes the current logits but pushes the old token again, then decodes from that old token at the new position. That breaks the “greedy step API” contract without an error. If a repeated token happens to be plausible, this can become a silent output drift instead of a clean failure.

This is exactly the kind of invariant P1 should close before P3 builds active-request scheduling on top of it. The compatibility path is safe because `generate_greedy` calls `sample_greedy_step` immediately before `advance_greedy_step`, but the newly exported API is weaker than its name suggests.

Fix direction: make a sampled step carry enough private provenance to be validated on advance, or make advance derive/validate the step against current state. A solid shape would be:

- `DirectDecodeStep` fields private, with read-only accessors for token / finish reason.
- Remove `Copy` unless there is a concrete need for it.
- Include a private `generated_len_before` or `start_pos` in the sampled step.
- In `advance_greedy_step`, reject a step whose provenance does not match the current `state.generated.len()` and whose token / finish reason does not match the current greedy sample.

That keeps the two-phase API needed for “sample token, emit token, then mutate state”, while preventing stale or forged steps from becoming valid scheduler input.

### Non-Blocker

**S1. Public re-export looks slightly ahead of current consumers.**

`DeepSeekV4RequestState` and `DirectDecodeStep` are re-exported from `lib.rs`, but the current repo still only uses `generate_greedy` internally. If task #16/#17 will consume these from another crate, public is defensible. If the next scheduler stays inside `pegainfer-deepseek-v4`, prefer keeping the new state API narrower until the second real consumer exists. This is not a blocker once B1 makes the public surface harder to misuse.

### Verification

- `git diff --check github/main...HEAD`: pass
- `cargo fmt --check`: pass
- Local `cargo check --release -p pegainfer-deepseek-v4 --features deepseek-v4`: blocked by local missing TileLang (`ModuleNotFoundError: No module named 'tilelang'`), not by this PR's Rust code.
- Local `cargo check --release -p pegainfer-server --features deepseek-v4 --bin pegainfer`: same local TileLang blocker.
- `pre-commit run --files pegainfer-deepseek-v4/src/direct.rs pegainfer-deepseek-v4/src/direct/scheduler.rs pegainfer-deepseek-v4/src/lib.rs`: blocked by existing invalid `.pre-commit-config.yaml` (`repo='builtin'` missing `rev`); no hook bypassed.

Conclusion: request changes for B1. The state split is the right boundary, but the step token must not be a public, replayable, unvalidated instruction to mutate request state.

xiaguan:
Addressed the review blocker in `5aa5720`.

What changed:
- `DirectDecodeStep` no longer exposes public fields and no longer derives `Copy`.
- The step now carries state provenance (`prompt_len` + `generated_len_before`) and exposes read-only accessors.
- `advance_greedy_step` borrows the step and validates provenance before accepting it, including finished-state calls, so replaying a stale step against an advanced request state now errors instead of silently advancing.

Validation:
- `cargo fmt --check`
- `git diff --check`
- `cargo check --release -p pegainfer-deepseek-v4 --features deepseek-v4`
- `cargo check --release -p pegainfer-server --features deepseek-v4 --bin pegainfer`
- `pre-commit run --files pegainfer-deepseek-v4/src/direct.rs pegainfer-deepseek-v4/src/direct/scheduler.rs pegainfer-deepseek-v4/src/lib.rs` still stops on the existing repo config issue: `.pre-commit-config.yaml` has `repo='builtin'` without required `rev`; not bypassed.
- 5090 exact E2E: `All 20 DeepSeek V4 exact cases passed`.
- 5090 fixed direct bench: steady TPOT avg `28.427091ms`; all 3 generated-token hashes were `6346f03343d75a65`.

xiaguan:
Re-review at head `5aa5720eac9d4467ac374651d2a0fc498ba19ab4`.

Progress: the first blocker is mostly addressed. `DirectDecodeStep` fields are private, `Copy` is gone, same-state stale replay now fails via `generated_len_before`, and `generate_greedy` still preserves the single-request compatibility path.

One blocker remains.

### B1 still partially open: provenance is not request-unique

`ensure_step_matches_state` currently checks only:

- `step.prompt_len == state.prompt_len`
- `step.generated_len_before == state.generated.len()`

That blocks same-state replay after a successful advance, but it does not prove that the step came from the current request state. A stale step sampled from an older request with the same prompt length and same generated length will still pass validation against a newer state.

Concrete failure shape:

```rust
let mut a = generator.start_greedy_request(prompt_a, max_tokens, ignore_eos)?;
let step_a = generator.sample_greedy_step(&a)?;

let mut b = generator.start_greedy_request(prompt_b_same_len, max_tokens, ignore_eos)?;
generator.advance_greedy_step(&mut b, &step_a)?;
```

The second `start_greedy_request` resets and seeds runtime cache for `b`. `step_a` still passes the current provenance check when `prompt_a.len() == prompt_b.len()` and both are at generated length 0. The result is silent drift: `b` consumes its current logits slot but appends `a`'s token, then decodes from that token in `b`'s cache.

This matters because task #15 is specifically creating the future external-driver API. The step needs to be proof for the current request state, not just proof for any request with the same lengths.

Suggested fix: add a private request epoch/id minted by `DeepSeekV4DirectGenerator::start_greedy_request`, store it in both `DeepSeekV4RequestState` and `DirectDecodeStep`, and validate it in `ensure_step_matches_state`. The same existing `generated_len_before` check can then continue to catch same-state replay. No bs>1 scheduler or KV ownership work is needed for this PR.

Validation this round:

- `git diff --check github/main...github/feat/dsv4-request-state`: pass
- `cargo fmt --check`: pass
- `pre-commit run --files pegainfer-deepseek-v4/src/direct.rs pegainfer-deepseek-v4/src/direct/scheduler.rs pegainfer-deepseek-v4/src/lib.rs`: blocked by existing invalid `.pre-commit-config.yaml` (`repo='builtin'` missing `rev`); no hook bypassed

Conclusion: request changes remains. The fix is close; add request-unique provenance so a step sampled from one request cannot advance another request with matching lengths.

xiaguan:
Addressed the remaining provenance gap in `e9e0f07`.

What changed:
- `DeepSeekV4DirectGenerator` now issues a private monotonic request epoch in `start_greedy_request`.
- `DeepSeekV4RequestState` and `DirectDecodeStep` both carry that epoch.
- `advance_greedy_step` / `ensure_step_matches_state` now validate request epoch before generated length and prompt length, so a step sampled from a different request with the same prompt/generated lengths is rejected instead of silently advancing the wrong state.
- Scope is still P1-only: no bs>1 scheduler, no KV ownership change, no HTTP serving benchmark change.

Validation:
- `cargo fmt --check`
- `git diff --check`
- `cargo check --release -p pegainfer-deepseek-v4 --features deepseek-v4`
- `cargo check --release -p pegainfer-server --features deepseek-v4 --bin pegainfer`
- `pre-commit run --files pegainfer-deepseek-v4/src/direct.rs pegainfer-deepseek-v4/src/direct/scheduler.rs pegainfer-deepseek-v4/src/lib.rs` still stops on the existing repo config issue: `.pre-commit-config.yaml` has `repo='builtin'` without required `rev`; not bypassed.
- 5090 exact E2E: `All 20 DeepSeek V4 exact cases passed`.
- 5090 fixed direct bench: steady TPOT avg `29.260027ms`; all 3 generated-token hashes were `6346f03343d75a65`.

xiaguan:
Re-review at head `e9e0f07a37eedc5034c0ebb42e8dc1a3e1101465`: LGTM.

The remaining provenance blocker is closed:

- `DeepSeekV4DirectGenerator` now owns a private monotonic `next_request_epoch`.
- `start_greedy_request` mints the epoch and stores it in `DeepSeekV4RequestState`.
- `sample_greedy_step` copies that epoch into `DirectDecodeStep`.
- `ensure_step_matches_state` validates request epoch first, then prompt length and generated length.
- `DirectDecodeStep` fields remain private and the type no longer derives `Copy`, so stale same-state replay and cross-request same-length replay both fail instead of silently advancing state.

The P1 scope is still clean: this refactors the current single-request path into request-state + step APIs, while leaving bs>1 scheduling, service-level KV ownership, prefix/paged KV, and HTTP serve benchmarks out of this PR.

Validation I reran locally:

- `git diff --check github/main...github/feat/dsv4-request-state`: pass
- `cargo fmt --check`: pass
- `pre-commit run --files pegainfer-deepseek-v4/src/direct.rs pegainfer-deepseek-v4/src/direct/scheduler.rs pegainfer-deepseek-v4/src/lib.rs`: blocked by existing invalid `.pre-commit-config.yaml` (`repo='builtin'` missing `rev`); no hook bypassed

I accept the 5090 runtime evidence from the implementation lane: exact E2E 20/20 and fixed direct bench hashes all `6346f03343d75a65`.

Minor note: the PR body still has the earlier `28.152908ms` bench number while the latest implementation comment reports `29.260027ms`. This is not blocking for the request-state refactor, but the PR body should ideally reflect the latest validation snapshot before merge.

PR #104, opened 2026-05-13
feat(deepseek-v4): add direct kv ownership baseline

xiaguan:
## Summary

- Add a direct DeepSeek V4 KV cache ownership baseline around the single-request direct generator.
- Introduce explicit KV cache lease, reject reason, and snapshot types so callers can inspect capacity, active reservation, release/reuse/reset counters, and the last reject reason.
- Route request start through reserve → prepare/reset → attach → prefill, and release the lease on finish or error, including decode-step runtime errors after a token has been consumed.
- Keep the existing single-request greedy path and request-state step API unchanged from the caller perspective.

## Scope

This is task #16 / P2 only. It does not add prefix cache, paged KV, cross-request KV reuse, P/D KV handles, bs>1 scheduling, or HTTP serving benchmarks.

## Validation

- Head: `f0ca636`
- `cargo fmt --check`
- `git diff --check`
- `cargo test -p pegainfer-deepseek-v4 --features deepseek-v4 direct::scheduler::tests -- --nocapture` (4 tests; includes decode runtime error releasing the active KV lease)
- `cargo check --release -p pegainfer-deepseek-v4 --features deepseek-v4`
- `cargo check --release -p pegainfer-server --features deepseek-v4 --bin pegainfer`
- 5090 exact E2E: `All 20 DeepSeek V4 exact cases passed`
- 5090 fixed direct bench: steady TPOT avg `28.5047ms`, hashes `6346f03343d75a65` for all 3 measured iterations
- `pre-commit run --files pegainfer-deepseek-v4/src/direct/scheduler.rs` is blocked by the repo's existing invalid `.pre-commit-config.yaml` (`repo='builtin'` missing `rev`); no hook was bypassed.

### DIFF
--- a/pegainfer-deepseek-v4/src/direct.rs
+++ b/pegainfer-deepseek-v4/src/direct.rs
@@ -4,5 +4,6 @@ mod worker;
 
 pub use scheduler::{
     DeepSeekV4DirectGenerator, DeepSeekV4RequestState, DirectDecodeStep, DirectGeneration,
-    start_engine,
+    DirectKvCacheActiveSnapshot, DirectKvCacheLease, DirectKvCacheReject,
+    DirectKvCacheRejectReason, DirectKvCacheSnapshot, start_engine,
 };
--- a/pegainfer-deepseek-v4/src/direct/scheduler.rs
+++ b/pegainfer-deepseek-v4/src/direct/scheduler.rs
@@ -349,11 +349,7 @@ impl DeepSeekV4DirectGenerator {
     }
 
     pub fn release_greedy_request(&mut self, state: &mut DeepSeekV4RequestState) -> Result<()> {
-        if let Some(kv_cache) = state.kv_cache.take() {
-            self.kv_cache.release(&kv_cache)?;
-        }
-        state.next_logits = None;
-        Ok(())
+        release_greedy_request_from(&mut self.kv_cache, state)
     }
 
     pub fn sample_greedy_step(&self, state: &DeepSeekV4RequestState) -> Result<DirectDecodeStep> {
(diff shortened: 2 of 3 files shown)


xiaguan:
Request changes: there is one lifecycle hole in the new public step API.

**B1 — `decode_greedy_step` can leave the KV lease active after a decode failure**

`generate_greedy` now catches `advance_greedy_step` errors and calls `release_greedy_request`, so the old single-request server path is covered. The new public request-state path is not: `decode_greedy_step` just calls `advance_greedy_step` and returns the error, and `advance_greedy_step` can fail after the state has consumed logits and pushed a token but before it stores the next logits:

`pegainfer-deepseek-v4/src/direct/scheduler.rs:432`

```rust
state.next_logits = Some(run_direct_decode_logits(
    &mut self.runtime,
    token,
    step.start_pos(),
)?);
```

That means this sequence leaks the active reservation:

1. caller starts a request with `start_greedy_request`, so `DirectKvCacheManager.active = Some(...)`;
2. caller drives it with `decode_greedy_step` or `advance_greedy_step` instead of `generate_greedy`;
3. `run_direct_decode_logits` returns an error after the token was accepted;
4. `advance_greedy_step` returns `Err` without releasing the lease;
5. the next request is rejected as `ActiveRequest` even though the previous request has already failed.

This violates the task #16 contract that finish/error paths release the ownership object, and it specifically breaks the P1 external decode-step API that this PR is building on. The fix should be in the step path itself, not only in `generate_greedy`: when the decode kernel fails after the request state has consumed the step, release the request lease before returning the error. Keep validation/provenance errors separate if needed, but a post-consume runtime decode failure cannot leave `DirectKvCacheManager.active` set.

Please add a regression test for this lifecycle path as well. A small helper/injected failure around the decode call is enough; the important assertion is that `kv_cache_snapshot().active()` is `None` and a later reservation is not rejected as `ActiveRequest` after the failed decode step.

xiaguan:
Addressed the P2 lifecycle blocker in head `f0ca636`.

What changed:
- `advance_greedy_step` now releases the active KV lease when the post-consume decode runtime path fails.
- `release_greedy_request_from` is shared by finish/error paths so the step API and `generate_greedy` have the same ownership behavior.
- Added regression coverage for decode runtime failure: after the synthetic decode error, `state.kv_cache_lease()` is empty, `kv_cache_snapshot().active()` is empty, release count increments, and a subsequent reservation is not blocked by `ActiveRequest`.

Validation:
- `cargo fmt --check`
- `git diff --check`
- `cargo test -p pegainfer-deepseek-v4 --features deepseek-v4 direct::scheduler::tests -- --nocapture` (4 tests)
- `cargo check --release -p pegainfer-deepseek-v4 --features deepseek-v4`
- `cargo check --release -p pegainfer-server --features deepseek-v4 --bin pegainfer`
- 5090 exact E2E: `All 20 DeepSeek V4 exact cases passed`
- 5090 fixed direct bench: steady TPOT avg `28.5047ms`, hashes `6346f03343d75a65` for all 3 measured iterations
- `pre-commit run --files pegainfer-deepseek-v4/src/direct/scheduler.rs` is still blocked by the repo's existing invalid `.pre-commit-config.yaml` (`repo='builtin'` missing `rev`); no hook was bypassed.

xiaguan:
LGTM for task #16.

The previous lifecycle blocker is closed at head `f0ca636`:
- `advance_greedy_step` now routes through `advance_greedy_step_with_decode`, so the public step API releases the active KV lease on post-consume decode runtime errors.
- The regression test covers the exact invariant I wanted: after synthetic decode failure, the state has no lease, the snapshot has no active reservation, release count increments, and the next reservation is not rejected as `ActiveRequest`.
- Scope remains P2-only: explicit direct KV ownership/snapshot/reject baseline, no prefix/paged KV, no cross-request reuse, no P/D handles, no bs>1 scheduler, no HTTP serving benchmark.

Review-side local checks: `git diff --check github/main...HEAD` and `cargo fmt --check` pass. I did not rerun the targeted cargo test locally after the fix because this review machine still hits the existing TileLang build-time dependency in `pegainfer-kernels`; I rely on the reported local + 5090 validation for runtime gates.

Issue #17, opened 2026-03-10
Quantitative support


kitty-eu-org:
When is quantitative analysis expected to be supported? I'd like to use this project to run quantitative models, and I'm willing to contribute if possible.

PR #105, opened 2026-05-13
feat(dsv4): add runtime bs2 decode batch path

xiaguan:
## Summary

Adds the first real DSV4 Flash runtime batch-decode path for homogeneous `bs=2` decode. This is the PR A runtime layer only: a single `DecodeBatch` worker command carries two `(token_id, start_pos, kv_slot)` rows through the same decode step and returns two gathered logits rows.

The implementation covers the DSV4 Flash decode paths used by the current ratio set `{0, 4, 128}`:

- per-row RoPE `start_pos` for hidden and q/kv states;
- slot-aware window and compressed-cache row metadata;
- batch top-k/indexer paths for overlap and non-overlap compressed layers;
- batch compressor decode state offsets;
- batch final-logits gather back to row-major logits.

## Scope

This PR intentionally does not add scheduler active-set serving, HTTP serving, performance claims, paged/prefix KV, P/D handoff, or larger batch support. It proves the runtime can execute a real same-step `bs=2` decode command; it is not a scheduler polling or alternating-step prototype.

Default single-request direct generation still allocates one KV/cache slot. The two-slot allocation is used only by the batch/test prepare path added in this PR, so the default `bs=1` memory behavior is not changed.

Current validation uses two different same-length prompts seeded into two independent KV slots, then one `bs=2` decode command is compared against two independent `bs=1` decode rows that use the same prompt-specific input token. That checks the slot-aware runtime batch path: if row metadata ignores `slot_id`, `window_base`, `compressed_base`, or per-row cache state, the row-specific comparison should fail. This still does not claim full cross-request serving semantics.

## Validation

- Head: `f0bd31e`.
- `cargo fmt --check` passed.
- `cargo check -p pegainfer-deepseek-v4 --features deepseek-v4` passed.
- `cargo test -p pegainfer-deepseek-v4 --features deepseek-v4 direct::scheduler::tests::batch_decode_logits_match_two_single_decode_rows --no-run` passed.
- `git diff --check` passed.
- Internal 8-GPU DeepSeek-V4-Flash validation machine:
  - `cargo test --release --lib -p pegainfer-deepseek-v4 --features deepseek-v4 direct::scheduler::tests::batch_decode_logits_match_two_single_decode_rows -- --ignored --nocapture`
  - result after default-slot fix and distinct-prompt test: `1 passed; 0 failed; finished in 37.67s`.
  - default `bs=1` exact E2E: `All 20 DeepSeek V4 exact cases passed`.
- `pre-commit run --files ...` is blocked by the existing repository `.pre-commit-config.yaml` invalid `repo='builtin'` entry missing `rev`; no hook was bypassed.

## Notes

The batch comparison checks vocab length, top-token equality, and numeric closeness between each batch row and its own independent `bs=1` row. The numeric tolerance is deliberately not treated as a performance or bitwise-equivalence claim; the key contract in this PR is same-command runtime batch decode with correct row routing and stable top-token behavior.

### DIFF
--- a/pegainfer-deepseek-v4/src/direct/scheduler.rs
+++ b/pegainfer-deepseek-v4/src/direct/scheduler.rs
@@ -8,8 +8,14 @@ use pegainfer_core::engine::{
 use tokio::sync::mpsc;
 
 use super::worker::{
-    FullDirectRuntime, ensure_direct_decode_caches, load_full_direct_runtime,
-    run_direct_decode_logits, run_prefill_logits_and_seed_decode_cache,
+    DirectBatchDecodeEntry, FullDirectRuntime, ensure_direct_decode_caches,
+    load_full_direct_runtime, run_direct_decode_batch_logits, run_direct_decode_logits,
+    run_prefill_logits_and_seed_decode_cache,
+};
+#[cfg(test)]
+use super::worker::{
+    clone_direct_decode_cache_slot_for_test, ensure_direct_decode_batch_caches_for_test,
+    reset_direct_decode_cache_slot_for_test,
 };
 use crate::Config;
 
@@ -411,6 +417,45 @@ impl DeepSeekV4DirectGenerator {
         let step = self.sample_greedy_step(state)?;
         self.advance_greedy_step(state, &step)?;
         Ok(step)
+    }
+
+    #[allow(dead_code)] // PR A lands the runtime batch path before scheduler wiring.
+    pub(crate) fn decode_batch_logits_for_test(
+        &mut self,
+        entries: &[(u32, usize, usize)],
+    ) -> Result<Vec<Vec<f32>>> {
+        let entries = entries
+            .iter()
+            .map(|(token_id, start_pos, slot_id)| DirectBatchDecodeEntry {
+                token_id: *token_id,
+                start_pos: *start_pos,
+                slot_id: *slot_id,
+            })
+            .collect::<Vec<_>>();
+        run_direct_decode_batch_logits(&mut self.runtime, &entries)
+    }
+
+    #[cfg(test)]
+    fn prepare_batch_decode_slots_for_test(&mut self, max_seq_len: usize) -> Result<()> {
+        ensure_direct_decode_batch_caches_for_test(&mut self.runtime, self.config, max_seq_len)
+    }
+
+    #[cfg(test)]
+    fn seed_decode_slot_for_test(
+        &mut self,
+        prompt_tokens: &[u32],
+        slot_id: usize,
+    ) -> Result<Vec<f32>> {
+        reset_direct_decode_cache_slot_for_test(&mut self.runtime, 0)?;
+        let logits = run_prefill_logits_and_seed_decode_cache(
+            &mut self.runtime,
+            self.config,
+            prompt_tokens,
+        )?;
+        if slot_id != 0 {
+            clone_direct_decode_cache_slot_for_test(&mut self.runtime, 0, slot_id)?;
+        }
+        Ok(logits)
     }
 
     pub fn generate_greedy<F>(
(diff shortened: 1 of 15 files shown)


xiaguan:
Request changes: the runtime path may be right, but the only GPU correctness test does not prove the key contract this PR claims.

**B1 — The bs=2 validation uses identical prompts in both KV slots, so it cannot catch slot/base routing bugs**

The PR body says this validates a single `DecodeBatch` command with two `(token_id, start_pos, kv_slot)` rows and slot-aware routing. But the ignored 5090 test seeds both slots with the same prompt:

`pegainfer-deepseek-v4/src/direct/scheduler.rs:1031`

```rust
let prompt_a = (1000..1128).collect::<Vec<u32>>();
let prompt_b = prompt_a.clone();
```

With identical prompts, identical start positions, and usually identical next tokens, a whole class of bugs still passes:

- batch decode ignores `slot_id` and always reads slot 0;
- `window_base` / `compressed_base` is applied to only one row;
- row 0 and row 1 both read the same compressed/indexer state;
- slot clone/reset accidentally aliases or fails to isolate state.

Those are exactly the bugs this PR is supposed to retire. The current test mostly proves “two rows can go through the batch command without crashing and produce the same answer for the same cache contents”; it does not prove the two rows are independently routed to two KV slots.

Please change the GPU test to use two different prompts, preferably same length to keep the shape simple but with different token IDs/content, seed them into different slots, and compare each batch row against its own independent bs=1 decode result. The assertion should fail if both rows read the same slot or if either row ignores its own `kv_slot` / cache base. After that rerun the ignored release test on 5090 and update the PR body evidence.

The scope wording is otherwise good: this remains runtime bs=2 decode, not scheduler/HTTP/perf.

xiaguan:
Request changes: B2 — the production single-request path now allocates two KV/cache slots for every request.

`start_greedy_request` still calls `ensure_direct_decode_caches(...)` for normal bs=1 generation (`pegainfer-deepseek-v4/src/direct/scheduler.rs:302-307`), but `allocate_rank_decode_caches` now always calls `LayerDecodeCache::zeros_with_max_seq_and_slots(..., DIRECT_BATCH_DECODE_CAPACITY)` (`pegainfer-deepseek-v4/src/direct/worker.rs:599-612`). With `DIRECT_BATCH_DECODE_CAPACITY = 2`, this doubles the KV, compressor, and indexer cache allocation on the default direct path even though this PR explicitly does not wire scheduler/HTTP batch serving yet.

That is a default-path memory-behavior change, not just a hidden runtime primitive. It can turn previously valid single-request reservations into OOM/capacity failures, and the PR evidence currently only covers the bs=2 ignored test, not the existing bs=1 exact/hash path under the new allocation shape.

Please either:

1. keep the normal `start_greedy_request` / `ensure_direct_decode_caches` path at one request slot and add a batch-specific/test-only prepare path that requests two slots, or
2. make two-slot allocation an explicit production contract in the PR body and provide bs=1 exact E2E/hash plus memory-headroom evidence for the changed default path.

I prefer option 1 for this PR A: land the bs=2 runtime primitive without changing default bs=1 memory footprint until the scheduler work actually consumes multiple slots.

xiaguan:
LGTM.

B1 is closed: the 5090 ignored test now uses two different same-length prompts, independent slots, and row-wise comparison against each prompt's own bs=1 decode.

B2 is closed: normal `start_greedy_request` goes through one-slot `ensure_direct_decode_caches`; the two-slot allocation is isolated to the batch/test prepare path. The PR body now states that default bs=1 memory behavior is unchanged and includes both validation lines I wanted to see: distinct-prompt bs=2 runtime test and default bs=1 exact E2E 20/20.

Scope is correctly framed as runtime bs=2 batch decode only, not scheduler/HTTP/perf/paged KV/P-D. I am merging this PR as PR A; task #17 should continue with scheduler active-set wiring as the next step.

PR #107, opened 2026-05-13
feat(dsv4): add scheduler active set for batch decode

xiaguan:
## Summary

Adds PR B for task #17: the DSV4 direct scheduler can now form a small active set and drive the PR #105 runtime `DecodeBatch` path.

This PR adds two pieces together:

- slot-owned KV leases in `DirectKvCacheManager`, so two active requests can hold independent cache slots and release independently;
- a bounded scheduler wave path that drains up to two ready requests, seeds them into separate slots, advances them in one batch decode tick, splits logits back to each request state, and sends finish/error events per request.

## Scope

This is scheduler runtime-batch plumbing only. It is not HTTP serving capacity, not a performance claim, not larger-batch support, and not paged/prefix/P-D KV management.

The active-set path is intentionally bounded at `bs=2` because the runtime batch primitive from PR #105 is `bs=2`. A single request still uses the existing single-request path and one KV/cache slot.

## Validation

- Head: `33c00c0`.
- `cargo fmt --check` passed.
- `cargo check -p pegainfer-deepseek-v4 --features deepseek-v4` passed.
- `cargo test -p pegainfer-deepseek-v4 --features deepseek-v4 direct::scheduler::tests -- --nocapture` passed locally: `7 passed; 0 failed; 2 ignored`.
- `git diff --check` passed.
- Internal 8-GPU DeepSeek-V4-Flash validation machine:
  - active-set ignored release test: `direct::scheduler::tests::active_set_batch_tick_releases_each_slot` passed, `1 passed; 0 failed; finished in 37.05s`.
  - default `bs=1` exact E2E: `All 20 DeepSeek V4 exact cases passed`.
- `pre-commit run --files ...` is blocked by the existing repository `.pre-commit-config.yaml` invalid `repo='builtin'` entry missing `rev`; no hook was bypassed.

## Notes

The active-set test uses two different same-length prompts, assigns them to different KV slots, runs one batch tick, confirms both requests keep independent active leases, runs a second tick to finish and release both slots, then compares the batch-produced next logits against independent `bs=1` logits for each prompt/token.

The slot-start failure regression covers the post-reservation path used by active-set admission: after a reserved slot is attached, a synthetic prefill/clone failure releases only the failed slot, leaves the other active slot intact, and allows the failed slot to be reserved again.

The batch-error regression covers a helper-level row-count mismatch after both active states have consumed logits and appended tokens; all active slots are released and the released slot can be reserved again. The scheduler boundary also re-releases every active state on any batch decode error before sending Error events and clearing the active set.

### DIFF
--- a/pegainfer-deepseek-v4/src/direct/scheduler.rs
+++ b/pegainfer-deepseek-v4/src/direct/scheduler.rs
@@ -459,25 +459,25 @@ impl DeepSeekV4DirectGenerator {
             max_new_tokens,
             Some(slot_id),
         )?;
-        self.kv_cache.attach_prepared(&kv_cache)?;
-        let next_logits = run_prefill_logits_and_seed_decode_cache(
-            &mut self.runtime,
-            self.config,
-            prompt_tokens,
-        )?;
-        if slot_id != 0 {
-            clone_direct_decode_cache_slot(&mut self.runtime, 0, slot_id)?;
-        }
-        Ok(DeepSeekV4RequestState {
+        start_greedy_request_from_reserved_slot(
+            &mut self.kv_cache,
+            kv_cache,
             request_epoch,
-            kv_cache: Some(kv_cache),
-            prompt_len: prompt_tokens.len(),
+            prompt_tokens.len(),
             max_new_tokens,
             ignore_eos,
-            generated: Vec::with_capacity(max_new_tokens),
-            next_logits: Some(next_logits),
-            finish_reason: None,
-        })
+            || {
+                let next_logits = run_prefill_logits_and_seed_decode_cache(
+                    &mut self.runtime,
+                    self.config,
+                    prompt_tokens,
+                )?;
+                if slot_id != 0 {
+                    clone_direct_decode_cache_slot(&mut self.runtime, 0, slot_id)?;
+                }
+                Ok(next_logits)
+            },
+        )
     }
 
     #[cfg(test)]
@@ -835,6 +835,51 @@ impl DirectKvCacheManager {
         }
         .into())
     }
+}
+
+fn start_greedy_request_from_reserved_slot<F>(
+    kv_cache_manager: &mut DirectKvCacheManager,
+    kv_cache: DirectKvCacheLease,
+    request_epoch: u64,
+    prompt_len: usize,
+    max_new_tokens: usize,
+    ignore_eos: bool,
+    seed_next_logits: F,
+) -> Result<DeepSeekV4RequestState>
+where
+    F: FnOnce() -> Result<Vec<f32>>,
+{
+    if let Err(err) = kv_cache_manager.attach_prepared(&kv_cache) {
+        if let Err(release_err) = kv_cache_manager.release(&kv_cache) {
+            warn!(
+                "failed to release DeepSeek V4 KV cache after slot cache attach error: {release_err:#}"
+            );
+        }
+        return Err(err);
+    }
+
+    let next_logits = match seed_next_logits() {
+        Ok(next_logits) => next_logits,
+        Err(err) => {
+            if let Err(release_err) = kv_cache_manager.release(&kv_cache) {
+                warn!(
+                    "failed to release DeepSeek V4 KV cache after slot request start error: {release_err:#}"
+                );
+            }
+            return Err(err);
+        }
+    };
+
+    Ok(DeepSeekV4RequestState {
+        request_epoch,
+        kv_cache: Some(kv_cache),
+        prompt_len,
+        max_new_tokens,
+        ignore_eos,
+        generated: Vec::with_capacity(max_new_tokens),
+        next_logits: Some(next_logits),
+        finish_reason: None,
+    })
 }
 
 fn release_greedy_request_from(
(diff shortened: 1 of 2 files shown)


xiaguan:
Blocking review item on PR #107:

`start_greedy_request_in_slot` reserves and attaches a slot-owned KV lease before prefill/slot-clone completes, but the new active-set admission path does not release that lease on the error paths before it returns a `DeepSeekV4RequestState`.

Concretely, after `reserve_in_slot` + `attach_prepared`, both of these can still fail:

- `run_prefill_logits_and_seed_decode_cache(prompt_tokens)`
- `clone_direct_decode_cache_slot(0, slot_id)`

If either fails, the caller never receives the state, so `handle_request_wave` cannot release it through its `active` vector cleanup. The manager is left with `active_count > 0`, which then blocks later `set_request_slots(...)` / reservation and violates the slot-owned lease contract this PR is establishing. This is the same lifecycle class we guarded earlier for decode runtime errors: once ownership is visible to `DirectKvCacheManager`, every admission failure must unwind it explicitly.

Please mirror the single-request `start_greedy_request` cleanup shape here: after reservation, every attach/prefill/clone failure should release the exact `DirectKvCacheLease` before returning the error. A regression covering “active-set admission fails after attach and leaves no active slot / next reservation is not blocked” would make this boundary much harder to regress.

xiaguan:
Fixed the slot-start cleanup blocker in head `39f7da0`.

What changed:
- `start_greedy_request_in_slot` now routes post-reservation startup through a shared helper that releases the reserved slot if attach, prefill, or slot clone fails before returning a request state.
- Added `slot_start_error_releases_only_failed_slot`: it reserves two slots, keeps slot 0 active, injects a synthetic start failure on slot 1 after reservation, verifies active_count drops back to 1 with slot 0 intact, then reserves slot 1 again successfully.
- PR body validation updated to head `39f7da0`.

Validation:
- `cargo fmt --check` passed.
- `git diff --check` passed.
- `cargo check -p pegainfer-deepseek-v4 --features deepseek-v4` passed.
- `cargo test -p pegainfer-deepseek-v4 --features deepseek-v4 direct::scheduler::tests -- --nocapture` passed locally: `6 passed; 0 failed; 2 ignored`.
- 5090 active-set ignored test passed: `1 passed; 0 failed; finished in 37.11s`.
- 5090 default bs=1 exact E2E passed: `All 20 DeepSeek V4 exact cases passed`.
- `pre-commit run --files pegainfer-deepseek-v4/src/direct/scheduler.rs` is still blocked by the repository's existing invalid `.pre-commit-config.yaml` (`repo='builtin'` missing `rev`); no hook was bypassed.

xiaguan:
Second blocking item on the same ownership boundary:

`handle_request_wave` still clears the active set after `decode_greedy_batch_step_from_steps` returns `Err` without releasing the remaining request states:

```rust
if let Err(err) = generator.decode_greedy_batch_step_from_steps(&mut pairs) {
    for active_req in &mut active {
        send_request_error(...);
    }
    active.clear();
    break;
}
```

The batch helper only releases one subset of error exits today: the `decode_next_logits(...)` closure returning `Err`. Other error exits can still happen after ownership is active and before the caller clears the vector, for example stale/provenance mismatch, missing consumed logits, or a runtime row-count mismatch after `run_direct_decode_batch_logits` returns the wrong number of rows. In those cases the scheduler drops the states while `DirectKvCacheManager` still has active slot leases.

The scheduler boundary should be fail-closed: on any batch decode error, call `release_greedy_request` for every active request before `active.clear()`. That cleanup is safe even when the helper already released a state, because `release_greedy_request` takes the optional lease and becomes a no-op after the first release.

Please add a regression for an error path not covered by the helper's current decode-closure `Err` branch, such as a synthetic batch decode returning the wrong row count, and assert the active slot count returns to zero / later reservation is not blocked.

xiaguan:
Fixed the second batch-error cleanup blocker in head `33c00c0`.

What changed:
- `handle_request_wave` now treats any `decode_greedy_batch_step_from_steps` error as a scheduler-boundary cleanup point: it releases every active request state, then sends Error events and clears the active set.
- `advance_greedy_batch_steps_with_decode` now also releases batch states on provenance errors, decode closure errors, and row-count mismatch errors. Already-released states are no-op through `release_greedy_request_from`.
- Added `batch_row_count_error_releases_all_active_slots`: two active slots enter a batch helper path, the synthetic decode returns one row for two requests after both states have consumed logits/pushed tokens, and the test verifies both leases are released and a slot can be reserved again.
- PR body validation updated to head `33c00c0`.

Validation:
- `cargo fmt --check` passed.
- `git diff --check` passed.
- `cargo check -p pegainfer-deepseek-v4 --features deepseek-v4` passed.
- `cargo test -p pegainfer-deepseek-v4 --features deepseek-v4 direct::scheduler::tests -- --nocapture` passed locally: `7 passed; 0 failed; 2 ignored`.
- 5090 active-set ignored test passed: `1 passed; 0 failed; finished in 37.05s`.
- 5090 default bs=1 exact E2E passed: `All 20 DeepSeek V4 exact cases passed`.
- `pre-commit run --files pegainfer-deepseek-v4/src/direct/scheduler.rs` is still blocked by the repository's existing invalid `.pre-commit-config.yaml` (`repo='builtin'` missing `rev`); no hook was bypassed.

xiaguan:
LGTM at head `33c00c0395b12a177e0c20ad34478edf9ac0ede1`; merged as `65e85f23587b3d8bec5f2f5f412fcf9c6a2e398f`.

Review focus closed:
- slot-owned lease now lives in `DirectKvCacheManager`, with per-slot reserve/attach/release and `active_count` visibility;
- default single-request path stays on one slot and the existing `generate_greedy` flow;
- active-set path drains at most two requests, seeds separate slots, calls PR #105 `DecodeBatch`, and releases requests independently;
- both blocker classes are fixed: admission/start failure releases the failed slot, and batch decode errors release active states before clearing them.

Local review validation:
- `git diff --check github/main...HEAD` passed.
- `cargo fmt --check` passed.
- Targeted scheduler tests could not run on my review host because the existing `pegainfer-kernels` build script requires TileLang at build time and this host does not have it. I used the posted local and 5090 evidence for runtime validation.

Scope remains PR B only: scheduler runtime-batch plumbing for bs=2, not HTTP serving, perf, larger batch, paged/prefix KV, or P/D.

PR #108, opened 2026-05-13
docs(dsv4): add HTTP serving benchmark gate

xiaguan:
## Summary

Adds a stdlib OpenAI-compatible HTTP streaming benchmark client for `/v1/completions` and documents the DSV4 HTTP serving benchmark gate for task #18.

This is a benchmark gate only. It does not change runtime behavior, tune performance, add paged/prefix KV, add P/D handoff, or claim production serving readiness.

## Benchmark Surface

`scripts/bench_http_serving.py` records:

- QPS plus completed / failed / timeout counts.
- TTFT from request send to first streamed text chunk.
- ITL and TPOT from streamed text chunks.
- End-to-end latency percentiles.
- Error and timeout rates.
- Per-request output hashes plus a combined output hash for stability checks.

The script targets the OpenAI-compatible HTTP endpoint directly. It does not use `bench_serving request` as serving evidence. When `max_tokens > 0`, a `200` streaming response with no text chunks is treated as a failed request so TTFT / TPOT / hash evidence cannot be satisfied by a `[DONE]`-only stream.

## Reproducible Commands

Build the server on the target GPU host:

```bash
cd /path/to/pegainfer
export PATH=/usr/local/cuda-13.1/bin:$PATH
export CUDA_HOME=/usr/local/cuda-13.1
export PEGAINFER_TILELANG_PYTHON=/path/to/venv/bin/python
export PEGAINFER_TRITON_PYTHON=/path/to/venv/bin/python
export PEGAINFER_NVCC_JOBS=8
export CARGO_TARGET_DIR=/path/to/pegainfer-target

cargo build --release -p pegainfer-server --features deepseek-v4 --bin pegainfer
```

Start the OpenAI-compatible endpoint:

```bash
$CARGO_TARGET_DIR/release/pegainfer \
  --model-path /data/DeepSeek-V4-Flash \
  --port 18118
```

Verify the endpoint:

```bash
curl -sS http://127.0.0.1:18118/v1/models
```

Run the HTTP benchmark:

```bash
python3 scripts/bench_http_serving.py \
  --base-url http://127.0.0.1:18118 \
  --model /data/DeepSeek-V4-Flash \
  --warmup 2 \
  --num-requests 8 \
  --concurrency 2 \
  --prompt-words 16 \
  --max-tokens 16 \
  --timeout 240 \
  --out /tmp/dsv4_http_bench_task18.json
```

## HTTP Evidence

| Field | Value |
| --- | --- |
[...]

PR #109, opened 2026-05-13
docs(dsv4): define prefix paged KV and P-D handoff contract

xiaguan:
## Summary

Adds the task #19 design contract for evolving DeepSeek V4 from today's direct slot-owned KV leases toward prefix cache, paged KV ownership, and transport-agnostic P-D handoff handles.

This is documentation only. It does not change runtime behavior, scheduler behavior, HTTP benchmark semantics, or `pegainfer-comm` APIs.

## Contract Shape

The document separates evidence layers explicitly:

- `code-fact`: current direct KV / slot lease behavior in `pegainfer-deepseek-v4` after PR #108.
- `user-input`: task #19 asks for a design contract, not a full runtime implementation.
- `derivation`: proposed interfaces and rollout boundaries inferred from the current code facts.

Main sections:

- current direct KV ownership facts (`DirectKvCacheManager`, `DirectKvCacheLease`, slot-owned release semantics);
- slot layout facts (`LayerDecodeCache`, compressed/indexer state, `DecodeBatchMeta` bases);
- page ownership and `KvLease` state model;
- prefix cache key/lifecycle and reject conditions;
- paged allocator contract and fail-closed allocation rules;
- eviction/reject policy;
- allocator telemetry contract;
- P-D export/import handle semantics, cancellation, cleanup ownership, and observable-signal checklist;
- scheduler boundary and next feature PR boundary.

## Boundaries

- Prefix/page/evict/capacity are kept allocator-owned, not scheduler-owned.
- P-D handoff defines ownership, cancel, cleanup, and observable poll signals only; it does not choose RDMA / IPC / serialization / transport wiring.
- Scheduler code should see allocator handles and poll states, not physical row offsets, DSV4 compressor/indexer math, or transport registration state.
- The recommended next feature PR is a single-node paged KV allocator contract + telemetry adapter behind the current direct runtime, not cross-node P-D.

## Validation

- `git diff --check` passed.
- `cargo fmt --check` passed.
- Private-path scan over changed docs passed for `/root`, `develop/xingming`, `PegaInfer-Dev`, `target-p`, `pegainfer-bs`, `/data/slock`, `host-192`, and `192.168`.
- `pre-commit run --files docs/projects/deepseek-v4/prefix-paged-kv-pd-handoff.md docs/index.md` remains blocked by the existing invalid `.pre-commit-config.yaml` (`repo='builtin'` missing `rev`); no hook was bypassed.

## Review Notes

@PegaInfer-Review: please review the full task #19 contract against the agreed gate: code-fact grounding, design-vs-runtime separation, scheduler boundary, and no premature transport detail.

@Pacer: please focus only on the P-D handoff section's transport boundary and owner / observable-signal checklist. Prefix/page/evict/capacity are intentionally not your review surface for this PR.


xiaguan:
Blocking design issue in the core ownership contract:

`KvLease` currently has a single `pages: KvPageSet` field, while the invariants say a request can hold both borrowed prefix pages and private decode pages. Those are not the same ownership relationship:

- private decode/suffix pages are request-owned and must be freed or transferred on release/cancel;
- borrowed prefix pages remain prefix-owned, are read-only for the request, and should be unpinned rather than freed;
- export/transfer-pinned pages have a third lifecycle and cleanup owner.

With all of these collapsed into one `KvPageSet`, the next feature PR cannot tell which pages `release_request` is allowed to free, which pages `cancel_request` should only unpin, and which pages an export handle is allowed to move into transfer ownership. That conflicts with the document's own invariant that a page has exactly one owner or one free-list membership, and it is exactly the kind of ambiguity that turns prefix cache into scheduler/runtime cleanup debt.

Please make the contract explicit before merging the design doc. A good shape would separate the relationships, for example:

- request-private mutable pages owned by the lease;
- borrowed prefix pins owned by prefix entries but referenced by the request;
- transfer/export pins or handles with their own owner/cleanup path;
- a `KvDecodeView` that composes borrowed prefix pages plus owned suffix/decode pages for runtime consumption.

Then state how `release_request`, `cancel_request`, `seal_prefix`, and `export` treat each category. This does not require runtime code in this PR, but the design contract needs that distinction so the first implementation PR does not encode the wrong ownership model.

xiaguan:
Updated PR #109 at head `eff32a60bc5d57913959087907e24afb8d86a02c`.

Closed the lease-cleanup blocker by splitting `KvLease` into three resource classes:

- `request_pages`: mutable request-owned pages; freed, sealed, or transferred by allocator operations.
- `prefix_pins`: borrowed read-only prefix pages; request cleanup unpins them and never frees them directly.
- `transfer_pins`: export/import pages held by transfer ownership; request cleanup cancels or delegates to transfer cleanup and never frees them directly.

Added operation semantics for:

- `release_request`: free request-owned pages, unpin prefix pins, cancel/delegate transfer pins.
- `cancel_request`: same resource-class cleanup plus cancellation counters and poisoned-page handling.
- `seal_prefix`: move a contiguous immutable subset from `request_pages` into a sealed prefix entry; borrowed prefix pins remain borrowed.
- `export_request_pages`: move exported request pages into `KvTransferPin` and `KvPageOwner::Transfer` until terminal export state.

Also incorporated the P-D scoped review gaps:

- timeout is `Failed(KvTransferError::Timeout)`;
- source cancel during pending export has a terminal poll / cleanup path;
- ready descriptors have explicit TTL/expiry handling;
- cancel and fail cleanup obligations are split, with failed staging pages recorded before return;
- transfer descriptors are single-consume and failed import retry requires a new descriptor or explicit re-export;
- `ImportStaging` state records target-side staging ownership before import readiness.

Re-validation:

- `git diff --check`
- `cargo fmt --check`
- private-path scan over changed docs
- `pre-commit run --files docs/projects/deepseek-v4/prefix-paged-kv-pd-handoff.md docs/index.md` remains blocked by the existing invalid `.pre-commit-config.yaml` (`repo='builtin'` missing `rev`); no hook was bypassed.

xiaguan:
### P-D handoff section — scoped review

Scoped to the **P-D Handoff Handle Contract** chapter only. Two parts.

#### 1. Transport boundary — no over-step

- Opening derivation explicitly states the handle describes ownership/cleanup and does not choose RDMA, IPC, serialization, or a specific `pegainfer-comm` operation.
- `KvExportHandle` / `KvImportHandle` carry only allocator-side fields (`request_id`, `layout`, `pages`, `epoch`).
- `KvTransferDescriptor` and `KvTransferError` are left opaque — their shape is intentionally not defined in this contract.
- Closing line re-locks: transport implementation can later map signals to RDMA / CUDA IPC / serialized path; scheduler code only sees allocator handles and poll results.

The boundary holds. No transport object, MR handle, QP, remote address, or collective-group concept leaks into the contract.

#### 2. Owner / observable-signal checklist — 5 gaps to close

a. **Timeout terminal state missing from enum.** The export prose mentions "until completion, cancellation, or timeout", but `KvExportPoll` / `KvImportPoll` only enumerate `Pending / Ready / Cancelled / Failed`. Suggest folding timeout into `Failed(KvTransferError::Timeout)` and adding a corresponding row in the checklist, or making it a distinct variant. Current state machine is ambiguous on this terminal.

b. **Source-cancel-during-export-pending API behavior unspecified.** The invariant "source request cannot release exported pages while export is pending" forbids the state, but does not say what `cancel_request(source_lease)` returns at the API boundary — block, reject with `TransferInFlight`, or queue until export reaches a terminal state. Suggest a row: *Source cancel during export pending → owner = transfer manager, signal = `KvExportPoll::Cancelled`, cleanup = source allocator after poll terminal.*

c. **Descriptor lifetime when `Ready` but never consumed.** The "Export ready" cleanup column reads "source or transfer manager per descriptor lifetime", but "per descriptor lifetime" is not defined in the contract. If a descriptor is published `Ready` and no importer claims it, source pages hang. Suggest a row: *Descriptor abandoned (Ready, no importer within TTL) → owner = transfer manager, signal = descriptor expiry, cleanup = source allocator.*

d. **`Cancelled` vs `Failed` cleanup symmetry not stated.** Both terminal states are collapsed into one checklist row ("Cancel before ready") with cleanup written as "phase owner named above". `Cancelled` is a caller-initiated abort; `Failed` is a transport/runtime fault that may leave half-committed side effects. State explicitly whether cleanup obligations are identical, or split the row.

e. **Retry / single-use semantics on descriptors.** After `KvImportPoll::Failed`, can the same descriptor be re-imported, or is it consumed? Single-use is the safer default for ownership reasoning (source pages return to source on Failed; descriptor is invalidated). State the invariant either way.

One observation, not an ask: `KvLeaseState` has `Imported` but no explicit pre-import staging variant — naming where the target-side staging pages live during `Pending → Ready` would tighten the "Import in flight" cleanup column, but page-state naming sits adjacent to paged-allocator territory, so flagging for awareness.

Transport boundary: pass. After (a)–(e) land, the P-D contract can directly drive the wiring PR's ownership tests.

xiaguan:
### P-D handoff section — re-review at head `eff32a6`

All five checklist gaps and the `ImportStaging` observation are closed at the contract level:

- (a) **Timeout terminal state** — folded into `Failed(KvTransferError::Timeout)` (export invariant) and reflected in the checklist row for source-cancel-during-export. No separate `Timeout` enum variant; state machine no longer ambiguous.
- (b) **Source-cancel-during-export-pending** — new checklist row "Source cancel during export pending → owner = transfer manager, signal = `KvExportPoll::Cancelled` or `Failed(Timeout)`, cleanup = source allocator after terminal poll". `cancel_request` cleanup semantics also clarify the API path (cancels pending transfer pins owned by the request, leaves completed transfer cleanup to the transfer manager).
- (c) **Descriptor Ready-but-unconsumed** — export invariants add "`KvTransferDescriptor` is single-consume" and "descriptor lifetime must have an explicit TTL or equivalent expiry signal"; new checklist row "Descriptor abandoned → owner = transfer manager, signal = descriptor expiry / TTL elapsed, cleanup = source allocator".
- (d) **Cancel vs Fail cleanup symmetry** — import invariants now state explicitly "`Failed` and `Cancelled` have the same cleanup obligation for target staging pages; `Failed` additionally records the failure reason and any poisoned pages before returning them to the allocator"; the checklist splits "Cancel before ready" and "Fail before ready" into separate rows.
- (e) **Retry / single-use** — locked in both directions: export side declares single-consume; import side declares failed imports do not retry through the same descriptor and require a new export descriptor or explicit source-side re-export.
- **`ImportStaging` observation** — new `KvLeaseState::ImportStaging` variant plus import invariant "target-side staging pages are represented by `ImportStaging` and are cleaned by the target allocator until the import becomes `Ready`". The "Import in flight" cleanup column now has a concrete state to bind to.

Transport boundary check holds: `KvTransferDescriptor` / `KvTransferError` remain opaque; the new `KvTransferDirection { Export, Import }` is an allocator-side pin direction, not a transport mechanism; `export_request_pages` is an allocator method. Opening and closing scope-locks of the chapter are unchanged.

P-D handoff chapter at `eff32a6`: pass on the scoped-review surface.

xiaguan:
One remaining contract mismatch before merge:

`KvPageAllocator::seal_prefix` is declared as:

```rust
fn seal_prefix(&mut self, lease: &KvLease, key: PrefixKey) -> Result<PrefixEntryId>;
```

but the operation contract below says `seal_prefix` removes pages from `lease.request_pages` and records the resulting `PrefixEntryId`. That is a lease mutation, and the rest of the cleanup contract now correctly depends on that mutation being explicit.

Please make the signature match the stated post-condition. Preferred shape for this contract is:

```rust
fn seal_prefix(&mut self, lease: &mut KvLease, key: PrefixKey) -> Result<PrefixEntryId>;
```

Alternatively the method can return an updated lease/delta, but the document should choose one explicit ownership update path. Without this, the next feature PR can copy a signature that hides the most important state transition in the allocator.

xiaguan:
Updated PR #109 at head `546654cebb930670e3980884024ee6a0d7493e42`.

Closed the remaining interface blocker by changing the allocator contract from:

```rust
fn seal_prefix(&mut self, lease: &KvLease, key: PrefixKey) -> Result<PrefixEntryId>;
```

to:

```rust
fn seal_prefix(&mut self, lease: &mut KvLease, key: PrefixKey) -> Result<PrefixEntryId>;
```

That matches the documented state transition: `seal_prefix` moves a contiguous immutable subset out of `lease.request_pages`, records the resulting `PrefixEntryId`, and leaves `prefix_pins` untouched for request cleanup.

Re-validation:

- `git diff --check`
- `cargo fmt --check`
- private-path scan over changed docs
- `pre-commit run --files docs/projects/deepseek-v4/prefix-paged-kv-pd-handoff.md docs/index.md` remains blocked by the existing invalid `.pre-commit-config.yaml` (`repo='builtin'` missing `rev`); no hook was bypassed.

xiaguan:
LGTM at head `546654cebb930670e3980884024ee6a0d7493e42`.

Review closure:

- The main ownership blocker is closed: `KvLease` now separates `request_pages`, `prefix_pins`, and `transfer_pins`, and the cleanup semantics for `release_request`, `cancel_request`, `seal_prefix`, and `export_request_pages` are explicit by resource class.
- The follow-up signature blocker is closed: `seal_prefix` now takes `&mut KvLease`, matching the documented state transition that removes pages from `request_pages`.
- P-D handoff scoped review is closed by @Pacer: timeout, source cancel during pending export, descriptor TTL/abandon, Cancel vs Failed cleanup, single-consume semantics, and `ImportStaging` are now covered while transport details remain opaque.

Local review checks passed:

- `cargo fmt --check`
- private-path scan over changed docs

Scope remains correct for task #19: design contract only, no runtime behavior change, no HTTP benchmark change, no transport implementation claim.

PR #111, opened 2026-05-14
bench(dsv4): profile HTTP TTFT and prefill path

xiaguan:
## Summary

Adds the P1 DSV4 HTTP TTFT trace + prefill profiling gate after the P0 concurrency correctness fix.

- threads request id and engine queue timestamp through the HTTP bridge into `GenerateRequest`
- emits `pegainfer_http_trace` records from the DSV4 direct scheduler with queue, prefill, first-decode, and first-token timing
- adds `--deepseek-prefill-profile`, which emits rank-0 `pegainfer_prefill_profile` phase records for DSV4 prefill
- extends `scripts/bench_http_serving.py` to attach server traces and report phase summaries
- extends `scripts/bench_http_sweep.py` to sweep concurrency, max_tokens, and prompt length with per-request hash stability as the correctness gate
- updates the HTTP serving benchmark doc with concurrency, prompt-length, and prefill-profile evidence

## Scope

This PR is diagnostic / benchmark infrastructure plus prefill phase attribution. It does not tune performance, re-enable HTTP wave batching, change paged/prefix KV, or claim serving parity. Hash stability remains the gate: any drift fails the sweep before interpreting TTFT or throughput.

The important interpretation change from the initial PR body: admission queue is a concurrency amplification term, not the root cause for low throughput. The current single-request service time is already high: short-prompt c1 TTFT is about `178ms`, and the remaining decode path is about `29ms/token`. The next optimization target is prefill/decode service time, with priority on the DSV4 prefill kernel/runtime path.

A candidate ratio-4 indexer compressed-KV reuse was tested and rejected because it changed the HTTP output hash gate. It is not included in this PR.

## Evidence

Commit: `1be57cc6d5af4d2cd31c125a4df16b39dc2b348b`

Local checks:

- `git diff --check` passed
- `cargo fmt --check` passed
- `python3 -m py_compile scripts/bench_http_serving.py scripts/bench_http_sweep.py` passed
- `python3 -m unittest tests/test_bench_http_serving.py` passed

Remote build checks on the validation host:

- `cargo check -p pegainfer-server --features deepseek-v4` passed
- `cargo build --release -p pegainfer-server --features deepseek-v4 --bin pegainfer` passed
- `/v1/models` smoke passed against the release server

HTTP correctness / concurrency sweep command:

```bash
python3 scripts/bench_http_sweep.py \
  --base-url http://127.0.0.1:18118 \
  --model /data/DeepSeek-V4-Flash \
  --warmup 2 \
[...]

PR #112, opened 2026-05-14
perf(dsv4): reuse prefill window topk

xiaguan:
## Summary

Optimizes DSV4 direct prefill by reusing the prefill window top-k index table across layers.

The table depends only on `(seq_len, sliding_window)`, but the previous rank-lane prefill path regenerated it inside each sparse / ratio-4 attention layer. This PR builds it once per prefill request and passes it through the block/attention path. It does not reuse or mutate compressed KV values; the ratio-4 indexer/compressor math stays unchanged.

## Scope

- first ratio-4 `block_prefill` optimization after PR #111 profiling
- no HTTP wave batching change
- no paged/prefix KV or P/D changes
- no serving parity or production claim

## Evidence

Commit: `8a5886ea164e04bf247255dae2afc2b858c1705f`

Local checks:

- `git diff --check` passed
- `cargo fmt --check` passed
- `python3 -m py_compile scripts/bench_http_serving.py scripts/bench_http_sweep.py` passed
- `python3 -m unittest tests/test_bench_http_serving.py` passed

Build checks on the validation host:

- `cargo check -p pegainfer-server --features deepseek-v4` passed
- `cargo build --release -p pegainfer-server --features deepseek-v4 --bin pegainfer` passed
- `/v1/models` smoke passed against the release server

HTTP correctness sweep, non-profile server:

```bash
python3 scripts/bench_http_sweep.py \
  --base-url http://127.0.0.1:18118 \
  --model /data/DeepSeek-V4-Flash \
  --warmup 2 \
  --num-requests 8 \
  --concurrency 1,2,4,8 \
  --max-tokens 16 \
  --repeats 3 \
  --prompt-words 16 \
  --timeout 240 \
  --server-log /tmp/dsv4_http_server_task5_opt.log \
  --out-dir /tmp/dsv4_http_sweep_task5_opt
```

All cells passed the correctness gate: failed `0`, timeout `0`, combined hash `22706877075acde0` in all three repeats.

| Concurrency | QPS (3 runs) | TTFT avg ms (3 runs) | TPOT avg ms (3 runs) |
| --- | --- | --- | --- |
| `1` | `1.737`, `1.730`, `1.744` | `156.5`, `154.6`, `153.2` | `27.92`, `28.20`, `28.00` |
| `2` | `1.746`, `1.748`, `1.747` | `653.3`, `652.9`, `653.9` | `28.01`, `27.98`, `27.97` |
| `4` | `1.746`, `1.746`, `1.748` | `1441.7`, `1442.6`, `1441.2` | `27.97`, `28.00`, `27.98` |
| `8` | `1.745`, `1.746`, `1.747` | `2162.5`, `2159.2`, `2157.3` | `28.01`, `27.97`, `27.99` |

Profile-mode evidence uses `--deepseek-prefill-profile`, which synchronizes CUDA phases and is not a serving-number source. It is used only to compare the ratio-4 `block_prefill` bucket against the PR #111 profile baseline.

| Prompt tokens | PR #111 ratio-4 `block_prefill` ms | This PR ratio-4 `block_prefill` ms | Delta |
| ---: | ---: | ---: | ---: |
| `661` | `200.7` | `207.3` | noisy / no win |
| `2645` | `818.6` | `815.4` | `-3.2ms` |
| `10580` | `6902.4` | `6729.9` | `-172.5ms` |

The clearest gain is at the long-prompt point where repeated window-index generation was most expensive. Shorter prompt profile points remain noisy, so this PR does not claim a broad TTFT improvement beyond the measured long-prompt ratio-4 bucket reduction and the non-profile HTTP correctness/QPS evidence above.

### DIFF
--- a/docs/projects/deepseek-v4/http-serving-benchmark.md
+++ b/docs/projects/deepseek-v4/http-serving-benchmark.md
@@ -209,6 +209,36 @@ the compressed attention / indexer / compressor path. A candidate reuse of the
 ratio-4 indexer compressed KV was evaluated and rejected because it changed the
 HTTP output hash gate; it is not included in this PR.
 
+### P2 Ratio-4 Block-Prefill Optimization
+
+The first accepted optimization removes repeated generation of the same prefill
+window top-k index table from the layer loop. The table depends only on
+`seq_len` and `sliding_window`, not on the layer. The direct prefill path now
+builds it once per request and reuses it for non-compressed sparse attention and
+ratio-4 compressed attention layers.
+
+HTTP correctness remains the gate. The same c1/c2/c4/c8 sweep still passes with
+`failed=0`, `timeout=0`, and combined hash `22706877075acde0` across all three
+repeats:
+
+| Concurrency | QPS (3 runs) | TTFT avg ms (3 runs) | TPOT avg ms (3 runs) |
+| --- | --- | --- | --- |
+| `1` | `1.737`, `1.730`, `1.744` | `156.5`, `154.6`, `153.2` | `27.92`, `28.20`, `28.00` |
+| `2` | `1.746`, `1.748`, `1.747` | `653.3`, `652.9`, `653.9` | `28.01`, `27.98`, `27.97` |
+| `4` | `1.746`, `1.746`, `1.748` | `1441.7`, `1442.6`, `1441.2` | `27.97`, `28.00`, `27.98` |
+| `8` | `1.745`, `1.746`, `1.747` | `2162.5`, `2159.2`, `2157.3` | `28.01`, `27.97`, `27.99` |
+
+Rank-0 profile shows the clearest long-prompt improvement in the ratio-4 bucket:
+
+| Prompt tokens | P1 ratio-4 `block_prefill` ms | P2 ratio-4 `block_prefill` ms | Delta |
+| ---: | ---: | ---: | ---: |
+| `661` | `200.7` | `207.3` | noisy / no win |
+| `2645` | `818.6` | `815.4` | `-3.2ms` |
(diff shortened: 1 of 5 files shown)


xiaguan:
Review pass. This matches the task #5 boundary: one real prefill-path optimization, no wave-batching change, no compressed-KV value reuse, and no broad serving-parity claim.

What I checked:

- The cached `PrefillWindowTopk` depends only on `seq_len` and `sliding_window`; call sites pass it only into the prefill sparse-window paths where the old code generated the same `window_topk_indices` per layer.
- Ratio-4 still regenerates the compressed/indexer top-k from scores per layer; only the window-index half is reused, so the rejected compressed-KV reuse from PR #111 is not back in the tree.
- The shape guard in `PrefillWindowTopk::parts()` catches future accidental reuse across a different sequence/window shape.
- PR body keeps the right evidence split: non-profile HTTP sweep is the serving correctness/latency evidence; profile-mode numbers are attribution only.
- The doc does not turn the short-prompt noisy profile rows into a broad claim. The actual claim is limited to the long-prompt ratio-4 bucket reduction plus clean HTTP hash evidence.

Local checks on `8a5886e`:

- `git diff --check github/main...HEAD` passed
- `cargo fmt --check` passed
- `python3 -m py_compile scripts/bench_http_serving.py scripts/bench_http_sweep.py` passed
- `python3 -m unittest tests/test_bench_http_serving.py -v` passed
- changed-file private-path scan passed

`cargo check -p pegainfer-deepseek-v4 --features deepseek-v4` is still blocked on my review box by missing TileLang/Triton build-time Python modules, so I’m relying on the validation-host build and HTTP/profile evidence in the PR body.

Merging as the first ratio-4 block-prefill optimization.

PR #136, opened 2026-05-17
feat(sim): add CPU-only simulated engine

### LINKED ISSUE #125: Simulated inference engine for server benchmarking and routing testbeds
## Why

Right now we can't measure the server itself in isolation. Every benchmark loads a real model, so the server's overhead (HTTP, vLLM frontend, SSE, scheduler, IPC bridge) gets buried under GPU computation. The TCP_NODELAY incident — where axum's default cost us ~30ms per request — is exactly the kind of bug a real-model benchmark hides.

Longer term, the same primitive unlocks **multi-node routing testbeds**: spin up N fake engines with configurable latency profiles and let a router/load-balancer drive them. No GPU, no model weights, no cluster.

This is hard because the abstraction needs to be honest enough to expose real bugs, but the value is high enough that it's worth doing.

## What we want

A "simulated inference engine" that:

- Serves the same HTTP API as a real engine (OpenAI `/v1/completions`, streaming + non-streaming).
- Generates fake but configurable output tokens with realistic timing.
- Requires no GPU and no model weights.

Everything else — crate vs. feature flag, configuration surface, code layout — is up to the implementer.

## Suggested timing model (not prescriptive)

Simple is fine. Examples:

- **TTFT** = `prompt_len / prefill_throughput`, where `prefill_throughput` defaults to something like `100 tokens/ms`.
- **TPOT** = `base_ms * (1 + Pareto(α))`, base ~10–15ms, α controlling tail thickness. A power-law tail mimics real engines' batch-switch / GC jitter better than a fixed delay.

You can pick something simpler or more elaborate — the goal is configurable timing that looks roughly like a real engine, not a physically accurate simulator.

## Acceptance criteria

The PR must demonstrate the simulated engine works end-to-end through the existing HTTP stack:

1. Start the simulated engine on a port (e.g. 8000).
2. Drive it with `vllm bench serve`, e.g.:
   ```
   vllm bench serve \
     --backend openai \
     --model <simulated-model-id> \
     --dataset-name random \
     --input-len 1024 --output-len 128 \
     --num-prompts 50 --max-concurrency 4 \
     --ignore-eos --temperature 0
   ```
3. **Attach a screenshot of the `vllm bench serve` results table** to the PR. It must complete without errors and show non-zero TTFT/TPOT/throughput consistent with the configured timing model.

That's the bar. Anything beyond that (configuration knobs, routing-testbed scaffolding, observability hooks) is bonus.

## Open questions for the implementer

- How to expose the configuration (CLI flag, YAML, env vars)?
- Does the tokenizer requirement of `vllm bench serve` cause friction? If so, what's the cleanest workaround?
- Should batching behavior be simulated (shared prefill across a batch), or is independent per-request timing enough for the first cut?

Discussion welcome in the comments before opening a PR.

CAICAIIs:
## Description

Fixes #125.

Adds a CPU-only simulated model crate (`pegainfer-sim`) that serves through the existing vLLM/OpenAI frontend with configurable TTFT/TPOT timing. This lets `vllm bench serve` validate the frontend path without CUDA, real model weights, or real-model performance claims.

Also splits the lightweight engine request/event contract into `pegainfer-engine` and moves the vLLM bridge into `pegainfer-vllm-frontend`, while keeping existing `pegainfer-core` and `pegainfer-server` re-exports for compatibility.

Validation:
- `git diff --check`
- `cargo fmt --check`
- `cargo test --release -p pegainfer-engine`
- `cargo test --release -p pegainfer-vllm-frontend`
- `cargo test --release -p pegainfer-sim`
- `cargo clippy --release -p pegainfer-vllm-frontend --all-targets -- -D warnings`
- `cargo clippy --release -p pegainfer-sim --all-targets -- -D warnings`
- `cargo build --release -p pegainfer-sim`
- Local HTTP smoke for `/v1/models`, non-streaming `/v1/completions`, and streaming `/v1/completions`
- Remote `vllm bench serve`: 50 successful requests, 0 failed requests

## Type of Change

- [x] Documentation update

## Checklist


docs/subsystems/frontend/simulated-inference-engine.md
```diff
@@ -0,0 +1,42 @@
...
+
+`pegainfer-sim` exposes CLI knobs for model id, port, max model length, base TTFT, prefill throughput, TPOT, and fallback token id.
+
+The timing model is intentionally simple: TTFT is `base_ttft_ms + prompt_len / prefill_tokens_per_ms`, and TPOT is a fixed delay between generated tokens. Output token ids cycle through the prompt tokens, using the fallback id for empty prompts.
+
+The frontend still needs tokenizer/model metadata, but the simulator never loads model weights.
+
+## Evidence
+
+- Format/metadata gates pass: `git diff --check`, `cargo fmt --check`, and `cargo metadata --no-deps --format-version 1`.
```
xiaguan: We could remove this section and document some future plans or implementation details instead

xiaguan (commented):
After this PR, we could add some e2e tests from the frontend. Feel free to open an issue to track this.

xiaguan:
Btw, we could also open another issue to profile the frontend (CPU) and make sure it performs well

CAICAIIs:
thx, done

xiaguan (approved):

PR #146, opened 2026-05-18
test(dsv4): cover real indexer topk shape

xiaguan:
## Summary

- update the ignored DSV4 prefill indexer top-k 10k gate to use the real runtime `index_topk=512`
- document the real 10k top-k shape source as DSV4 config plus `indexer_topk_indices_prefill` using `config.index_topk.min(compressed_len)`
- mark old topk32 odd / tie-heavy / 10k-like gates as synthetic or historical task-gate evidence, not future runtime perf-gate sources
- correct the 10k top-k shared-memory entry to the current launcher formula: `(compressed_len + 256) * sizeof(float) + 256 * sizeof(int) = 12628 bytes`

## Scope

- test + docs contract correction only
- no runtime kernel, Rust, FFI, launcher, HTTP, or profile changes

## Validation

- `cargo fmt --check`
- `git diff --check`
- `pre-commit run --files pegainfer-kernels/tests/deepseek_indexer_topk.rs docs/models/deepseek-v4/http-serving-benchmark.md`
- after the docs-only shared-memory fix: `pre-commit run --files docs/models/deepseek-v4/http-serving-benchmark.md`
- 5090: `cargo test -r -p pegainfer-kernels --features deepseek-v4 --test deepseek_indexer_topk -- --ignored --nocapture`
  - `indexer_topk_prefill_matches_reference_odd_shape` PASS
  - `indexer_topk_prefill_matches_reference_10k_real_index_topk_shape` PASS (`seq_len=10580, compressed_len=2645, topk=512, ratio=4, offset=10580`)
  - `indexer_topk_prefill_matches_reference_tie_heavy_rows` PASS

### DIFF
--- a/docs/models/deepseek-v4/http-serving-benchmark.md
+++ b/docs/models/deepseek-v4/http-serving-benchmark.md
@@ -330,17 +330,20 @@ The current code path is:
 - output indices are `offset + compressed_idx`, where the ratio-4 prefill path
   uses the raw window length as `offset`.
 
-For the 10k prefill workload from the P3 observation, this gives the concrete
-shape:
+For the 10k prefill workload from the P3 observation, the runtime shape source
+is the DSV4 model config (`index_topk=512`) combined with
+`indexer_topk_indices_prefill`, which passes
+`config.index_topk.min(compressed_len)` to
+`deepseek_indexer_topk_prefill_cuda`. That gives the concrete runtime shape:
 
 | Field | Value |
 | --- | --- |
 | `seq_len` | `10580` |
 | `compressed_len` | `2645` |
-| `topk` | `32` |
+| `topk` | `512` |
 | `ratio` | `4` |
 | `offset` | `10580` |
-| shared memory per block | `10580` bytes |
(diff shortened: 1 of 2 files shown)


xiaguan:
One small correction before merge: the shape source is fixed, but the same table still carries the old/inaccurate shared-memory number.

`docs/models/deepseek-v4/http-serving-benchmark.md` now correctly says the 10k runtime top-k shape is `topk=512`, but the row still says `shared memory per block = 10580 bytes`. The launcher computes shared memory as:

```cpp
(compressed_len + threads) * sizeof(float) + threads * sizeof(int)
```

with `compressed_len=2645` and `threads=256`, so the value is `12628` bytes. Since this section is explicitly being promoted back into the canonical runtime shape source, leaving a stale derived number in the same table is exactly the kind of footgun this PR is supposed to remove.

Please update that row. The rest of the PR scope looks right: test/doc contract only, real `index_topk=512`, synthetic topk32 cases labeled as non-runtime, and no runtime code changes.

PR #145, opened 2026-05-17
perf(dsv4): cuBLAS-based non-overlap prefill compressor

xiaguan:
## Summary

Replace the hand-rolled `deepseek_compressor_nonoverlap_weighted_kernel` implementation of `deepseek_compressor_nonoverlap_prefill_cuda` with a two-cuBLAS-GEMM + fused-epilogue structure mirroring the overlap path introduced in PR #140. The previous kernel iterated each `(compressed, head_dim)` output element across `ratio` source tokens with a thread-cooperative GEMV pair, redundantly reloading `Wgate` and `Wkv` and never reaching tensor cores. This bucket dominated the standalone CUDA cost on the 10k-token DSV4 prefill profile.

New flow inside the host wrapper:

1. `cudaMallocAsync` one FP32 scratch holding scores and values back to back (`2 * seq_len * head_dim` floats).
2. cuBLAS `GemmEx` (BF16 inputs, FP32 accumulator, tensor-core path) for `scores = X @ Wgate^T`.
3. cuBLAS `GemmEx` of the same shape for `values = X @ Wkv^T`.
4. New `deepseek_compressor_nonoverlap_fused_epilogue_kernel`: one block per compressed token, threads collaborate across `head_dim`. Each thread streams over `ratio` routes from the FP32 scratch twice — first pass finds `max(score)` for numerical stability, second pass accumulates the softmax denominator and weighted-value numerator together. A block-wide warp-shuffle reduction produces the RMS-norm scale; the same threads then write the BF16 output `out[c * head_dim + d]` using the per-dimension norm weight. The streaming softmax leaves `ratio` uncapped (DSV4 actually uses `ratio` up to 128).
5. `cudaFreeAsync` the scratch.

`seq_len` does not need to be a multiple of `ratio`: the GEMMs run over the full input, the epilogue reads only `compressed_len * ratio` tokens, and any trailing partial group is ignored — matches the pre-cuBLAS contract that online prompts of arbitrary prefill length rely on.

The FFI signature (`deepseek_compressor_nonoverlap_prefill_cuda`) and all Rust call sites are unchanged. `deepseek_compressor_norm_serial_kernel` is preserved because the overlap prefill path still uses it. The hand-rolled `_nonoverlap_weighted_kernel`, `_nonoverlap_weighted_serial_kernel`, and `_nonoverlap_norm_kernel` are removed; they had no remaining callers after the rewrite.

[...]

PR #147, opened 2026-05-18
test(dsv4): cover non-overlap compressor real shapes

xiaguan:
## Summary

- add a permanent ignored GPU gate for `deepseek_compressor_nonoverlap_prefill_cuda`
- cover the real shapes that mattered for PR #145 acceptance: ratio2 odd, ratio4 representative, ratio128 minimal, 10k ratio2, and 10k ratio128
- use token-specific sparse BF16 inputs in the reference so wrong route consumption or ratio truncation changes weighted/output values

## Scope

- test/gate correction only
- no runtime kernel, Rust runtime, FFI, launcher, HTTP, or profile changes
- shape source is PR #145 5090 acceptance plus DSV4 `compress_ratios` ratio4/ratio128 coverage; this is not a standalone performance claim

## 5090 ignored GPU gate

`cargo test -r -p pegainfer-kernels --features deepseek-v4 --test deepseek_compressor_nonoverlap -- --ignored --nocapture`

| Shape | weighted max_abs / first diff | BF16 out max_abs / first diff |
| --- | --- | --- |
| ratio2 odd (`21,64,32,2`) | `1.1641532e-10`, first diff idx 10 | `0`, none |
| ratio4 representative (`20,64,32,4`) | `1.1641532e-10`, first diff idx 3 | `0`, none |
| ratio128 minimal (`128,64,32,128`) | `1.1641532e-10`, first diff idx 1 | `0`, none |
| 10k ratio2 (`10580,4096,512,2`) | `1.1641532e-10`, first diff idx 1 | `0.00390625`, first diff idx 22544 |
| 10k ratio128 (`10580,4096,512,128`) | `4.656613e-10`, first diff idx 0 | `0`, none |

## Validation

- `cargo fmt --check`
- `git diff --check`
- `pre-commit run --files pegainfer-kernels/tests/deepseek_compressor_nonoverlap.rs`
- 5090 ignored GPU gate above: PASS

### DIFF
--- a/pegainfer-kernels/tests/deepseek_compressor_nonoverlap.rs
+++ b/pegainfer-kernels/tests/deepseek_compressor_nonoverlap.rs
@@ -1,0 +1,306 @@
+#![cfg(feature = "deepseek-v4")]
+
+use std::ffi::c_void;
+use std::mem::size_of;
+use std::ptr;
+
+use anyhow::{Context, Result, ensure};
+use cudarc::driver::sys::CUstream;
+use half::bf16;
+use pegainfer_kernels::ffi;
+
+const CUDA_MEMCPY_HOST_TO_DEVICE: i32 = 1;
+const CUDA_MEMCPY_DEVICE_TO_HOST: i32 = 2;
+
+unsafe extern "C" {
+    fn cudaMalloc(dev_ptr: *mut *mut c_void, size: usize) -> i32;
+    fn cudaFree(dev_ptr: *mut c_void) -> i32;
+    fn cudaMemcpy(dst: *mut c_void, src: *const c_void, size: usize, kind: i32) -> i32;
+    fn cudaDeviceSynchronize() -> i32;
+}
+
+struct DeviceBuffer<T> {
+    ptr: *mut T,
+    len: usize,
+}
+
+impl<T: Copy + Default> DeviceBuffer<T> {
+    fn from_host(data: &[T]) -> Result<Self> {
+        let mut ptr = ptr::null_mut();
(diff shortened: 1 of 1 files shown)


xiaguan:
Request changes.

The new gate has the right target area, but the reference input makes the most important part of the test degenerate.

`make_case` fills `x` with all ones (`pegainfer-kernels/tests/deepseek_compressor_nonoverlap.rs:106`), and `reference_nonoverlap` then reduces the projection to one `wkv_sums[dim]` / `wgate_sums[dim]` value per output dimension (`:142-168`). That means every token in every route has the same projected value for a given dim. The softmax may scan 2 routes, 16 routes, or 128 routes; the weighted value is still the same constant because all route values are equal.

This weakens the exact blocker this PR is supposed to lock down. A future regression that silently caps ratio128 routing or groups the wrong token pair can still pass the numeric check as long as the host wrapper no longer rejects the shape. The odd-tail case also mostly proves “does not return invalid value”, not that the consumed `compressed * ratio + route` token range is correct.

Please make the input/reference non-degenerate: route/token-specific values must affect both the score and value projections, while keeping the 10k reference cheap enough to run. One acceptable shape is to vary only a small number of input dimensions per token and compute the projection analytically, but the output must change if the kernel reads the wrong route or silently uses a ratio subset. Keep the same five shape sources and the same no-runtime-change scope.

Local review checks on `875cdca`:
- `git diff --check github/main...HEAD` passed
- `cargo fmt --check` passed
- `pre-commit run --files pegainfer-kernels/tests/deepseek_compressor_nonoverlap.rs` passed

Issue #174, opened 2026-05-26
Qwen3-4B base greedy drift vs HF bf16 due to RoPE bf16 materialization boundary


NolanHo:
## Summary

Qwen3-4B base greedy generation can diverge from Hugging Face GPU bf16 on tie-sensitive logits because PegaInfer previously missed several bf16 materialization boundaries that HF/PyTorch preserves.

This was discovered while working on Qwen3 LoRA support in #173, where base-model numerical drift had to be separated from LoRA-specific correctness issues.

Tracked repro:

- Model: Qwen3-4B local weights matching the public Qwen3-4B release
- Prompt: `Tell me a story`
- Prompt token ids: `[40451, 752, 264, 3364]`
- Truth source: HF `transformers` incremental decode on CUDA, `torch.bfloat16`

The original visible mismatch happened at generated step 5:

| Step | HF GPU bf16 | PegaInfer before fix |
| ---: | --- | --- |
| 5 | token `6941` (`" named"`) | token `879` (`" who"`) |

HF had a real one-bucket bf16 margin:

| Candidate | HF raw bf16 logit | PegaInfer before fix |
| --- | ---: | ---: |
| `6941` (`" named"`) | `20.75` | tied |
| `879` (`" who"`) | `20.625` | tied |

## Root Cause

This was not a LoRA route issue, scheduler issue, CUDA Graph issue, lm_head issue, or sampler-only issue. The tracked mismatch came from numerical policy differences in the Qwen3 forward path.

The accepted fix set has four parts:

1. Preserve bf16 materialization on residual add before RMSNorm.
2. Match HF/PyTorch greedy tie behavior by selecting the lowest token id on exact logits ties.
3. Split Qwen3 MLP `gate_proj` and `up_proj` GEMMs instead of using one stacked `gate_up_proj` GEMM.
4. Match HF/PyTorch RoPE elementwise bf16 materialization inside fused q/k norm+RoPE.

The final remaining root cause was the RoPE arithmetic boundary. PegaInfer fused the RoPE multiply/add as:

```text
result = bf16(lo * cos - hi * sin)
result = bf16(lo * sin + hi * cos)
```

HF/PyTorch bf16 eager behavior materializes each elementwise multiply first:

```text
lo_cos = bf16(lo * cos)
hi_sin = bf16(hi * sin)
result = bf16(lo_cos - hi_sin)

lo_sin = bf16(lo * sin)
hi_cos = bf16(hi * cos)
[...]

Issue #173, opened 2026-05-25
feat(qwen3): add staged LoRA inference support starting with Qwen3-4B


NolanHo:
## Background

PegaInfer already has an OpenAI-compatible serving path and a Qwen3-4B model crate. LoRA support would make it possible to serve fine-tuned variants without materializing merged full-model checkpoints.

The important constraint is to keep the current base-model path unchanged by default. LoRA behavior should only be enabled when the server is started in LoRA mode, similar to vLLM's serving design.

## Goal

I would like to contribute staged LoRA inference support, starting with Qwen3-4B (`pegainfer-qwen3-4b`).

Initial target:

- Adapter format: HuggingFace / PEFT LoRA safetensors.
- Request semantics: follow vLLM-style adapter selection where possible.
- Target modules: `q_proj`, `k_proj`, `v_proj`, `o_proj`, `gate_proj`, `up_proj`, `down_proj`.
- First serving shape: one active LoRA adapter at a time.

## Scope

In scope:

- Add startup flags for LoRA mode, such as `--enable-lora`, `--max-lora-rank`, `--max-loras`, and `--lora-modules name=path`.
- Add Qwen3-4B LoRA adapter loading and validation.
- Add a Qwen3-4B adapter registry / manager.
- Support dynamic adapter loading through `/v1/load_lora_adapter`.
- In the first PR, make `/v1/load_lora_adapter` wait until the scheduler is idle before mutating active adapter state.
- Apply LoRA deltas for the Qwen3-4B attention and MLP target modules listed above.
- Accept PEFT logical target module names even though Qwen3-4B stores some base projections fused internally:
  - `qkv_proj = [q_proj; k_proj; v_proj]`
  - `gate_up_proj = [gate_proj; up_proj]`
- Keep base-model requests on the existing fast path when LoRA mode is disabled.

Out of scope for the first PR:

- Multiple concurrent LoRA adapters in the same batch.
- Hot-swapping an adapter while requests are actively executing without waiting for idle.
- Optimized fused multi-LoRA kernels.
- CUDA Graph performance optimization for LoRA requests.

## Suggested PR Split

I suggest splitting this into three PRs:

### PR 1: LoRA mode and exclusive single-adapter MVP

Get correct Qwen3-4B LoRA inference with minimal scheduler disruption.

- Add LoRA startup flags and Qwen3-4B adapter loading.
- Support one active adapter at a time.
[...]

PR #175, opened 2026-05-26
fix(qwen3): align bf16 greedy numerics

NolanHo:
## Summary

Fixes Qwen3-4B base greedy drift against HF GPU bf16 by matching the materialization boundaries that affect tie-sensitive logits.

This was discovered while working on Qwen3 LoRA support in #173, where base-model numerical drift had to be separated from LoRA-specific correctness issues. The root-cause issue is tracked in #174.

Changes:

- preserve bf16 residual-add materialization before RMSNorm in Qwen3 prefill/decode/unified paths
- use lower-token-id greedy tie-break for exact logit ties
- split Qwen3 MLP gate/up projections into separate row-sliced GEMMs
- match HF/PyTorch RoPE bf16 elementwise multiply materialization in fused q/k norm+RoPE

## Validation

Local:

```bash
cargo fmt --all --check
PEGAINFER_CUDA_SM=80 cargo check --release -p pegainfer-qwen3-4b
```

GPU worker evidence from the investigation:

- `Tell me a story`, Qwen3-4B, HF CUDA bf16 vs PegaInfer
- no first diff through 128 generated tokens
- no first diff through 256 generated tokens

## Notes

The existing Qwen3 e2e golden text appears to encode the old drift branch, so a follow-up should add or refresh a regression gate for this HF bf16 greedy trace.


xiaguan (approved):
LGTM

PR #180, opened 2026-05-27
feat(qwen3): add single-adapter LoRA inference mode

NolanHo:
## Summary

This PR implements the first Qwen3 LoRA inference slice from #173.

It adds an opt-in `--enable-lora` serving mode for Qwen3-4B. In that mode, PegaInfer can dynamically load one PEFT-style LoRA adapter through a vLLM-compatible `POST /v1/load_lora_adapter` endpoint and apply the adapter during generation.

The existing no-LoRA serving path stays unchanged unless `--enable-lora` is passed.

## What is included

- Adds an engine control path for LoRA-capable engines:
  - `EngineCommand`
  - `EngineHandle::load_lora_adapter(...)`
  - unsupported-engine error handling for non-LoRA paths
- Adds Qwen3 LoRA startup support:
  - `--enable-lora`
  - Qwen3-specific LoRA control startup
  - idle-only scheduler mutation for adapter loading
- Adds PEFT adapter loading for Qwen3-4B:
  - reads `adapter_config.json`
  - reads `adapter_model.safetensors`
  - validates tensor names, dtypes, shapes, rank, alpha, layer count, and target modules
  - supports `BF16`, `F16`, and `F32` adapter tensors, converted to host-side `bf16`
- Supports these Qwen3 LoRA target modules:
  - `q_proj`
  - `k_proj`
  - `v_proj`
  - `o_proj`
  - `gate_proj`
  - `up_proj`
  - `down_proj`
- Applies LoRA deltas with correctness-first extra GEMMs:
  - `B @ (A @ x) * alpha / rank`
  - handles fused QKV layout
  - handles fused `gate_proj` / `up_proj` with row-range scaled adds
- Exposes a live vLLM-compatible load route:
  - `POST /v1/load_lora_adapter`
  - accepts `lora_name` and `lora_path`
  - rejects unsupported advanced fields for PR1
- Adds test and validation tooling:
  - Qwen3 adapter loader tests
  - Qwen3 scheduler/control tests
  - ignored Qwen3-4B GPU smoke test
  - live HF/PEFT parity tool

## Scope and intentional limits

This PR is intentionally a single-adapter MVP:

- Qwen3-4B only.
- One active LoRA adapter at a time.
- Adapter loading is serialized through the scheduler and only mutates state when generation is idle.
- `--tp-size > 1` is rejected when `--enable-lora` is set.
- CUDA Graph is disabled in Qwen3 LoRA mode.
- No per-request mixed LoRA batching yet.
- No `/v1/unload_lora_adapter` yet.
- No grouped/optimized LoRA kernels yet.

Those limits keep this PR focused on API shape, adapter loading, and correctness. Multi-GPU TP LoRA, batch-level adapter selection, CUDA Graph cache keys, unload, and performance kernels can be added in follow-up PRs.

## Frontend note

The pinned Rust `vllm_server::serve` API currently owns router construction internally and does not expose a route-extension hook.

To keep the public API compatible with vLLM without forking that dependency, Qwen3 LoRA mode starts the normal vLLM server on an internal loopback port and exposes a same-port axum wrapper:

- `/v1/load_lora_adapter` is handled by PegaInfer.
- Existing OpenAI/vLLM routes are proxied to the internal vLLM server.
- Non-LoRA mode still calls `vllm_server::serve` directly.

This is meant as a narrow workaround for PR1. If `vllm_server` exposes a route-extension API later, this can be replaced with a direct route merge.

## Validation

Local checks:

- `cargo fmt --check`
- `cargo test -p pegainfer-engine --lib`
- `cargo test -p pegainfer-vllm-frontend --lib`
- `PEGAINFER_CUDA_SM=80 cargo test -p pegainfer-qwen3-4b --lib scheduler`
- `PEGAINFER_CUDA_SM=80 cargo test -p pegainfer-qwen3-4b --lib lora -- --nocapture`
- `PEGAINFER_CUDA_SM=80 cargo check -p pegainfer-qwen3-4b`
- `PEGAINFER_CUDA_SM=80 cargo check -p pegainfer-server`
- `python -m py_compile tools/qwen3_lora_live_parity.py`

GPU/model checks on `mint_dev_worker5` with Qwen3-4B:

- `cargo test --release -p pegainfer-qwen3-4b --test lora_smoke -- --ignored --nocapture`
- Live `/v1/load_lora_adapter` + `/v1/completions` HF/PEFT parity smoke passed.
- After rebasing onto the Qwen3 bf16 greedy numerics fix, the deterministic live parity fixture passed with PEFT adapter autocast disabled through:
  - `--max-tokens 48`
  - `--max-tokens 64`
  - `--max-tokens 128`
  - `--max-tokens 180`

The same deterministic fixture first diverges at generated token 181. I am not using that as a broad long-text parity claim for this PR; the PR claim is that the LoRA path loads, serves, applies the adapter, and matches the HF/PEFT reference for the validated smoke range.

## Notes

- PEFT defaults to `autocast_adapter_dtype=True`, which promotes the synthetic bf16 adapter to fp32. PegaInfer currently loads that fixture as bf16, so the longer parity runs use `--disable-peft-adapter-autocast` to compare against the same adapter dtype semantics.
- worker5 required `LD_LIBRARY_PATH=/usr/local/cuda-12.9/compat:$LD_LIBRARY_PATH` because its system driver library does not expose `cuCtxCreate_v4`.

Part of #173.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3665,6 +3665,7 @@ dependencies = [
  "pegainfer-kernels",
  "pegainfer-vllm-support",
  "rand 0.10.1",
+ "safetensors",
  "serde",
  "serde_json",
  "sha2 0.11.0",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3740,6 +3740,7 @@ name = "pegainfer-vllm-frontend"
 version = "0.1.0"
 dependencies = [
  "anyhow",
+ "axum",
  "log",
  "pegainfer-engine",
  "rmp-serde",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3744,6 +3744,7 @@ dependencies = [
  "axum",
  "log",
  "pegainfer-engine",
+ "reqwest",
  "rmp-serde",
  "rmpv",
  "serde",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -93,6 +93,7 @@ tracing-log = "0.2.0"
 tracing-subscriber = { version = "0.3.20", features = ["env-filter", "json"] }
 rmp-serde = "1.3"
 rmpv = "1.3"
+reqwest = { version = "0.12", default-features = false, features = ["stream"] }
 safetensors = "0.7"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0.149"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ members = [
 
 [workspace.dependencies]
 anyhow = "1.0"
+axum = "0.8"
 bincode = "2.0.1"
 bindgen = "0.72.1"
 bytes = { version = "1.10.1", features = ["serde"] }
--- a/pegainfer-core/src/ops.rs
+++ b/pegainfer-core/src/ops.rs
@@ -18,7 +18,8 @@ pub use pegainfer_kernels::ops::{
     add_batch, add_batch_into, embedding_decode_into, extract_vec, extract_vec_into,
     fused_add_rms_norm_into, gemm, gemv, linear, qk_norm_partial_rope_batched_decode_hd256_into,
     rms_norm, rms_norm_batch_offset_into, rms_norm_gated_batch_into, rms_norm_into,
-    rms_norm_offset_into, silu_mul_batch, silu_mul_batch_into, write_vec_into,
+    rms_norm_offset_into, scaled_add_batch_into, scaled_add_rows_into, silu_mul_batch,
+    silu_mul_batch_into, write_vec_into,
 };
 #[cfg(not(feature = "kernel-call-trace"))]
 pub use pegainfer_kernels::ops::{
--- a/pegainfer-engine/Cargo.toml
+++ b/pegainfer-engine/Cargo.toml
@@ -6,5 +6,8 @@ edition = "2024"
 [dependencies]
 tokio = { workspace = true, features = ["sync"] }
 
+[dev-dependencies]
+tokio = { workspace = true, features = ["macros", "rt"] }
+
 [lints]
 workspace = true
--- a/pegainfer-engine/src/engine.rs
+++ b/pegainfer-engine/src/engine.rs
@@ -1,10 +1,12 @@
 use std::{
+    error::Error,
+    fmt,
     path::PathBuf,
     sync::Arc,
     thread::{self, JoinHandle},
 };
 
-use tokio::sync::mpsc;
+use tokio::sync::{mpsc, oneshot};
 
 use crate::sampler::SamplingParams;
 
@@ -59,6 +61,47 @@ pub struct GenerateRequest {
     pub echo: bool,
 }
 
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct LoadLoraAdapterRequest {
+    pub lora_name: String,
+    pub lora_path: PathBuf,
+}
+
+pub enum EngineControlRequest {
+    LoadLoraAdapter {
+        request: LoadLoraAdapterRequest,
+        response_tx: oneshot::Sender<std::result::Result<(), String>>,
+    },
+}
+
+pub enum EngineCommand {
+    Generate(GenerateRequest),
+    Control(EngineControlRequest),
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub enum EngineControlError {
+    Unsupported(&'static str),
+    ChannelClosed,
+    OperationFailed(String),
+}
+
+impl fmt::Display for EngineControlError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Unsupported(message) => f.write_str(message),
+            Self::ChannelClosed => f.write_str("engine control channel closed"),
+            Self::OperationFailed(message) => {
+                write!(f, "engine control operation failed: {message}")
+            }
+        }
+    }
+}
+
+impl Error for EngineControlError {}
+
+pub type EngineControlResult<T> = std::result::Result<T, EngineControlError>;
+
 pub enum TokenEvent {
     Scheduled {
         queued_at_unix_s: f64,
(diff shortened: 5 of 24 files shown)


xiaguan (changes requested):
## Key Issues

### 1. Reverse proxy workaround — needs upstream fix

The ~140-line axum reverse proxy in `pegainfer-vllm-frontend/src/lib.rs` is understandable given that `vllm_server::serve` doesn't expose a route-extension hook. That said, this workaround carries real cost:

- **TOCTOU race in `reserve_loopback_port()`**: bind → close → re-bind leaves a window for port theft. The 5-retry loop patches the symptom.
- **Full request body buffering**: `to_bytes(body, PROXY_BODY_LIMIT)` reads up to 128 MB into memory before forwarding. Response streaming is fine (`bytes_stream()`), but the request path is not.
- **Hand-rolled hop-by-hop header filtering**: missing `Proxy-Connection`, and a blacklist approach is fragile compared to a whitelist.
- **Two heavy new deps** (`axum`, `reqwest`) for a temporary workaround.

**Action item:** Please file an issue (or PR) upstream to `vllm_server` requesting a route-extension hook (e.g., accepting an `axum::Router` to merge). We'll track cleanup on our side (see #183) — once upstream exposes the hook, the entire proxy layer, `reqwest` dep, and retry logic can be deleted.

### 2. Scheduler loop copy-paste

`scheduler_loop_with_lora_control` (278 new lines) duplicates ~90% of the existing `scheduler_loop` — admission, plan building, execution, effect application are all copied verbatim. The only delta is control-command draining at idle.

Next person who fixes a scheduler bug has to patch two places. This should be a single `scheduler_loop` that optionally accepts a control channel (or uses the unified `EngineCommand` channel for both generation and control).

### 3. EngineHandle dual-channel design

`EngineHandle` now has `submit_tx: Option<...>` + `command_tx: Option<...>` — two `Option`s that are mutually intended to be exclusive but not enforced at the type level. `from_parts` takes three `Option` params = 8 possible states, most invalid.

Use an enum to distinguish the two modes, or unify on `EngineCommand` everywhere.

### 4. 21 repeated LoRA if-let-apply blocks

The same 10-line pattern appears 7× in each of `batch_decode.rs`, `prefill.rs`, `unified_forward.rs`:

```rust
if let Some((lora_layer, scale)) = self.lora_layer(layer_idx)
    && let Some(projection) = &lora_layer.q_proj
{
    apply_lora_projection_delta(&self.ctx, projection, input, out, 0, scale)?;
}
```

A helper like `self.maybe_apply_lora(layer_idx, |l| l.q_proj.as_ref(), input, out, 0)` would collapse all 21 sites to one-liners.

### 5. Temporary GPU allocations on every decode step

`apply_lora_projection_delta` allocates two fresh GPU buffers (`rank_out`, `delta`) per call. With 7 projections × 36 layers, that's up to 504 GPU mallocs per decode step. These shapes are constant during decode — pre-allocate them into the decode buffer struct.

### 6. Arbitrary file read via `/v1/load_lora_adapter`

The endpoint accepts any `lora_path` and does `fs::read` with no path validation. Any client that can reach the HTTP port can trigger reads of arbitrary files on the host. vLLM mitigates this with `--allowed-lora-modules` / path allowlists. At minimum, consider a `--lora-base-dir` flag that restricts adapter loading to a specific directory subtree.

### 7. BF16 precision divergence at token 181

PR description notes parity with HF/PEFT breaks at token 181. The workaround is disabling PEFT's `autocast_adapter_dtype` — i.e., degrading the reference to match. The root cause is that `apply_lora_projection_delta` runs entirely in BF16 while PEFT defaults to FP32 intermediates. Worth tracking as a known limitation with a follow-up issue.

---

## Minor / style (nice-to-have cleanups)

These don't block merging but would improve the PR if you're willing to tidy up:

- **`_task_type` unused variable** in `inspect_lora_adapter` — either use it or delete it.
- **`inspect_lora_adapter` deserializes safetensors twice** — once for validation, returns raw bytes, then `load_lora_adapter` deserializes again. Merge or return parsed `SafeTensors` directly.
- **`DeviceLoraLayer` uses named fields vs `LoraLayer`'s `BTreeMap`** — two representations for the same data. Consider aligning.
- **`reqwest` without TLS** — `default-features = false` at workspace level means any future crate using `reqwest` also loses TLS.
- **`effective_cuda_graph` fix mixed into feature PR** — the change from `args.cuda_graph` to `effective_cuda_graph` for non-LoRA Qwen3 is an independent bug fix; ideally a separate commit.
- **`start_with_executor_with_lora_control`** — naming could be more concise.

---

## Summary

The core LoRA functionality — adapter config parsing, tensor validation, PEFT-compatible loading, delta application — is solid and well-tested at the unit level. The main concerns are:

1. **Proxy hack** → file upstream vllm issue for route extension; we'll track cleanup (#183)
2. **Scheduler duplication** → unify into one loop
3. **EngineHandle type safety** → enum over dual-Option
4. **21× code repetition in forward path** → extract helper
5. **Per-step GPU allocation** → pre-allocate
6. **Path traversal risk** → add `--lora-base-dir` restriction

xiaguan (approved):
Approving — the core LoRA functionality (adapter loading, validation, delta application) is solid. The reverse proxy workaround is acceptable given the upstream `vllm_server` limitation; tracked in #183 for cleanup once upstream exposes a route-extension hook.

A few items to consider for follow-ups or quick fixes before merge:

### Worth addressing now

- **Scheduler loop duplication**: `scheduler_loop_with_lora_control` copies ~90% of `scheduler_loop`. Unifying into one loop with an optional control channel would prevent future divergence.
- **EngineHandle dual-channel `Option`s**: `submit_tx: Option` + `command_tx: Option` — 8 possible states, most invalid. An enum or unified `EngineCommand` channel would be type-safer.
- **21 repeated LoRA if-let-apply blocks** across `batch_decode.rs`, `prefill.rs`, `unified_forward.rs`: a helper like `self.maybe_apply_lora(layer_idx, |l| l.q_proj.as_ref(), input, out, 0)` would collapse each to a one-liner.
- **Per-decode-step GPU allocations**: `apply_lora_projection_delta` allocates two temp buffers per call (up to 504 mallocs/step). Pre-allocate into decode buffers.
- **Arbitrary `lora_path`**: no path validation on the HTTP endpoint. Consider a `--lora-base-dir` flag to restrict loading to a directory subtree.

### Follow-up issues

- **BF16 precision divergence at token 181**: root cause is BF16 intermediates vs PEFT's FP32 default. Worth a tracking issue.
- **Upstream `vllm_server` route extension**: please file an issue/PR upstream so we can remove the proxy layer (#183).

### Minor style (nice-to-have)

- `_task_type` unused variable — delete or use it
- `inspect_lora_adapter` deserializes safetensors twice — mergeable
- `effective_cuda_graph` bug fix mixed into this PR — ideally a separate commit
- `reqwest` without TLS at workspace level
- `start_with_executor_with_lora_control` naming

PR #197, opened 2026-05-30
test(qwen35): add HF logits golden gate

### LINKED ISSUE #186: Add Qwen3.5 HF logits gate and PegaInfer rand/hash regression corpus
## Problem

`pegainfer-qwen35-4b` currently has an exact-text e2e regression test backed by `test_data/Qwen3.5-4B.json`, but this fixture mixes two different meanings:

- HF parity: proving PegaInfer matches a pinned external Hugging Face oracle.
- PegaInfer regression: detecting output drift after a known-good PegaInfer baseline is accepted.

The current JSON is useful as a PegaInfer regression guard, but it is not traceable enough to serve as an HF parity gate.

## Direction

Split the Qwen3.5 accuracy checks into two gates:

1. A small HF-backed logits parity gate.
2. A deterministic PegaInfer-owned rand/hash regression corpus.

The HF gate should establish that PegaInfer matches the external oracle for a small pinned shape. Once that gate is green, PegaInfer can become the accepted baseline for broader deterministic regression checks.

The rand/hash corpus can use fixed seeds to generate many prompt token sequences and store output token hashes, instead of relying on a small set of natural-language Q&A prompts.

## Current Evidence

At commit `627ff5f8a540d982ba9f9535f4800d029a0fa943`, using local HF refs/main `851bf6e806efd8d0a36b00ddf55e13ccb7b8cd0a`:

- HF current snapshot vs checked-in `test_data/Qwen3.5-4B.json`: `6 / 10` mismatches.
- Pega current regen vs checked-in golden: `5 / 10` mismatches.
- Pega current regen vs HF current snapshot: `4 / 10` mismatches.

This shows that the existing fixture creates ambiguous signals. Some failures are stale-baseline regressions, while others may be real HF parity gaps.

## Acceptance Criteria

- Add a small Qwen3.5 HF logits parity gate.
- Record the model revision, tokenizer/config revision, dtype/backend assumptions, prompt token ids, forward path, and comparison tolerance.
- Treat `test_data/Qwen3.5-4B.json` as PegaInfer regression-only, or move HF parity data into a separate fixture path.
- Add or prepare a deterministic rand/hash regression corpus generated from fixed random seeds.
- Store enough metadata for the rand/hash corpus to identify the producer commit/config and accepted PegaInfer baseline.
- Compare token ids or token hashes first; decoded text can remain diagnostic.
- If the local model revision does not match fixture metadata, fail or skip with a clear message.

## Out Of Scope

- Running HF for every broad regression case.
- Treating natural-language Q&A text as the main long-term regression oracle.
- Fixing every Qwen3.5 correctness mismatch found by the current fixture in this issue.

## Claim Boundary

This issue is about separating oracle semantics and making Qwen3.5 accuracy regression reproducible.

The HF logits gate proves a small external-oracle parity surface. The rand/hash corpus is a broad PegaInfer regression gate after that parity surface is accepted.

CAICAIIs:
## Summary

Fixes #186 for the current maintainer-requested gate shape.

This replaces the retired Qwen3.5 exact-text self-baseline with a small HF-backed logits golden gate. The new gate dumps a pinned `AutoModelForCausalLM.eval()` / `use_cache=True` / `past_key_values` oracle, then teacher-forces the same token sequences through PegaInfer's Qwen3.5 runtime.

## Scope

In scope:

- Add a model-local Qwen3.5 logits executor for teacher-forced prefill/decode replay.
- Add `tools/accuracy/dump_qwen35_4b_hf_golden.py`.
- Add `test_data/qwen35-4b-hf-golden.safetensors` with model/config/backend/tolerance metadata.
- Add `pegainfer-qwen35-4b/tests/hf_golden_gate.rs`.
- Cover the Qwen3.5 replay surfaces: sequential CUDA-graph decode, bucket-straddling batched graph decode, and slot-compaction replay after dropping a mid-batch request.
- Retire the old exact-text `e2e.rs`, `regen_test_data.rs`, and `test_data/Qwen3.5-4B.json` self-baseline path.
- Keep `e2e_scheduler.rs` as scheduler liveness/request-flow coverage only.

Out of scope:

- A cross-architecture exact-token rand/hash corpus. The earlier hash fixture path is deferred because exact generated tokens may drift across GPU architectures.
- Running HF for every broad regression prompt.
- Claiming all Qwen3.5 long-prompt or GSM8K accuracy gaps are fixed.

## Evidence

Local checks on `5056587daa1601addf9150d1f5b9542c7bb62b6b`:

- `cargo fmt --check`
- `git diff --check xiaguan/main...HEAD`
- `python3 -m py_compile tools/accuracy/dump_qwen35_4b_hf_golden.py`

RTX 5090 `sm_120` checks on `7bdb65f6d1c2034d0d742417484a4791d2f290b2`, before rebasing onto the latest upstream Kimi-K2/frontend-only changes:

- `cargo +nightly test --release -p pegainfer-qwen35-4b --test hf_golden_gate -- --nocapture`
  - sequential graph: mean `0.0248`, p99 `0.0862`, max `0.1193`
  - batched graph 5 padded: mean `0.0256`, p99 `0.0757`, max `0.1068`
  - batched graph 3 padded: mean `0.0260`, p99 `0.1007`, max `0.1206`
  - slot-compaction graph: mean `0.0285`, p99 `0.1031`, max `0.1168`
- `cargo +nightly test --release -p pegainfer-qwen35-4b --test e2e_scheduler -- --nocapture`
- `cargo +nightly test --release -p pegainfer-qwen35-4b --tests --no-run`

## Claim Boundary

This PR proves a small external-oracle Qwen3.5 logits parity surface against a pinned HF bf16 reference and locks the Qwen3.5-owned replay paths that exist today.

It does not claim broad prompt-level accuracy closure. Long-prompt cases should be added after the current long-prompt root cause work has a stable target shape.


xiaguan (changes requested):
test_data/qwen35-4b-rand-hash.json
I'm not sure whether this is stable across different archs (e.g., sm 120, sm 90). They may give different results. But the approach is good. I just think we could remove the hash.json and only keep the safetensors from HF?

CAICAIIs:
> test_data/qwen35-4b-rand-hash.json I'm not sure whether this is stable across different archs (e.g., sm 120, sm 90). They may give different results. But the approach is good. I just think we could remove the hash.json and only keep the safetensors from HF?

good point

CAICAIIs:
I removed the rand/hash fixture path from this PR and narrowed the scope to the HF safetensors-backed logits gate only. The broader cross-arch exact-token/hash corpus is now deferred until we have a policy for sm_90/sm_120 token drift.

Current commit keeps:
- the HF `qwen35-4b-hf-golden.safetensors` fixture,
- the teacher-forced Qwen3.5 logits executor,
- sequential graph, bucket-straddling batched graph, and slot-compaction replay,
- retired exact-text e2e/regen self-baseline.

I also updated the PR title/body to match this narrower scope.

xiaguan (approved):
LGTM, thanks @CAICAIIs! Ran the gate 6 times on a local RTX 5070 Ti (sm_120, CUDA 13.3) against the pinned Qwen3.5-4B: parity metrics pass consistently and match your reported numbers (sequential: mean 0.0243 / p99 0.0760 / max 0.1097; batched and slot-compaction surfaces in the same band). The executor replays the real production prefill/decode/CUDA-graph path, the fixture is pinned and crash-early on env mismatch, and the earlier review request (drop the rand-hash corpus, keep only the HF safetensors) is fully addressed.

One observation, not a blocker: on the very first run (cold, freshly built binary) the sequential determinism assert tripped — runs 1 and 2 within that execution diverged on ~9 consecutive positions (~1e-2 logprob deltas), then 5 subsequent executions were bit-identical every time. The assert is doing its job: that looks like rare run-to-run nondeterminism in the engine on sm_120, worth tracking as its own engine issue rather than a gate problem.

Issue #160, opened 2026-05-22
Frontend hot path: kernel/syscall time dominates under sustained SSE load


xiaguan:
Follow-up from #138 (frontend CPU profile) and #159 (jemalloc switch).

This issue records the **objective profile data** from running `pegainfer-sim` under `vllm bench serve` and points at the biggest remaining CPU sinks in the Rust frontend path, so we can target them with concrete work.

## Setup

- Branch: `perf/jemalloc`
- Sim binary: `pegainfer-sim --tpot-ms 0 --base-ttft-ms 0 --prefill-tokens-per-ms 10000`
- Sim pinned: `taskset -c 0,1` (all worker threads inherit)
- Load: `vllm bench serve --backend openai --endpoint /v1/completions --dataset-name random --random-input-len 256 --random-output-len 128 --num-prompts 30000 --max-concurrency 128`
- Profiler: `perf record -F 999 -e cpu-clock -g --call-graph dwarf,16384 -m 64M -C 0,1 -- sleep 20`
- Bench client on CPUs 2-15 (separate from sim)

## Bench result (jemalloc build)

```
Successful requests:                     30000
Benchmark duration (s):                  57.92
Request throughput (req/s):              517.98
Output token throughput (tok/s):         66,301
Peak output token throughput (tok/s):    76,623
Mean TTFT (ms):                          85.59
Mean TPOT (ms):                          1.18
P99 TPOT (ms):                           1.51
```

Sim used **1.50 cores** during the 20s perf window — meaning each output token costs ≈ **22 µs of frontend CPU**. End-to-end throughput is **bench-client-bound** (the Python `vllm bench` client saturates its cores first), so the sim has spare capacity at this load.

## Where the frontend CPU goes (39,976 samples, 74.09% in `tokio-rt-worker`)

Aggregated by category, expressed as **% of busy CPU** (sample share / tokio-rt-worker share):

| Category | % of busy CPU | Notes |
|---|---|---|
| **kernel / syscalls** | **~63%** | epoll, futex, sched, network write/recv, timer |
| **vllm frontend stream poll** | 4.4% | `decoded_text_event_stream` → `completion_chunk_stream` → `completion_sse_stream` three-layer wrap, each polled per token |
| **tokio runtime / scheduler** | 3.7% | `run_task`, worker schedule |
| **tokenizer (fastokens)** | 3.4% | BPE encode on prompt + `byte_level_decode` per output token, `RefCell<FlatCache>::with` overhead visible |
[...]

PR #199, opened 2026-05-31
perf(vllm-frontend): batch consecutive token bridge outputs

peter941221:
1. Summary

This batches consecutive TokenEvent::Token items in the vLLM frontend bridge into fewer EngineCoreOutputs payloads.

The old path paid one local channel send, one msgpack encode, and one ZMQ send per token. Under the simulated frontend benchmark from #160, that left fixed per-message cost on the table.

The new path keeps the first streamed token on the direct path, then coalesces later decode bursts by draining ready token events with try_recv() into one multi-token EngineCoreOutput.

2. Invariants

The patch does not change request scheduling, decode semantics, or the external HTTP API shape.

The first batch still carries queued/scheduled events and prefill stats exactly once.

Batched logprobs now keep strict positional alignment with token_ids, including mixed batches where some tokens do not carry logprobs.

Terminal outputs and finished_requests behavior stay the same.

3. Validation

rustfmt +nightly-2026-01-01 --edition 2024 --check pegainfer-vllm-frontend/src/lib.rs
cargo +nightly-2026-01-01 test -p pegainfer-vllm-frontend --lib -- --nocapture

Focused tests added in this PR:

1. consecutive_tokens_are_batched_into_one_output
2. first_token_metadata_is_only_sent_with_first_batch
3. mixed_logprob_batch_keeps_token_alignment

4. Benchmark

Commands and machine notes:
- temporary WSL verification environment
- server: pegainfer-sim --tpot-ms 0 --base-ttft-ms 0 --prefill-tokens-per-ms 10000
- server pinned to CPUs 0-1, client pinned to CPUs 2-15
- shape: 3000 requests, max concurrency 128, random prompt words 256, random output len 128

benchmark                         shape                      before         after          delta
request throughput                3000 req / conc 128       679.32 req/s   761.21 req/s   +12.1%
output token throughput           out len 128               86952.55 tok/s 97434.27 tok/s +12.1%
mean TTFT                         same                       9.56 ms        9.60 ms        +0.4%
mean TPOT                         same                       1.3681 ms      1.2407 ms      -9.3%
p99 TPOT                          same                       5.6240 ms      6.2400 ms      +11.0%
mean streamed token-event count   same                       128.0          124.646        -2.6%
server CPU seconds                same                       3.96           4.19           +5.8%

The strongest signal here is lower streamed event count with higher request and token throughput at essentially unchanged TTFT. The CPU-seconds number moved the wrong way in this rerun, so I do not claim a clean CPU reduction from this version of the patch.

5. Related

Related to #160.

### DIFF
--- a/pegainfer-vllm-frontend/src/lib.rs
+++ b/pegainfer-vllm-frontend/src/lib.rs
@@ -14,6 +14,7 @@ use axum::response::{IntoResponse, Response};
 use axum::routing::post;
 use log::{info, warn};
 use serde::{Deserialize, Serialize};
+use tokio::sync::mpsc::error::TryRecvError;
 use tokio::sync::{RwLock, mpsc};
 use tokio::task::JoinHandle;
 use tokio_util::sync::CancellationToken;
@@ -843,7 +844,15 @@ async fn run_request_stream(
 ) {
     let mut first_token_events = None;
     let mut first_token_prefill_stats = None;
-    while let Some(event) = token_rx.recv().await {
+    let mut pending_event = None;
+    loop {
+        let event = match pending_event.take() {
+            Some(event) => event,
+            None => match token_rx.recv().await {
+                Some(event) => event,
+                None => return,
+            },
+        };
         match event {
             TokenEvent::Scheduled {
                 queued_at_unix_s,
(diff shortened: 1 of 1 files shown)


peter941221:
Addressed the two review points in de16880.

1. Mixed batches now keep logprobs aligned with token_ids by inserting an empty PositionLogprobs entry for tokens that do not carry logprobs. I added mixed_logprob_batch_keeps_token_alignment to cover that case.

2. I did not remove coalescing outright. A full removal pushed mean streamed token-event count back to about 128.0 in the same harness, which meant batching mostly disappeared. The current version keeps the first streamed token on the direct path, then yields only after the request has already emitted one token output.

Rerun on the same async harness:
- mean streamed token-event count: 128.0 -> 124.646
- request throughput: 679.32 req/s -> 761.21 req/s
- output token throughput: 86952.55 tok/s -> 97434.27 tok/s
- mean TTFT: 9.56 ms -> 9.60 ms

xiaguan (approved):
LGTM

PR #218, opened 2026-06-04
refactor(vllm-frontend): use router extension for LoRA routes

### LINKED ISSUE #183: chore(vllm-frontend): remove LoRA reverse proxy once upstream exposes route-extension hook
## Context

PR #180 adds a LoRA serving mode for Qwen3. Because `vllm_server::serve` doesn't expose a route-extension hook, the PR works around this by running the vLLM server on an internal loopback port and proxying through an axum wrapper that intercepts `/v1/load_lora_adapter`.

This workaround adds ~140 lines of proxy code, `reqwest` + `axum` deps to `pegainfer-vllm-frontend`, a TOCTOU-prone port reservation with 5-retry loop, full request body buffering (128 MB limit), and hand-rolled hop-by-hop header filtering.

## Action items

1. File an issue or PR upstream to `vllm_server` requesting a route-extension API (e.g., accepting additional `axum::Router` routes to merge into the server).
2. Once upstream provides the hook, remove:
   - `serve_lora_proxy`, `proxy_to_upstream_inner`, `reserve_loopback_port`, `wait_for_http_health`
   - `should_forward_request_header` / `should_forward_response_header`
   - `ProxyState`, `PROXY_BODY_LIMIT`
   - `reqwest` dependency from `pegainfer-vllm-frontend`
3. Replace with direct route registration on the existing server.

## Related

- PR #180

NolanHo:
## Summary

Closes #183.

This PR replaces the LoRA frontend's same-port reverse-proxy workaround with vLLM Rust's native router-extension hook.

Before this change, LoRA mode had to start an internal vLLM HTTP server on a reserved loopback port, then put a PegaInfer proxy on the public port to mount `/v1/load_lora_adapter` and rewrite adapter-as-model OpenAI requests. That worked, but it was heavier than necessary and had to preserve HTTP behavior manually.

Now that vLLM exposes `serve_with_router_extension`, PegaInfer can build on the normal Rust vLLM frontend directly:

- the public server is the vLLM server;
- PegaInfer mounts LoRA control routes on the same router;
- adapter-as-model OpenAI requests are forwarded in-process through the vLLM router;
- no secondary loopback HTTP server or `reqwest` proxy is needed.

Non-LoRA serving remains on the existing normal vLLM path.

## What Changed

- Updated the vLLM Rust dependencies to a revision that includes `serve_with_router_extension`.
- Removed the LoRA same-port proxy implementation:
  - no reserved loopback upstream port;
  - no internal vLLM HTTP server;
  - no `reqwest` forwarding path;
  - no manual proxy header forwarding.
- Added a shared helper that serves PegaInfer through `vllm_server::serve_with_router_extension`.
- In LoRA mode, merged three route surfaces into one public server:
  - existing vLLM OpenAI routes;
  - PegaInfer LoRA control routes such as `/v1/load_lora_adapter`;
  - adapter-aware OpenAI route overlays for `/v1/completions`, `/v1/chat/completions`, and `/v1/models`.
- Preserved adapter-as-model behavior:
  - if `model` is a loaded adapter name, rewrite it to the base served model before vLLM validation;
  - pass the selected adapter through `vllm_xargs.pegainfer_lora_adapter`;
  - dispatch the rewritten request to the vLLM router in-process.
- Kept `/v1/models` overlay behavior so loaded adapter names are listed alongside the base served model.
- Added a simulated frontend e2e test that starts the real vLLM frontend with LoRA routes enabled and verifies `/v1/load_lora_adapter` is mounted on the OpenAI server port.

## Scope

This PR is intended to be a frontend integration cleanup only.

It does not change:

- Qwen3 LoRA weight loading;
- LoRA numerical kernels;
- scheduler adapter selection;
- request-level LoRA semantics;
- tensor-parallel LoRA behavior;
- the public LoRA API shape.

The goal is to remove the workaround that was needed before vLLM had a Rust router-extension API, while preserving the behavior added by the earlier LoRA PRs.

## Validation

Local checks:

```bash
cargo test -p pegainfer-vllm-frontend --lib
cargo check -p pegainfer-vllm-frontend
cargo test -p pegainfer-sim --test frontend_e2e -- --nocapture
git diff --check
```

Remote simulated route e2e:

```bash
cargo test -p pegainfer-sim --test frontend_e2e \
  simulated_lora_routes_are_mounted_on_openai_frontend -- --nocapture
```

Result: passed.

Remote Qwen3-4B live parity smoke:

```bash
python tools/lora/qwen3_lora_live_parity.py \
  --model-path "$QWEN3_4B_MODEL_PATH" \
  --startup-timeout-s 600 \
  --max-tokens 8 \
  --disable-peft-adapter-autocast
```

Result:

```json
{
  "load_response": "Success: LoRA adapter 'parity' added successfully.",
  "hf_token_ids": [911, 264, 3908, 3743, 6941, 444, 10524, 879],
  "pegainfer_token_ids": [911, 264, 3908, 3743, 6941, 444, 10524, 879],
  "first_token_mismatch": null,
  "match": true
}
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,9 @@ jobs:
           toolchain: nightly
           components: rustfmt, clippy
 
+      - name: Install build dependencies
+        run: sudo apt-get update && sudo apt-get install -y protobuf-compiler
+
       - name: Check formatting
         run: cargo fmt --all --check
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4232,6 +4232,7 @@ version = "2.19.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "805bfd7352166bae857ee569628b52bcd85a1cecf7810861ebceb1686b72b75d"
 dependencies = [
+ "indexmap 2.14.0",
  "memo-map",
  "serde",
  "serde_json",
@@ -5462,13 +5463,13 @@ dependencies = [
  "axum",
  "log",
  "pegainfer-engine",
- "reqwest",
  "rmp-serde",
  "rmpv",
  "serde",
  "serde_json",
  "tokio",
  "tokio-util",
+ "tower",
  "uuid",
  "vllm-engine-core-client",
  "vllm-server",
@@ -9106,7 +9107,7 @@ checksum = "051eb1abcf10076295e815102942cc58f9d5e3b4560e46e53c21e8ff6f3af7b1"
 [[package]]
 name = "vllm-chat"
 version = "0.1.0"
-source = "git+https://github.com/vllm-project/vllm.git?rev=65b7a812a2dabd212d78c7b5b8a320b4efb9750d#65b7a812a2dabd212d78c7b5b8a320b4efb9750d"
+source = "git+https://github.com/vllm-project/vllm.git?rev=4aaed4ca225a3745aa1e18864dad0599d3ac7626#4aaed4ca225a3745aa1e18864dad0599d3ac7626"
 dependencies = [
  "anyhow",
  "asynk-strim-attr",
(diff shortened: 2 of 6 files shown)


xiaguan (approved):
This is how it should be done — you upstreamed the router-extension hook into vllm itself (vllm-project/vllm#43774) and then came back to delete the workaround. Fixing the limitation at the source instead of maintaining a proxy forever is exactly the right call, and −246 lines of reserved-port/health-poll/reqwest machinery is a very satisfying diff. Also appreciated: the live parity run in the validation this time. 🎉

I checked the middleware semantics against the pinned vllm rev: the extension receives the finalized router (with `track_server_load`/`track_http_metrics`/`TraceLayer`), the overlay itself carries no layers, and rewritten requests hit the inner router exactly once via `oneshot` — so metrics and tracing still cover LoRA traffic, and streaming responses pass through unbuffered.

LGTM.

Issue #263, opened 2026-06-05
Engine panics with "undefined symbol: cuCtxCreate_v4" on CUDA 12.2 drivers (cudarc pinned to cuda-12090)


wjinxu:
**Symptom:** any GPU run panics at startup:
`libcuda.so.1: undefined symbol: cuCtxCreate_v4`
This comes from the workspace `cudarc` pin (`cuda-12090`), which eagerly resolves `cuCtxCreate_v4` — a symbol only present in driver >= 560 (CUDA 12.6+).
**Question, not a bug report:** is a minimum driver/CUDA version intended? If so, it'd help to state it in the Prerequisites. If not, the qwen35 path at least builds and runs (and `--test e2e_scheduler` passes) after lowering the feature to `cuda-12020`, suggesting the v4 requirement may be just the feature level rather than actual API usage. I haven't checked the deepseek-v4 / kimi-k2 paths, which may genuinely need newer cudarc APIs.

xiaguan:
Good catch, and your diagnosis is exactly right — there is no intended driver minimum; the v4 requirement was purely the feature level.

What happens mechanically: cudarc 0.19.4's dynamic-loading mode resolves the *entire* symbol table for the pinned binding level (`cuda-12090`) the first time `libcuda.so` is opened, panicking on the first missing symbol. `cuCtxCreate_v4` enters the table at `cuda-12050`, so the pin silently imposed a ≥ CUDA 12.5 driver floor that nothing in the codebase actually needs.

Rather than lowering the feature, the fix is upstream: cudarc 0.19.5 ([PR #573](https://github.com/chelsea0x3b/cudarc/pull/573)) switched to per-symbol lazy resolution — each function resolves its own symbol on first call. After that, `cuda-12090` is just the compile-time binding level and your driver only needs the symbols actually called. I audited the workspace: every cudarc symbol any model path calls is available on all supported drivers (primary context, graph capture v2, plain mem/stream ops).

To your caveat about deepseek-v4 / kimi-k2: they don't need newer *cudarc* APIs either. Their genuinely new driver calls (`cuMulticast*`, fabric handles) go through `pegainfer-comm`'s own `-sys` bindings, independent of the cudarc pin. The one real extra requirement on those paths is NCCL ≥ 2.27 at runtime (`ncclAlltoAll`), which now also resolves lazily — the default Qwen build no longer touches it.

[...]

PR #264, opened 2026-06-05
fix(deps): bump cudarc to 0.19.7 to drop eager driver symbol loading

xiaguan:
## Problem (#263)

Any GPU run on a CUDA 12.2 (R535) driver panics at startup:

```
libcuda.so.1: undefined symbol: cuCtxCreate_v4
```

Root cause: cudarc 0.19.4's dynamic-loading mode resolves **every** symbol at the pinned feature level (`cuda-12090`) when `libcuda.so` is first opened, with `.expect()` per symbol. `cuCtxCreate_v4` enters the binding table at `cuda-12050`, so the pin imposed a driver ≥ CUDA 12.5 floor — even though no code path ever calls it.

## Fix

cudarc 0.19.5 (upstream [PR #573](https://github.com/chelsea0x3b/cudarc/pull/573)) replaced eager loading with per-symbol lazy resolution (`OnceLock` per function, resolved on first call). After that, the `cuda-XXXX` feature is only the compile-time binding level; the runtime driver only needs the symbols actually called.

Bump the manifest to `0.19.7` (latest). The floor matters: the README driver claim depends on ≥ 0.19.5 lazy-loading behavior, so it belongs in the requirement itself, not in `Cargo.lock` state — an old lock would silently reintroduce the startup panic.

## Why this is safe for every model path

Audited the full workspace against cudarc's version gates:

- Every cudarc symbol pegainfer calls is ungated (available at all binding levels): `CudaContext::new` → `cuDevicePrimaryCtxRetain`, graph capture (`cuStreamBeginCapture_v2`, `cuGraphInstantiateWithFlags`), basic mem/stream ops. The only version-dispatched constructor (`new_non_primary`, v3/v4) is never called.
- The genuinely new driver APIs (`cuMulticast*` 12.1+, `CU_MEM_HANDLE_TYPE_FABRIC` 12.3+) live in `pegainfer-comm`'s own `-sys` bindings, not cudarc — unaffected by this change and only built for EP features.
- NCCL bindings get the same lazy property: `ncclAlltoAll` (NCCL ≥ 2.27) resolves only when EP all-to-all actually runs; the default Qwen build no longer requires it at load time.
- The one breaking change in 0.19.5 (`CUmemAllocationHandleType` enum → u32) touches only cudarc's VMM types, which pegainfer doesn't use.

README prerequisites now state the driver floor (R535+) and the EP-path NCCL ≥ 2.27 requirement.

## Validation

On RTX 5070 Ti, driver 580.126.09 (CUDA 13.0), toolkit 13.3 — covers the new-driver side:

| Check | Result |
|---|---|
| `cargo test --release --workspace --lib` | 63 crates pass (one pre-existing `kvbm-config` figment flake under full-workspace parallelism; passes 3/3 standalone, identical on 0.19.4) |
| qwen3 `hf_golden_gate` (GPU) | pass, 37.8s |
| qwen35 `e2e_scheduler` (GPU) | pass, 17.4s |
| qwen35 `hf_golden_gate` | skipped by fixture (local weights lack revision metadata) — numerics covered by the qwen3 gate over shared kernels |

The old-driver side (12.2) follows analytically from lazy resolution + the symbol audit; asked the reporter to confirm on their box in #263.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1776,9 +1776,9 @@ dependencies = [
 
 [[package]]
 name = "cudarc"
-version = "0.19.4"
+version = "0.19.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f071cd6a7b5d51607df76aa2d426aaabc7a74bc6bdb885b8afa63a880572ad9b"
+checksum = "1cea5f10a99e025c1b44ae2354c2d8326b25ddbd0baf76bde8e55cfd4018a2cc"
 dependencies = [
  "half",
  "libloading 0.9.0",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,7 +89,9 @@ colored = "3"
 comfy-table = "7.2"
 criterion = { version = "0.8", features = ["html_reports"] }
 crossbeam-channel = "0.5"
-cudarc = { version = "0.19.4", features = [
+# 0.19.5+ required: per-symbol lazy loading keeps the cuda-12090 binding
+# level from imposing a runtime driver floor (issue #263).
+cudarc = { version = "0.19.7", features = [
     "cuda-12090",
     "cublas",
     "f16",
--- a/README.md
+++ b/README.md
@@ -46,8 +46,10 @@ Measured on **RTX 5070 Ti** (16 GB), BF16, CUDA Graph enabled, single request:
 ### Prerequisites
 
 - Rust (2024 edition), CUDA Toolkit (nvcc, cuBLAS), CUDA-capable GPU
+- NVIDIA driver R535 (CUDA 12.2) or newer; driver symbols resolve lazily at call time, so the `cuda-12090` cudarc feature does not raise the driver floor
 - Python 3 + Triton (build-time only — no Python at runtime)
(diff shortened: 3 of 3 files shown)


wjinxu:
Confirmed on real CUDA 12.2 hardware.

- GPU: RTX 4090, driver 535.104.05 (CUDA 12.2), built with a 12.2 toolkit (sm_89).
- Before: with cudarc 0.19.4 + `cuda-12090`, any GPU run panics at startup with `libcuda.so.1: undefined symbol: cuCtxCreate_v4` (eager symbol load).
- After: `cargo update -p cudarc` -> 0.19.7, Cargo.toml left at `cuda-12090`. No startup panic; `cargo test --release -p pegainfer-qwen35-4b --test e2e_scheduler` passes (16s). Lazy per-symbol resolution means cuCtxCreate_v4 is never resolved since nothing calls it.

So the analytical claim holds on the 12.2 side: lock bump alone fixes it, no Cargo.toml/feature change needed. (Scope of my check: single-GPU qwen35 path; I didn't exercise the EP/NCCL all-to-all path.)

PR #266, opened 2026-06-05
fix(qwen35): size RoPE cache from config for long prompts

### LINKED ISSUE #250: qwen35: long-prompt accuracy collapse — GSM8K 8-shot scores ~2% vs ~79% on the HF reference
### Background

Qwen3.5-4B passes its short-prompt exact-greedy e2e and serves at vLLM-parity speed, but a standard lm-eval run against the OpenAI endpoint shows GSM8K 8-shot collapsing to ~2% where the HF reference scores ~79%. Short prompts match HF exactly, so the divergence onsets with prompt length — pointing at the prefill path rather than decode.

### Current state

- Candidates include RoPE position handling past some boundary (the cos/sin cache covers only 4096 positions — see the RoPE sub-issue), the linear-attention chunk recurrence at long sequence length, and the prefill staging path. None is isolated yet; the first step is establishing the onset length.
- The dump/inspect tooling earlier accuracy work used has been deleted, so bisection tooling needs rebuilding — the planned teacher-forcing executor is the right vehicle.

### Desired outcome

- The divergence is root-caused and fixed; GSM8K 8-shot lands within a few points of the HF reference.
- A long-prompt case is added to whatever accuracy gate exists at that point, so this regression class is permanently covered.

### Acceptance boundary

- Reproducible on a single GPU with lm-eval (or an equivalent harness) against `/v1/completions`.
- The fix is verified at the logits level, not just task score.

CAICAIIs:
## Summary
Fix Qwen3.5 long-prompt accuracy by sizing the RoPE cache from `max_position_embeddings` instead of the old 4096-position constant. The patch also adds fail-closed prefill/decode cache coverage checks, a long HF logits golden over 4097/8192-token prompts, and docs recording the recovered GSM8K 8-shot score.

Fixes #250

## Scope
In scope:
- Qwen3.5 RoPE cache length and coverage checks
- long-prompt HF logits golden fixture and dumper options
- GSM8K 8-shot recovery evidence for `/v1/completions`

Out of scope:
- MMLU/HellaSwag/ARC
- non-greedy sampling
- long-context admission semantics
- `batch_size > 1` task-score evals

## Evidence
- `git diff --check`
- `cargo +nightly fmt --check`
- `python3 -m py_compile tools/accuracy/dump_qwen35_4b_hf_golden.py`
- RTX 5090 `sm_120`: `cargo +nightly test --release -p pegainfer-qwen35-4b --test hf_golden_gate -- --nocapture`
  - 2 tests passed
  - long replay: 18 positions, mean 0.0216, p99 0.0700
- `/v1/completions` HTTP smoke:
  - 5001 prompt tokens
  - HTTP 200
  - server reported `max_model_len=262144`
- Full GSM8K 8-shot `lm-eval`, 1319 examples, no `--limit`:
  - strict-match: 79.38%
  - flexible-extract: 79.30%
  - HF baseline: 79.45%

## Claim Boundary
This recovers the measured Qwen3.5 GSM8K 8-shot serving path and adds a logits-level gate for the old RoPE cache boundary. It does not claim broader benchmark recovery beyond the listed eval.


xiaguan (approved):
Thanks for the fix

Issue #220, opened 2026-06-04
qwen3: RoPE cache covers 4096 positions but 40960-token contexts are admitted (silent corruption)


xiaguan:
## Problem

The Qwen3-4B RoPE cos/sin cache is precomputed for a hard-coded **4096 positions** (`pegainfer-qwen3-4b/src/weights.rs:314`), but nothing limits request context to that length:

- RoPE kernels index `cos_cache[pos * head_dim + d]` with no bounds check (`pegainfer-kernels/csrc/prefill_attention.cu:75`, `prefill_attention_hd256.cu:90`, decode qk_norm_rope path).
- Scheduler admission limits requests by **KV blocks only** (`max_request_blocks`), which on any reasonable GPU allows far more than 4096 tokens.
- The frontend advertises `max_model_len = max_position_embeddings = 40960` from `config.json`.

Any request whose context (prompt + generated) crosses position 4096 reads past the cache → garbage rotations for those positions → silently corrupted output. No error anywhere.

## Repro

Serve Qwen3-4B, send a >4096-token prompt (or a ~4090-token prompt with `max_tokens=64`, which crosses the boundary mid-decode). Output degrades with no error. Note the 4k/64 split-K bench sits just *under* the boundary, which is why this never showed up.

## Acceptance

Precompute the cache up to `config.max_position_embeddings` (40960 × 128 × bf16 ≈ 10.5 MB per cache — negligible), and crash-early reject at admission if `prompt + max_tokens` exceeds the cache length as a backstop. YaRN beyond native length stays #8.

xiaguan:
Roadmap note (now tracked under #241): the checkpoint ships `rope_scaling: null`, so sizing the plain cos/sin precompute from config is a *correct* fix today — but #8 (YaRN) is the prerequisite for any rope-scaled checkpoint, where the precompute length must come from the scaled schedule. The two should stay split. Qwen3.5 has the identical bug (cache 4096 vs `max_position_embeddings: 262144`), filed as #253 — worth fixing with one shared pattern.

scatyf3:
Hi, I''m interested in this issue, can you assign this to me

xiaguan:
Thanks.

xiaguan:
Closing as fixed. Current Qwen3 builds the RoPE cache from , rejects over-context requests at scheduler admission, and has both oversized-reject and 4097-token in-window regression tests (, ). YaRN for rope-scaled checkpoints remains tracked separately in #8.

[...]

PR #273, opened 2026-06-05
fix(qwen3): position-encoding context window 

scatyf3:
### summary

fix issue #220 's rope encoding error

code
1. frontend: 
	1. let frontend parse `max_position_embeddings` from model's `config.json` and set default to 40960
	2. add warning to pegainfer server, if config.json parse error, fallback to 4096 and raise an warning
2.  `precompute_rope` with `max_position_embeddings` instead of fix value
3. `pegainfer-kernels/csrc/shared/prefill_attention.cu`: add index check to `qk_norm_rope_batched_decode_cuda`, avoid index out of bound error
4. `pegainfer-qwen3-4b/src/scheduler.rs`
	1. add `max_context_tokens` rejecetion mechism in scheduler. `admit_deferred_requests` 
	2. add `RejectReason` enumrate
	3. change other tests accordingly

test:
1. test the `admit_deferred_requests` function in `requests_exceeding_context_window_are_rejected`
2. test executor end to end scheduling with `FakeExecutor` in `over_context_window_request_is_rejected_through_scheduler_loop`, where it should reject  long request with explicit reason and keep serving after request is rejected
3. GPU e2e test:  `oversized_prompt_is_rejected_with_context_length_error`

## Type of Change

Bug fix (non-breaking change which fixes an issue)

## Checklist
-  My code follows the style guidelines of this project (see `docs/areas/coding-style.md`).
-  I have performed a self-review of my own code.
-  I have formatted my commits according to Commitizen conventions.
-  I have run the local test suite and all tests pass (see `CLAUDE.md`).


CAICAIIs:
One acceptance/regression gate is missing before merge: add a >4096 but <=40960 Qwen3 request test, e.g. 4097 prompt tokens with max_tokens=1, to prove the resized RoPE cache is actually exercised

xiaguan (changes requested):
Thanks for the fix. The code path looks coherent to me: config.max_position_embeddings now sizes the RoPE cache, scheduler admission rejects over-window requests before prefill, and the kernel-side cos_max_pos trap is a useful last-resort guard.

The remaining blocker is documentation. This PR closes the Qwen3 #220 RoPE OOB item, but docs/models/qwen3/roadmap.md still describes #220 as a live correctness bug, and docs/index.md still lists it in the Qwen3 open set. Please update those docs in the same PR so the roadmap does not become stale the moment this lands.

One non-blocking note: pegainfer-qwen3-4b/src/scheduler.rs is now about 1.5k lines. It was already too large before this PR, so I would not ask this bugfix PR to solve it, but we should probably open a follow-up issue to split out the admission policy / rejection reasons / inline scheduler tests into a smaller module. The current direction adds more logic to an already oversized scheduler file.

xiaguan (approved):
Toxic-reviewed with code pulled and run locally (RTX 5070 Ti sm_120, Qwen3-4B weights). **LGTM.**

## Verified by reading

- **Fix chain is complete and consistent.** `precompute_rope` sized from `config.max_position_embeddings` (weights.rs:313), admission rejects `prompt + max_tokens > max_context_tokens` *before* any prefill (scheduler.rs:504), kernel `__trap`s an out-of-range position as last-resort backstop (prefill_attention.cu:71). Since `qk_norm_rope_batched_decode_cuda` launches the same kernel, the trap covers the decode path too.
- **No off-by-one.** Max position ever used = `prompt + max_tokens − 1 ≤ max_context_tokens − 1`, and the cos table holds indices `0..max_context_tokens`. A request using the window exactly is admitted.
- **`cos_max_pos = cos_cache.data.len() / head_dim` is correct** against the `precompute_rope` layout (`total = max_seq_len * head_dim`, half-split duplication per position — weight_loader.rs:222).
- **C ABI matches**: `.cu` launcher param order (`rms_eps, cos_max_pos, stream`) = ffi declaration (ffi/shared.rs:202). All 4 Rust call sites threaded; compiler enforces the rest.
- **Blast radius is qwen3-only.** qwen35 does not call this kernel (own Triton path); its hardcoded 4096 (qwen35/weights.rs:297) is the known sibling, failure mode unchanged by this PR.
- Docs honestly keep the `rope_scaling: null` caveat and leave YaRN #8 open.

## Verified by running (all green)

- `cargo test --release -p pegainfer-qwen3-4b --lib` → 25/25, incl. both new scheduler unit tests
- `--test context_window` → oversized 60k-token prompt rejected with context-length error in 3.4s (never reaches prefill), engine keeps serving after
- `--test context_window_in_window` → 4097-token prompt (position 4096, first index past the old table) served end-to-end
- `--test hf_golden_gate` → 7/7 scenarios, mean 0.0317 / p99 0.1196 — RoPE table resize causes zero accuracy drift, as expected (per-position values are size-independent)

## Minor (non-blocking)

Frontend fallback when `max_position_embeddings` is absent is 4096 (lib.rs:445) while the engine serde default is 40960 (config.rs:42) — asymmetric, but the conservative direction with a loud warn; real Qwen3 configs always carry the field.

Nice test hygiene: one engine-starting test per binary to avoid GPU contention, with the reasoning documented in the file header.

PR #283, opened 2026-06-06
test(kvbm): cover prefix cache eviction rematch

### LINKED ISSUE #247: qwen3: prefix-cache eviction path has no behavioral test
### Background

The prefix cache (#216) evicts blocks under pool pressure, and correctness after eviction depends on the radix layer truncating matches to what is actually resident. The accuracy gate replays cache-hit passes, but never the evict-then-rematch sequence.

### Current state

- No test registers a prefix, releases it, drives the pool to eviction, and matches again. A partial-eviction bug (matching blocks that were reclaimed) would surface as silent output corruption in production, not as a test failure.
- The block-matching layer is CPU-side logic; this needs no GPU.

### Desired outcome

- A behavioral test of the evict-then-rematch cycle: full hit before eviction, truncated/zero match after, correct recompute on the next request.

### Acceptance boundary

- Runs in the standard `cargo test` suite without GPU or model weights.

jackyYang6:
## Description

Fixes #247

Adds a CPU-only contract test for the prefix-cache evict-then-rematch cycle in `kvbm-logical`. The test verifies that a released resident prefix can be matched, evicted prefix blocks are not returned as stale matches, and recomputed prefixes can be matched again by a later request.

## Type of Change

- [ ] Documentation update
- [x] Test update

## Checklist

- [x] My code follows the style guidelines of this project.
- [x] I have run the local test suite and all tests pass.

Test run:

```bash
cargo test --release -p kvbm-logical --lib test_prefix_cache_eviction_rematch_zero_then_recompute

### DIFF
--- a/kvbm/kvbm-logical/src/integrations/scheduled.rs
+++ b/kvbm/kvbm-logical/src/integrations/scheduled.rs
@@ -1993,6 +1993,75 @@ mod tests {
         assert_eq!(seq.prefill_position(), 0);
     }
 
+    #[test]
+    fn test_prefix_cache_eviction_rematch_zero_then_recompute() {
+        let manager = create_test_manager::<TestMeta>(4);
+        let tokens = make_tokens(12);
+
+        let mut cold = SchedulableSequence::<TestMeta>::new(
+            tokens.clone(),
+            0,
+            BLOCK_SIZE,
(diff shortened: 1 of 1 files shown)


CAICAIIs (approved):
lgtm

Issue #252, opened 2026-06-04
qwen35: prefill stages the whole sequence in a transient buffer (large allocation + hard length cap)


xiaguan:
### Background

Qwen3.5-4B prefill materializes the full sequence's K/V in a contiguous HND staging buffer before copying into the paged pool, with the staging size fixed by a hard `MAX_SEQ` constant.

### Current state

- Each long-prompt prefill transiently allocates on the order of hundreds of MB, then pays an extra device-to-device pass to scatter into pages.
- Prompts beyond the constant cap cannot be served regardless of available KV memory; the cap is far below the model's configured max positions.
- The full-attention layers already read paged KV at decode — the staging is a prefill-only detour.

### Desired outcome

- Prefill writes K/V directly into paged blocks; the transient buffer and the hard cap disappear.
- Admissible prompt length becomes a function of KV pool capacity, consistent with the RoPE and admission fixes on this roadmap.

### Acceptance boundary

- Greedy parity on existing e2e baselines (or the logits gate, once landed) — this is a memory-layout change, not a numeric one.
- Peak prefill memory measured before/after for a long prompt.

wjinxu:
I'd like to take this one if that's OK.

PR #290, opened 2026-06-06
fix(qwen35): reserve full KV lifetime during admission

### LINKED ISSUE #254: qwen35: admission ignores generation budget; KV exhaustion mid-decode kills the whole batch
### Background

Qwen3-4B had this exact failure class (#85) and fixed it with full-lifetime KV admission, waiting-queue deferral, and per-request rejection. Qwen3.5-4B still has the pre-fix semantics.

### Current state

- Admission sizes KV by prompt length only — `max_tokens` is not reserved, so concurrent decodes can over-commit the pool.
- The engine contract defines a rejection event, but this scheduler never emits it.
- When KV runs out mid-decode, the failure aborts the batch — innocent in-flight requests die with the offending one.

### Desired outcome

- Admission accounts for the full request lifetime (prompt + max generation), deferring requests that don't currently fit.
- Impossible requests are rejected with a clear error; KV exhaustion, if it can still occur, fails only the affected request.

### Acceptance boundary

- A pressure test in the style of the #85 verification: sustained over-capacity load completes all admissible requests and cleanly rejects the rest — no hang, no batch-wide abort.

CAICAIIs:
## Summary

Fixes #254.

Qwen3.5 admission now reserves each admitted request's full KV lifetime budget (`prompt_len + max_tokens - 1`) instead of prompt-only pages. Temporarily over-budget requests stay deferred in FCFS order, requests that can never fit this model instance are rejected with `TokenEvent::Rejected`, and Qwen3.5 execution/sampling failures surface as `TokenEvent::Error`.

## Scope

In scope:
- Qwen3.5 scheduler admission budgeting.
- Active-request future KV page reservation.
- Impossible-request rejection.
- Explicit request error semantics for scheduler execution failures.
- Direct rejection-event coverage and Qwen3.5 issue #254 validation notes.

Out of scope:
- Qwen3.5 prefill scratch / HND staging OOMs. That remains covered by #252.
- Full per-offender isolation for arbitrary batch-level GPU execution failures.

## Evidence

Local:
- `cargo fmt --check`
- `git diff --check`
- Residual scan found no `SchedulerDriver` / `FakeSchedulerDriver` / `max_context_tokens=` / `debug_assert!(current_pages...)` leftovers.

Remote H20 review-fix:
- `cargo test --offline --release -p pegainfer-qwen35-4b send_rejection_reports_kv_lifetime_request_tokens --lib -- --nocapture` — 1 passed.
- `cargo test --offline --release -p pegainfer-qwen35-4b --lib -- --nocapture` — 20 passed.
- Real-model e2e was not rerun on this H20 host because Qwen3.5 weights were absent and HF download failed with `Network is unreachable`.

Prior RTX 5090 / model validation:
- `cargo test --offline --release -p pegainfer-qwen35-4b --test e2e_scheduler -- --nocapture` — 1 passed.
- `cargo test --offline --release -p pegainfer-vllm-frontend rejected_request_is_reported_as_error --lib -- --nocapture` — 1 passed.
- `cargo build --offline --release -p pegainfer-server` — passed, with existing unused-import warnings.
- 100 concurrent admissible real HTTP requests completed `100/100`, with post-pressure `/v1/models` and `/v1/completions` healthy.
- Direct impossible `bench_serving request --prompt-len 1 --output-len 600000 --warmup 0 --iters 1` reached scheduler rejection; the current rejection event label is `max_request_tokens`.

## Claim Boundary

This fixes the KV lifetime admission failure described in #254: admissible over-capacity pressure defers and completes, impossible scheduler requests reject cleanly, and scheduler failures are reported as errors.

It does not claim to solve Qwen3.5 long-prompt prefill scratch pressure or remove the prefill staging buffer; that is #252.


xiaguan (changes requested):
Pulled the branch and ran it locally — the fix itself is solid: lib 22/22, e2e green, and `bench_serving --prompt-len 1 --output-len 600000` rejects instantly with the right message (on a stale pre-PR binary the same request gets admitted and decodes forever, which is a nice live demo of the #254 bug).

Requesting changes on the test seam, not the fix:

1. **Drop `SchedulerDriver` + `FakeSchedulerDriver` + the two loop tests.** The fake reimplements the production event semantics (`fail_fake_plan` is a copy of the error-broadcast logic), so the loop tests exercise the copy, not the code this PR changes — the Finished→Error conversion has zero real coverage, and the copy will drift. If we need a fake to test the loop, the coupling is the problem: push rejection/error-event decisions down into `plan.rs` as pure functions (state + result in → events out), keep `scheduler_loop` concrete and trivial, and let e2e + the bench gate cover the shell. The PR shrinks ~300 lines and coverage goes up.
2. `debug_assert!` in `active_future_pages` is compiled out in release, but the doc claims a "release assertion". Make it `assert!` — if that invariant breaks we under-reserve, which is exactly the bug this PR fixes.
3. Rejection message: `max_context_tokens=` is misleading (it's the request's lifetime demand, not the model window). qwen3 just renamed the same label to `max_request_tokens=` in #273 — align.

CAICAIIs:
> Pulled the branch and ran it locally — the fix itself is solid: lib 22/22, e2e green, and `bench_serving --prompt-len 1 --output-len 600000` rejects instantly with the right message (on a stale pre-PR binary the same request gets admitted and decodes forever, which is a nice live demo of the #254 bug).
> 
> Requesting changes on the test seam, not the fix:
> 
> 1. **Drop `SchedulerDriver` + `FakeSchedulerDriver` + the two loop tests.** The fake reimplements the production event semantics (`fail_fake_plan` is a copy of the error-broadcast logic), so the loop tests exercise the copy, not the code this PR changes — the Finished→Error conversion has zero real coverage, and the copy will drift. If we need a fake to test the loop, the coupling is the problem: push rejection/error-event decisions down into `plan.rs` as pure functions (state + result in → events out), keep `scheduler_loop` concrete and trivial, and let e2e + the bench gate cover the shell. The PR shrinks ~300 lines and coverage goes up.
> 2. `debug_assert!` in `active_future_pages` is compiled out in release, but the doc claims a "release assertion". Make it `assert!` — if that invariant breaks we under-reserve, which is exactly the bug this PR fixes.
> 3. Rejection message: `max_context_tokens=` is misleading (it's the request's lifetime demand, not the model window). qwen3 just renamed the same label to `max_request_tokens=` in #273 — align.

done, thx

Issue #294, opened 2026-06-06
vllm-frontend: engine Rejected events surface as generic HTTP 500 without the message


xiaguan:
## Current state

The kimi engine emits `TokenEvent::Rejected` with an actionable message ("per-request capacity 8192 ...", "DP prompts are capped at 2048 ...", "echo is not supported ...", TP8 logprobs/sampling honor-or-reject). The vllm-frontend bridge maps it to `EngineCoreFinishReason::Error` with no detail (pegainfer-vllm-frontend/src/lib.rs:826-832), and the API layer renders that as:

```
HTTP 500 {"error": {"message": "Internal server error", "type": "server_error"}}
```

Verified on 8×H200 /v1/completions (Kimi TP1/DP8): over-cap `max_tokens` and a 2049-token DP prompt both return the opaque 500; the client never sees why.

## Desired

- `Rejected` → HTTP 400 (invalid_request_error) carrying the engine's message.
- 500 stays reserved for `TokenEvent::Error` (engine faults).

Applies to all model lines (the bridge is shared). Same gap for `TokenEvent::Error`'s message (lib.rs:818-823) — it is also dropped.

## Acceptance

- /v1/completions with `max_tokens` over the lifetime cap returns 400 + the capacity message; server keeps serving (already true today).
- A unit test on the bridge asserting the rejection message reaches the EngineCoreOutputs payload.

FeathBow:
This looks like a follow-up to #209. I'd like to try it.

xiaguan:
Status after digging through the wire protocol (rev 5d5591d):

**The engine-core protocol has no request-level error-message channel.** `EngineCoreOutput.stop_reason` is only preserved for `Stop` finishes (`finish_reason_from_engine` in `llm/src/output.rs:117`); for `Error` it is dropped and `completion_finish_reason_to_openai` renders a bare 500. Our bridge already attaches the message (`StopReason::Text`, covered by the `rejected_request_is_reported_as_error` unit test) — it dies upstream, not here. So the full fix is upstream, not a fork.

**What already holds on qwen3 (verified live on Qwen3-4B):**
- `max_tokens` over the servable cap → 400 + message on all three endpoints (#354/#357 unified guard; chat `max_completion_tokens` covered).
- Engine admission rejects are unreachable via HTTP for qwen3: the handshake reports `servable_len` (40832), so upstream's own tokenized-length validation fires first.
[...]

PR #295, opened 2026-06-06
fix(frontend): log rejected and errored requests instead of swallowin…

### LINKED ISSUE #209: [Bug] Rejected requests return a bare 500 and leave no trace in the logs
### What happens

```bash
curl -s localhost:8000/v1/completions -H 'Content-Type: application/json' \
  -d '{"model":"<model>","prompt":"hi","max_tokens":99999999}'
# → HTTP 500 {"error":{"message":"Internal server error", ...}}
```

The only log output (even at `RUST_LOG=debug`) is a generic line from the external client crate:

```
ERROR vllm_engine_core_client::client::stream: request failed with an internal error during generation
```

The scheduler actually builds a precise reason — `send_rejection` (`pegainfer-qwen3-4b/src/scheduler.rs:512`) produces `"request requires more KV blocks than this model instance can provide: prompt_tokens=…, max_context_tokens=…"` — but it is only forwarded as `TokenEvent::Rejected` and never logged. The bridge (`run_request_stream` in `pegainfer-vllm-frontend/src/lib.rs`) silently converts it to a terminal output, and vllm-server flattens it into a bare 500. Neither the operator nor the client ever sees the root cause.

Same for the two silent `let Some(...) else` guards in `start_request`.

### Suggested fix

Add `warn!` with request_id + message in the `TokenEvent::Error` and `TokenEvent::Rejected` arms of `run_request_stream`, and in the two silent guards in `start_request`. The bridge is the single chokepoint shared by all model crates, so one change covers every model.

(Returning 400 instead of 500 would require changes in the external vllm-server crate — out of scope.)

### Acceptance

The curl above produces a `warn!` line containing the request id and the "requires more KV blocks" message.

FeathBow:
…g them

## Description

This adds the missing `warn!`s in `run_request_stream` (Rejected/Error arms) and the two silent guards in `start_request`.

Fixes #209 

## Type of Change

- [ ] Documentation update

## Checklist

This is a log change, `cargo test` passes on this crate. `--workspace --lib` needs RDMA deps absent on my work dev.

### DIFF
--- a/pegainfer-vllm-frontend/src/lib.rs
+++ b/pegainfer-vllm-frontend/src/lib.rs
@@ -398,10 +398,12 @@ impl LocalEngineBridge {
             ..
         } = request;
         let Some(prompt_tokens) = prompt_token_ids else {
+            warn!("request {request_id} dropped: missing prompt_token_ids");
             send_terminal_output(output_tx, request_id, EngineCoreFinishReason::Error, None)?;
             return Ok(());
         };
         let Some(sampling_params) = sampling_params else {
+            warn!("request {request_id} dropped: missing sampling_params");
             send_terminal_output(output_tx, request_id, EngineCoreFinishReason::Error, None)?;
             return Ok(());
         };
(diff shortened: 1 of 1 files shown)


xiaguan (approved):
LGTM. Verified locally: all four silent paths from #209 now log (two start_request guards + Error/Rejected arms), `warn` was already imported, crate compiles clean. Surfacing the message to the HTTP client stays tracked in #294.

Issue #246, opened 2026-06-04
qwen3: prefix-cache hits are invisible — cached_tokens never reaches the API


xiaguan:
### Background

Prefix caching is on by default for Qwen3-4B (#216) and delivers large warm-TTFT wins, but neither a user nor a benchmark can tell whether a request hit the cache.

### Current state

- The executor computes the per-request cached-token count, but it is dropped at the scheduler boundary — it never reaches the engine event stream.
- The OpenAI-compatible frontend hard-codes `num_cached_tokens: 0` in usage (`pegainfer-vllm-frontend/src/lib.rs`).
- There is no hit-rate logging, so cache effectiveness in production is unobservable.

### Desired outcome

- `usage.prompt_tokens_details.cached_tokens` reports the real per-request value.
- Engine-level hit/miss counters logged or exposed for operators.

### Acceptance boundary

- An integration test asserts a warm repeat of the same prompt reports nonzero cached tokens while the cold run reports zero.
- Related surface: #78 (streaming usage accounting) — same neighborhood, independently fixable.

jackyYang6:
Hi, I'm interested in this Rust inference project and the KV cache; Could you assign this issue to me?

jackyYang6:
@xiaguan Hello, I looked through the current event flow and wanted to confirm the intended shape for the Qwen3 fix.

Other engines already emit `TokenEvent::Scheduled` when a request enters their scheduling / prefill-candidate path, and the vLLM frontend uses that event to build `PrefillStats`.

Qwen3 currently does not emit `TokenEvent::Scheduled`. The real prefix-cache hit count is computed later inside
`Qwen3Executor::execute_prefill` / `execute_unified`, when `RequestKv::match_and_add_prefix` runs.

So if we want to surface `cached_tokens` through `TokenEvent::Scheduled`, the minimal Qwen3 fix would be:
1. record the real `scheduled_at_unix_s` when the request is selected for prefill/unified execution;
2. let the executor compute `cached_tokens` during prefix matching as it does;
3. emit `TokenEvent::Scheduled { ..., cached_tokens }` after prefix matching / executor return, but before the first `Token` or
`Finished` event.

This keeps the frontend contract intact, but means Qwen3 sends the `Scheduled` event later than other engines, while preserving the original scheduled timestamp. Does this approach look good for this issue or would you prefer a different impl path?

xiaguan:
Thanks, this direction sounds reasonable to me.

[...]

PR #299, opened 2026-06-07
Qwen3 prefix-cache usage in OpenAI responses to enhance observability

jackyYang6:
## Description

Fixes issue #246 by propagating Qwen3 prefix-cache hit information from the model scheduler to OpenAI-compatible API usage fields, so users and benchmark tooling can observe whether a request reused cached prompt tokens.

### What changed

- Extended `TokenEvent::Scheduled` with `cached_tokens`.
- Preserved Qwen3 executor-computed `PrefillRequestResult.cached_tokens` through scheduler planning/effects.
- Emitted Qwen3 `Scheduled` events before prompt echo, first token, or terminal finish events so frontend usage accounting has the cache metadata in time.
- Updated non-Qwen3 scheduled senders to report `cached_tokens: 0`, preserving existing behavior for engines without prefix-cache accounting.
- Updated the vLLM/OpenAI frontend to map engine-provided cache metadata into usage:
  - `usage.prompt_tokens_details.cached_tokens`
  - vLLM `PrefillStats.num_cached_tokens`
  - vLLM `PrefillStats.num_local_cached_tokens`
  - `num_computed_tokens = prompt_tokens.saturating_sub(cached_tokens)`
- Added usage patching for OpenAI response bodies where the Rust vLLM server does not currently copy cached-token `PrefillStats` into the final OpenAI `usage` object.
- Added Qwen3 engine-level prefix-cache debug stats for operators.

### Why

Prefix caching is enabled by default for Qwen3-4B and can significantly reduce warm-request TTFT, but before this change the cache hit count was dropped at the scheduler/frontend boundary. API users, benchmarks, and operators could not tell whether a repeated prompt actually hit the prefix cache because usage always reported zero cached tokens and there was no engine-level cache hit/miss visibility.

This change makes per-request cache reuse visible in OpenAI-compatible usage while keeping cache accounting owned by the Qwen3 executor/scheduler rather than guessing in the frontend.

### Usage patch design

The Rust vLLM frontend receives `PrefillStats` from `TokenEvent::Scheduled`, but OpenAI response usage did not reliably preserve the cached-token fields in the final response object. To avoid recomputing cache hits in the frontend, this PR adds a small response wrapper in `pegainfer-vllm-frontend/src/patch_usage.rs`.

The wrapper:

- stores the engine-reported cached-token count by external request id when `TokenEvent::Scheduled` is processed;
- wraps both `/v1/completions` and `/v1/chat/completions`;
- patches only the returned `usage.prompt_tokens_details.cached_tokens` field;
- handles both non-streaming responses and streaming final usage chunks when `stream_options.include_usage=true`;
- removes the cached-token entry after patching to avoid stale request-id state;
- defaults to `0` if an engine did not report cache metadata.

This keeps the frontend as a pass-through/patch layer: it does not infer prefix-cache behavior from prompt length, token ids, or repeated requests. The Qwen3 executor remains the source of truth for `cached_tokens`.

### Engine-level cache info design

Issue #246 also asks for engine-level hit/miss counters logged or exposed for operators. This PR adds a minimal Qwen3-local debug log rather than a new metrics subsystem.

Qwen3 scheduler effects now maintain cumulative prefix-cache stats for the lifetime of the scheduler loop:

- `total_requests`
- `hit_requests`
- `miss_requests`
- `hit_rate`
- `total_prompt_tokens`
- `total_cached_tokens`
- `token_hit_rate`

The stats are logged at `debug!` after each scheduled observation:

```text
pegainfer_qwen3_4b::scheduler::effects: effects.rs:98 Qwen3 prefix cache stats: total_requests=8, hit_requests=6, miss_requests=2, hit_rate=0.7500, total_prompt_tokens=52184, total_cached_tokens=26080, token_hit_rate=0.4998
```

The log is debug-level by design: operators can enable it when investigating prefix-cache effectiveness, while default info logs remain quiet under normal serving load. No prompt text, token ids, generated text, or request payloads are logged.

Example startup filter:

```bash
RUST_LOG=debug  cargo run --release -- --model-path Qwen/Qwen3-4B --port 8000
```

## Type of Change

- [x] Bug fix
- [x] Tests
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Performance change

## Validation

GPU-machine validation performed:

- `cargo test --release -p pegainfer-qwen3-4b --lib scheduler`
  - `17 passed; 0 failed`
- Real `/v1/completions` cold/warm validation:
  - cold: `cached_tokens=0`
  - warm repeated prompt: `cached_tokens=6752`
- Real `/v1/chat/completions` non-streaming cold/warm validation after chat route patch:
  - cold: `cached_tokens=0`
  - warm repeated prompt: `cached_tokens=6512`
- Real streaming validation with `stream_options: {"include_usage": true}`:
  - `/v1/completions` usage chunk includes `prompt_tokens_details.cached_tokens`
  - `/v1/chat/completions` usage chunk includes `prompt_tokens_details.cached_tokens`

Local/sub-agent validation performed:

- `cargo test --release -p pegainfer-vllm-frontend --lib`
- `cargo test --release -p pegainfer-sim --test frontend_e2e cached_tokens`
- `cargo clippy --release -p pegainfer-vllm-frontend -p pegainfer-sim --all-targets -- -D warnings`
- `cargo fmt --check`
- `git diff --check`

Suggested final checks before merge, if not already run on the final clean branch:

```bash
cargo test --release -p pegainfer-qwen3-4b --lib scheduler
cargo test --release -p pegainfer-vllm-frontend --lib
cargo test --release -p pegainfer-sim --test frontend_e2e cached_tokens
cargo clippy --release -p pegainfer-vllm-frontend -p pegainfer-sim --all-targets -- -D warnings
```

## Checklist

- [x] API usage reports real Qwen3 cached-token counts instead of hard-coded zero.
- [x] `TokenEvent::Scheduled` carries `cached_tokens`.
- [x] Qwen3 scheduler preserves executor-computed cached-token counts through effects.
- [x] Qwen3 emits `Scheduled` before prompt echo, token, or finish events.
- [x] Non-Qwen3 engines preserve current behavior with `cached_tokens: 0`.
- [x] `/v1/completions` non-streaming and streaming usage paths are covered.
- [x] `/v1/chat/completions` non-streaming and streaming usage paths are covered.
- [x] Engine-level Qwen3 prefix-cache hit/miss counters are available through debug logs.
- [x] No frontend-side cache inference or recomputation was added.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5464,7 +5464,9 @@ name = "pegainfer-vllm-frontend"
 version = "0.1.0"
 dependencies = [
  "anyhow",
+ "async-stream",
  "axum",
+ "futures",
  "log",
  "pegainfer-engine",
  "rmp-serde",
--- a/pegainfer-deepseek-v2-lite/src/engine.rs
+++ b/pegainfer-deepseek-v2-lite/src/engine.rs
@@ -35,6 +35,7 @@ fn handle_request(generator: &mut DeepSeekV2LiteEp2Generator, req: &GenerateRequ
         queued_at_unix_s: req.queued_at_unix_s.unwrap_or(now),
         scheduled_at_unix_s: now,
         prompt_tokens,
+        cached_tokens: 0,
     });
     if req.echo {
         let _ = req.token_tx.send(TokenEvent::PromptTokens {
--- a/pegainfer-deepseek-v4/src/direct/scheduler.rs
+++ b/pegainfer-deepseek-v4/src/direct/scheduler.rs
@@ -1179,6 +1179,7 @@ fn handle_request(generator: &mut DeepSeekV4DirectGenerator, req: GenerateReques
         queued_at_unix_s,
         scheduled_at_unix_s,
         prompt_tokens: prompt_len,
+        cached_tokens: 0,
     });
     if req.echo {
         let _ = req.token_tx.send(TokenEvent::PromptTokens {
@@ -1356,6 +1357,7 @@ fn handle_request_wave(generator: &mut DeepSeekV4DirectGenerator, requests: Vec<
             queued_at_unix_s,
             scheduled_at_unix_s,
             prompt_tokens: prompt_len,
+            cached_tokens: 0,
         });
         if req.echo {
             let _ = req.token_tx.send(TokenEvent::PromptTokens {
--- a/pegainfer-engine/src/engine.rs
+++ b/pegainfer-engine/src/engine.rs
@@ -134,6 +134,7 @@ pub enum TokenEvent {
         queued_at_unix_s: f64,
         scheduled_at_unix_s: f64,
         prompt_tokens: usize,
+        cached_tokens: usize,
     },
     Token {
         id: u32,
--- a/pegainfer-kimi-k2/src/runner/scheduler/lifecycle.rs
+++ b/pegainfer-kimi-k2/src/runner/scheduler/lifecycle.rs
@@ -95,6 +95,7 @@ pub(in crate::runner) fn send_scheduled(req: &GenerateRequest) {
         queued_at_unix_s: req.queued_at_unix_s.unwrap_or(scheduled_at),
         scheduled_at_unix_s: scheduled_at,
         prompt_tokens: req.prompt_tokens.len(),
+        cached_tokens: 0,
     });
 }
 
--- a/pegainfer-qwen3-4b/src/scheduler.rs
+++ b/pegainfer-qwen3-4b/src/scheduler.rs
@@ -11,6 +11,7 @@ mod resolve;
 
 use std::collections::{HashSet, VecDeque};
 use std::thread;
+use std::time::{SystemTime, UNIX_EPOCH};
 
 use anyhow::Result;
 use log::{info, warn};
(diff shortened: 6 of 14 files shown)


xiaguan:
@codex review

xiaguan:
Heads up — the root cause is upstream: the vllm rust frontend never fills `usage.prompt_tokens_details.cached_tokens` (`Usage::from_counts` hardcodes it to `None`). Filed as vllm-project/vllm#44824 — the cleanest path is to land that upstream and bump our pinned rev, then drop the response-patch layer here.

jackyYang6:
Totally agreed. I initially resorted to this response-patch layer precisely because of this upstream limitation, and as a result, the patching code has grown increasingly bloated and ugly.

Once the upstream fix lands and we bump our pinned rev, I'll gladly drop the patch layer, rewrite it with clean code, and resubmit.

xiaguan:
Closing in favor of #352, which lands the same outcome end-to-end now that the upstream half exists (vllm#44887 fixed the `Usage::from_counts` gap this PR surfaced — tracked as vllm#44824).

@jackyYang6 thank you for pushing this — your PR is what identified the gap and got the upstream issue filed. You are credited as co-author on the #352 commit, so the contribution lands with your name on it. Verified end-to-end: a warm repeat of a 177-token prompt now reports `prompt_tokens_details.cached_tokens: 176` in both non-streaming and streaming (`include_usage`) responses.

jackyYang6:
Thanks for keeping the co-author attribution, I really appreciate it! Looking forward to making more contributions to the project in the future.

PR #302, opened 2026-06-07
fix(frontend): reject max_tokens over servable cap with 400

FeathBow:
## Description

Follow-up to #209, not a full fix for #294. #209 logged the rejection but the client still got a bare 500.

Before this change the handshake advertised `max_model_len = max_position_embeddings`, but the engine only serves `max_request_blocks × block_size` (smaller). The frontend admitted oversized requests without checking before tokenization; the scheduler rejected them later and the bridge turned that into an opaque 500.

The handshake now advertises `max_model_len` capped to `min(max_position_embeddings, max_request_blocks × block_size)`, carried as `EngineHandle::servable_len` and set by the qwen3 scheduler.
 
An admission layer rejects `max_tokens` over that cap with a 400 + `warn!` before the engine, reading the same `servable_len()`. Inert for model lines that don't set the cap (only qwen3 today).

## Behavior

Test Env: Qwen3-4B, single GPU. Servable cap 242016 (below the 262144 position window — the trigger).

| Request | Before | After | Decided by |
|---|---|---|---|
| `max_tokens` > cap (stream + non-stream) | 500 `Internal server error` | 400 `max_tokens (99999999) exceeds the model's maximum context length of 242016 tokens` | admission layer |
| prompt within cap | 200 | 200 | — |
| prompt ≥ cap (242785 token-ids) | 500 | 500 `...maximum context length is 242016 tokens, but the prompt contains 242785...` | vLLM `PromptTooLong` |
| other engine rejects (Kimi DP-cap, echo, TP8 logprobs) | 500 | 500 | engine `Rejected` → bridge → vLLM |

## Type of Change

## Not covered

The core #294 - ask every engine-rejected request returning a 400 that carries the engine's own message, which is not done here. EngineCoreFinishReason has no InvalidRequest variant, so a Rejected event maps to Error and vLLM renders it as a 500 with the message replaced. Closing that gap may need a vLLM protocol change.

This admission layer also can't catch long-prompt or Kimi DP-cap rejections. It runs before tokenization and has no tokenizer, so it can only check max_tokens against the cap, not the prompt's token count. Those requests still reach the engine and surface as 500s.

Refs #294

## Checklist

- [x] I have run the local test suite and all tests pass (see `CLAUDE.md`). `cargo test` passes. `--workspace --lib` needs RDMA deps absent on my work dev.


xiaguan:
Could you run a quick end-to-end check on a real qwen3 server — send a `max_tokens`-over-cap request to `/v1/completions` and confirm it returns 400 with the cap message, and that an in-cap request still returns 200? This only had CPU clippy here, the qwen3 crate wasn't built.

FeathBow:
## Environment & setup

- Model: Qwen3-4B, single GPU.
- `model` below is `qwen3-4b`.
- Servable cap on this box is 242016 tokens (`max_request_blocks × block_size`, below the 262144 position window), so the handshake advertises `max_model_len=242016`. 

Build:

```bash
cargo build --release -p pegainfer-server --bin pegainfer
```

Serve:

```bash
./pegainfer --model-path <Qwen3-4B> --served-model-name qwen3-4b --port 8000
```

## Results

1. max_tokens over the cap, non-stream → 400

```bash
curl -sw ' [%{http_code}]\n' localhost:8000/v1/completions -H 'content-type: application/json' \
  -d '{"model":"qwen3-4b","prompt":"hi","max_tokens":99999999}'
```

```json
{"error":"max_tokens (99999999) exceeds the model's maximum context length of 242016 tokens"} [400]
```

2. max_tokens over the cap, stream → 400

```bash
curl -sw ' [%{http_code}]\n' localhost:8000/v1/completions -H 'content-type: application/json' \
  -d '{"model":"qwen3-4b","prompt":"hi","max_tokens":99999999,"stream":true}'
```

```json
{"error":"max_tokens (99999999) exceeds the model's maximum context length of 242016 tokens"} [400]
```

3. In-cap request → 200

```bash
curl -sw ' [%{http_code}]\n' localhost:8000/v1/completions -H 'content-type: application/json' \
  -d '{"model":"qwen3-4b","prompt":"The capital of France is","max_tokens":16}'
```

```json
{"id":"cmpl-ffe7d4d7","object":"text_completion","created":1780906176,"model":"qwen3-4b","choices":[{"index":0,"text":":\nA. Paris\nB. Marseille\nC. Lyon\nD. T","finish_reason":"length"}],"usage":{"prompt_tokens":5,"total_tokens":21,"completion_tokens":16}} [200]
```

4. Prompt over the cap (242785 token-ids) → 500 (vLLM's `PromptTooLong`)

```bash
python3 -c "import json;print(json.dumps({'model':'qwen3-4b','prompt':[1]*242785,'max_tokens':1}))" > big.json
curl -sw ' [%{http_code}]\n' localhost:8000/v1/completions -H 'content-type: application/json' -d @big.json
```

```json
{"error":{"message":"failed to submit completion request: this model's maximum context length is 242016 tokens, but the prompt contains 242785 input tokens","type":"server_error","code":"server_error"}} [500]
```

xiaguan:
First off, nice work — turning the bare 500 left over from #209 into a 400 with a clear message, and tightening the handshake down to the real servable length, is exactly the right direction and landing spot. The behavior table is great too 👍

One thing I'd like to check with you, around `EngineHandle::with_servable_len`:

```rust
pub fn with_servable_len(mut self, servable_len: u32) -> Self {
    if let Some(inner) = Arc::get_mut(&mut self.inner) {
        inner.servable_len = Some(servable_len);
    }
    self
}
```

`Arc::get_mut` only writes when the strong count is exactly 1. Today the call site is `EngineHandle::new(..).with_servable_len(..)`, so the Arc was just created (count == 1) and it works. But if someone ever clones the handle before setting the cap (`EngineHandle` is `Clone`), or the call order shifts, this silently takes the else branch — `servable_len` stays `None`, and both the advertised handshake and the admission middleware quietly go inert, with no error and a still-green CI.

`from_parts` already writes `servable_len: None`, so maybe pass it in as a constructor argument instead — that way the "must be set before the handle is cloned" invariant is guaranteed at the type level and doesn't depend on call order. If you'd rather keep the builder style, at least swap the `if let` for `.expect(...)` so a violation fails loudly instead of silently.

One smaller, non-blocking note: the handshake advertises `min(max_model_len, cap)`, but the admission middleware checks against the bare `cap`. When `max_model_len < cap` (e.g. when `load_max_model_len` falls back to 4096) the two disagree — checking against the value you actually advertise would be more self-consistent.

Overall it looks solid; the `Arc::get_mut` one is the main thing I'd love you to take a second look at 🙏

FeathBow:
Thanks! Fixed both. 
On the `Arc::get_mut` one, passing the cap into `new` would make every other model pass `None` (only qwen3 has a cap today), so I put `servable_len` on a field outside the `Arc` instead. Same guarantee fewer call sites. Can switch if you prefer :)

FeathBow:
For another prob, a temporary log was added at the wiring site to print the advertised val vs the enforced val (log removed before merge):

1. before fix, normal config

advertise == enforce:

```text
[f-log] advertise=242016, enforce=242016
curl max_tokens=300000  →  400, max_tokens (300000) exceeds the model's maximum context length of 242016 tokens
```

2. before fix, config with max_position_embeddings removed

frontend falls back to 4096, qwen3 engine defaults to 40960:

```text
WARN  max_position_embeddings not found ...; capping max_model_len at 4096
[f-log] advertise=4096, enforce=40960
curl max_tokens=5000   →  reaches the completions handler (admitted), generates ~43s, downstream truncates at 4096  →  200, completion_tokens=4095, finish_reason=length
curl max_tokens=50000  →  400, ... exceeds ... 40960 tokens (error message uses 40960)
```

3. after fix, same removed-field config

advertise == enforce:

```text
[f-log] advertise=4096, enforce=4096 
curl max_tokens=5000   →  400, max_tokens (5000) exceeds the model's maximum context length of 4096 tokens
curl max_tokens=50000  →  400, ... exceeds ... 4096 tokens 
```

After test no regression on the behavior table.

Issue #228, opened 2026-06-04
kimi-k2: overlap MoE communication with compute on the PPLX path


xiaguan:
### Background

Each of the 60 MoE layers on the PPLX path executes dispatch → expert GEMM → combine strictly serially: communication never overlaps compute. The profile attributes a ~1.6ms/token structural bubble to this (PPLX all-to-all ~7.8ms/step where the NCCL path's equivalent communication achieves ~7.6ms with overlap).

### Current state

- The communication backend already exposes dispatch-send / dispatch-recv / combine-send / combine-recv as separate phases, so overlap is structurally possible.
- No pipelining or double-buffering across layers exists; the next layer's compute waits for the previous layer's combine.
- The optimization ledger names collectives/MoE — not GEMM tuning — as the dominant remaining decode cost at the current shape.

### Desired outcome

MoE-layer communication overlaps neighboring compute (shared-expert work and/or adjacent layers), reducing the serial bubble with measured per-step evidence.

### Acceptance boundary

- All ranks must still enter every MoE layer's collective in the same order (the variable-batch EP contract, including empty ranks) — correctness on uneven per-rank batches is part of acceptance.
- Greedy parity unchanged; gains measured at bs64 service level above the noise band.

xiaguan:
**Reframe needed (PPLX → DeepEP) + measured 2026-06-08 (8×H200, bs64 graph).** PPLX was removed in #298/#301; the serial-bubble problem persists on DeepEP, so the "PPLX path" framing in the body is stale.

nsys decode composition (graph replay, 8-rank aggregate):
- `combine_impl` (DeepEP combine a2a): 18.7% of GPU, median **87.9 µs** vs ~37 µs NVLink theory (2.4×).
- `dispatch_impl` (DeepEP dispatch a2a): 9.1% of GPU, median 14.7 µs (rank-skew tail, p99 311 µs, one 15 ms outlier).
- → MoE all-to-all is **~28% of GPU time, strictly serial** dispatch→Marlin→combine, no overlap.

Decode is GPU-bound (`cudaLaunchKernel` < 0.5%; host enqueue already gone via #227), so this a2a overlap is the top remaining decode lever. The recv/combined buffers in `moe_deepep.rs` are persistent + worst-case-sized → double-buffering layer N+1 dispatch against layer N combine is feasible. Roadmap updated in #306.

PR #308, opened 2026-06-08
fix(bench-serving): give concurrent requests distinct seeded prompts

### LINKED ISSUE #225: kimi-k2: isolate the ~33% HTTP-vs-in-process serving overhead
### Background

The same engine measures `14.39ms` TPOT in-process but `19.13ms` through `/v1/completions` at bs4 — a ~33% (4.74ms/token) serving overhead. The docs flag it as an open question. On Qwen3 this class of gap had concrete causes that are now fixed engine-side lessons: TCP_NODELAY off by default on axum + SSE/Nagle delayed ACKs, and zombie decode from missing disconnect-abort (#215).

### Current state

- The overhead is reproducible and documented (`docs/models/kimi-k2/optimization.md`, dashboard notes) but unattributed: streaming JSON serialization, the frontend bridge, and scheduler dispatch are the candidates.
- Nobody has checked whether the Qwen3-era causes apply to the Kimi path.

### Desired outcome

- The 4.74ms/token is attributed to its components with measurements, and the dominant component reduced.
- HTTP TPOT approaches the in-process number, preserving Kimi-K2's real decode advantage end-to-end for actual clients.

### Acceptance boundary

- Same-client comparison (the standard `vllm bench serve` setup) before/after.
- Findings that turn out to be server-wide (not Kimi-specific) should land server-wide and be cross-referenced, not forked.

xiaguan:
## Problem

The `request` benchmark replicated a single synthetic prompt across every concurrent worker — `synthetic_prompt_tokens` returned the same `[100, 101, …]` for all of them. Under greedy decode the streams stay byte-identical, so every step's tokens route to the **same MoE experts** and the DeepEP dispatch/combine all-to-all runs at a degenerate minimum.

For a dense model this is harmless; for an MoE model it under-measures decode TPOT by ~30%.

## Why it matters (#225)

This is the root cause of the false "+51% HTTP overhead" attribution. On 8×H200, Kimi-K2 TP1/DP8/EP8, bs64:

| in-process workload | `forward_decode_batch` p50 |
|---|---|
| 64 **identical** prompts (old behavior) | **26.4 ms** |
| 64 **distinct** prompts (this PR), no HTTP | **32.7 ms** |
| HTTP `vllm bench serve` random dataset | **33.9 ms** |

Per-step instrumentation of the DP coordinator showed the gap is **not** the serving bridge or scheduler — host work is ~0.1 ms/step and the ranks are balanced 8/8/…/8. The entire delta is inside `forward_decode_batch`, whose cost is MoE-routing-diversity-dependent. The transport adds ≈0; the in-process 26 ms simply measured a degenerate workload.

## Change

- For synthetic (`--prompt-len`) workloads, generate **one seeded random prompt per request** so concurrent streams diverge and exercise the real expert all-to-all. Reproducible from `--seed`.
- Explicit `--prompt` / `--prompt-file` and the single-stream `matrix` / `curve` / `snapshot` paths are unchanged (snapshot baselines stay stable).
- The batch contract now takes per-request prompts (`&[Vec<u32>]`) instead of one prompt + a `concurrency` count. The DeepSeek-V2-Lite same-prompt kernel keeps using the first prompt by design.

## Verification

Built and ran on 8×H200 (Kimi-K2 TP1/DP8). The report descriptor now reads `synthetic-random[100..100000) seed=42 per-request` and steady TPOT is 32.70 ms (was 26.3 ms). `cargo fmt` + `clippy` clean.

Resolves #225 — the "+51% HTTP overhead" was a benchmark artifact, not a serving cost. The real lever for the ~34 ms decode floor is the DeepEP combine (#228).

### DIFF
--- a/pegainfer-server/src/bin/bench_serving.rs
+++ b/pegainfer-server/src/bin/bench_serving.rs
@@ -33,6 +33,7 @@ use pegainfer_core::{
     parallel::ParallelConfig,
 };
 use pegainfer_vllm_support::load_tokenizer as load_vllm_tokenizer;
+use rand::RngExt;
 use rand::SeedableRng;
 use rand::rngs::StdRng;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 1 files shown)


xiaguan:
Superseded by #309 — it contains this fix plus the `--distinct-prompts` sweep knob and the measured evidence that corrects the `~30%` figure in this PR's commit message (the real routing-diversity cost is ~7–15%, entirely the Marlin expert GEMM). Closing in favor of #309.

PR #316, opened 2026-06-09
feat(kv-offload): in-process pegaflow KV-cache offload on Qwen3-4B

xiaguan:
## What

Integrate [pegaflow](https://github.com/novitalabs/pegaflow) (Apache-2.0) into pegainfer as an **in-process Rust KV-cache offload tier**, dense-attention-first, on Qwen3-4B. A prefix that has fallen out of HBM — or was never GPU-resident — can be restored from pegaflow's host (and, later, SSD/RDMA) tiers instead of recomputed.

End-to-end is live and validated on real Qwen3-4B weights: async SAVE + async LOAD wired into `Qwen3Executor` + the scheduler, with the restored KV reproducing the cold first-token logits to the bf16 floor.

## How

```
per-model scheduler   ← policy: which blocks should be resident
  ↓ load/save intents (a set of blocks)
pegainfer-kv-offload  ← the connector "brain": block identity, GPU-slot
  (OffloadEngine)        orchestration, transfer scheduling (kvbm logical/physical idea)
  ↓ semantics-free raw transfer
pegaflow-core         ← muscle: D2H/H2D + DRAM/SSD/RDMA tiers
```

- **`pegainfer-kv-offload` (new crate)** — `OffloadEngine` owns a `PegaEngine` + a small tokio runtime. `Registration::from_buffer` maps pegainfer's fused **page-first** single buffer into pegaflow's per-layer strided registration: within a page K and V are contiguous (`layer_stride` = one `[K|V]` segment), so `segments = 1` and `block_stride` decouples the page interleave — *not* the K/V-split path. Save is best-effort fire-and-forget; load is a polled, non-blocking `LoadHandle`.
- **Shared cache primitives** (`pegainfer-kv-cache`) — `KvBuffer::device_ptr`, the `KvBlockGuard` / `assigned_block_guards` strong pins, and the `PrefixProbe` / `LoadReservation` reservation path. `BlockManager::evict_inactive` gives a cold-cache flush that tolerates still-pinned blocks.
- **Qwen3-4B wiring** — `Qwen3OffloadOptions` opt-in (default off). SAVE after prefill; `begin_kv_prefetch` probes the GPU prefix, queries pegaflow, reserves + async-loads the CPU continuation, and settles it at the right prefix offset, stacking CPU blocks onto any GPU hit. The scheduler grows a non-blocking `loading` admission state so prefetch never stalls the GPU thread.

## Correctness hardenings (post toxic-review)

1. **Query-lease leak** — `release_query_lease` on the reserve/load failure paths so pinned host blocks aren't stranded for the 600s TTL.
2. **Prefetch-state leak on rejection** — `release_rejected` calls `drop_request` when a request is rejected at admission, so committed prefix blocks can't leak.
3. **SAVE D2H race** — source blocks are pinned via `KvBlockGuard` until the D2H lands, closing the "request ends → slot reused → D2H reads the wrong KV" silent-corruption window.

## Tests

- `pegainfer-kv-offload/tests/cpu_roundtrip.rs` — byte-level save → query → load on a real `KvBuffer`, plus a zero-block negative control.
- `pegainfer-qwen3-4b/tests/kv_offload_cpu_hit.rs` — live gate on real weights: one executor, two sequential scenarios (pure CPU-tier restore after HBM eviction, and a combined GPU+CPU split prefix), asserting warm logits match cold (mean Δ ≈ 0.03 nat).

Run: `PEGAINFER_TEST_MODEL_PATH=models/Qwen3-4B cargo test --release -p pegainfer-qwen3-4b --test kv_offload_cpu_hit`

## Dependency

`pegaflow-core` is a **git dependency pinned to upstream rev `07cac7e`**, which carries the two upstream changes this needs: `block_stride_bytes` (novitalabs/pegaflow#331) and the in-process load API (#333). `default-features = false` drops its `cuda-12`/`rdma`; the workspace cudarc gains `nvrtc` for pegaflow's embedded JIT copy kernel. The dead `kvbm-config`/`kvbm-engine`/`kvbm-physical` crates (older-tokio pin, unused) are dropped from the workspace.

## Scope

Default off; not yet wired into the server CLI (only `start_engine_with_offload` + tests). Qwen3.5 linear/SSM state is **excluded**; DeepSeek sparse is **deferred**. See `docs/subsystems/runtime/pegaflow-offload-integration.md`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -201,43 +201,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "async-nats"
-version = "0.45.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "86dde77d8a733a9dbaf865a9eb65c72e09c88f3d14d3dd0d2aecf511920ee4fe"
-dependencies = [
- "base64 0.22.1",
- "bytes",
- "futures-util",
- "memchr",
- "nkeys",
- "nuid",
- "once_cell",
- "pin-project",
- "portable-atomic",
- "rand 0.8.6",
- "regex",
- "ring",
- "rustls-native-certs 0.7.3",
- "rustls-pemfile",
- "rustls-webpki 0.102.8",
- "serde",
- "serde_json",
- "serde_nanos",
- "serde_repr",
- "thiserror 1.0.69",
- "time",
- "tokio",
- "tokio-rustls 0.26.4",
- "tokio-stream",
- "tokio-util",
- "tokio-websockets",
- "tracing",
- "tryhard",
- "url",
-]
-
-[[package]]
 name = "async-stream"
 version = "0.3.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -317,15 +280,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "atomic"
-version = "0.6.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a89cbf775b137e9b968e67227ef7f775587cde3fd31b0d8599dbd0f598a48340"
-dependencies = [
- "bytemuck",
-]
-
-[[package]]
 name = "atomic-polyfill"
 version = "1.0.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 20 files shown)


xiaguan:
## L2 (host-tier) TTFT benchmark + review fixes

### Pure-L2 TTFT measurement

Added `--kv-offload` / `--kv-offload-host-gib` / `--no-prefix-cache` to the Qwen3 server (`f5db7dd`) and ran `vllm bench serve` twice against the same seeded prompt set, so pass 2 finds its prefix **only** on the host tier.

**Setup:** Qwen3-4B on RTX 5070 Ti (16 GB HBM), 30 GiB host pool, 64 × 2048-token prompts, `--max-concurrency 1` (bs=1), `--no-prefix-cache` = pure-L2 mode (cross-request HBM/L1 reuse off, all reuse from L2).

| metric | Cold (first sight) | Warm (pure-L2 restore) | Δ |
|---|---|---|---|
| Mean TTFT | 195.2 ms | **40.6 ms** | **−79% (4.8×)** |
| Median TTFT | 194.8 ms | 38.1 ms | −80% |
| P99 TTFT | 203.9 ms | 54.5 ms | −73% |
| duration (64 req) | 14.9 s | 4.8 s | 3.1× |

Server logs confirm it is genuinely L2, not a residual L1 hit:

```
Prefetch local-hit: total_keys=127 hit=127 missing=0      (gpu_hit=0)
Load task completed: layers=36 blocks=127 bytes=299630592 (286 MiB)
                     elapsed_ms≈20  bandwidth≈15 GB/s  backend=direct
```

Every request restores the full 127-block prefix from the host tier (`gpu_hit=0` — the evict-before-probe path is doing its job). Warm TTFT ≈ 286 MiB H2D (~20 ms) + the 16-token tail prefill. The ~15 GB/s is consumer-PCIe-bound; a server GPU would push the restore cost down further.

**Memory:** pegaflow `cudaHostAlloc`s the whole pool at startup — server RSS measured **30.8 GB** with `--kv-offload-host-gib 30`, confirming the pool is eager, not lazy.

> Note on `--no-prefix-cache` semantics under offload: L1-resident and L2-stored blocks share one content hash, so the cache cannot be told to "prefer L2" for a block still in HBM. Pure-L2 therefore means *not retaining* completed blocks in HBM (evict before the prefetch probe → `gpu_hit=0`), which is exactly vLLM's `--no-enable-prefix-caching` behaviour (free, don't keep). Prefix *matching* stays on, since the L2 restore rides on it.

### Review fixes (`b5ac797`)

Both bot findings were real and are fixed:

- **@gemini-code-assist (HIGH — in-flight DMA corruption):** confirmed. `drop_request` released the `LoadReservation` while the H2D load could still be in flight, so a cancel-mid-load freed pages that pegaflow's worker kept writing → silent KV corruption. `drop_request` now blocks on the in-flight `LoadHandle` before the reservation drops — the load-side mirror of the SAVE keep-alive pin. The scheduler is a dedicated sync thread, so the wait is free.
- **@chatgpt-codex-connector (P1 — admission accounting):** confirmed. A settled prefetch's blocks are already out of `available_blocks`, but `admit_deferred_requests` still charged the full request footprint, so a near-budget CPU-hit request could fail admission forever and never release its prefetch. Admission now credits each request's held prefetch blocks (`prefetched_blocks` → `PrefixProbe::held_blocks`) against the free-pool budget, while keeping the full footprint for the per-request cap.

Verified: `kv_offload_cpu_hit` integration gate passes unchanged (mean Δ 0.028 / 0.030 nat), 17 scheduler unit tests pass, and an adversarial re-review found no residual holes in either fix.

xiaguan:
@codex review

xiaguan:
Fixed the P2 echo finding in `f958c57`.

**@chatgpt-codex-connector (P2 — echo + prefetch):** confirmed and fixed. Echo requests forward the whole prompt for prompt logprobs, so their prefill skips `match_and_add_prefix` and can never reuse a restored prefix. Offering them to prefetch parked restored KV that admission credited via `prefetched_blocks` but prefill never spent, starving the request under tight budgets. `offer_prefetch` now guards with `!req.echo`, so echo never probes — with no prefetch entry, `prefetched_blocks(echo)` is always `0` and admission charges the full footprint, matching the cold prefill echo actually runs. The `echo ⇒ no prefetch state ⇒ zero credit` invariant is now structural rather than a runtime patch. Added a unit test (`echo_requests_are_never_offered_to_prefetch`).

The two other latest inline comments (gemini @ `executor.rs:1045` in-flight DMA; codex @ `executor.rs:1110` admission double-count) are the original findings re-anchored to the new commit — both were already fixed in `b5ac797`.

Verified: 35 lib tests (incl. the new echo test) + `kv_offload_cpu_hit` integration gate pass; clippy clean on `scheduler.rs`; adversarial re-review found no residual holes.

@codex review

xiaguan:
Fixed the LoRA-startup P2 in `ff9fab3`.

**@chatgpt-codex-connector (P2 — LoRA startup ignored offload flags):** confirmed and fixed. `start_engine_with_lora_control` hard-coded `Qwen3OffloadOptions::disabled()` and had no `no_prefix_cache` parameter, so `--enable-lora` silently dropped `--kv-offload` / `--kv-offload-host-gib` / `--no-prefix-cache`. Now `offload_options` + `no_prefix_cache` thread through the LoRA-control path exactly as the non-LoRA path does (`offer`/`set_no_prefix_cache`), and the offload options are built once in `main.rs` for both branches.

Offload under LoRA is correct by construction — the prefix block hash is salted with the adapter name (`compute_salt_hash`, `pool.rs:90-102`, guarded by the `salt (lora) must scope the prefix cache` test), so restored KV (HBM or host tier) never crosses adapters.

Verified: server + qwen3-4b build clean, `lora_smoke` compiles against the new signature, 35 lib tests pass.

The three older inline comments now re-anchored to `ff9fab3` (gemini `executor.rs:1045` DMA, codex `executor.rs:1110` admission, codex `scheduler.rs:186` echo) are all already fixed — in `b5ac797`, `b5ac797`, and `f958c57` respectively.

@codex review

PR #317, opened 2026-06-09
fix: handle f64 extraction in kvbm-config under arbitrary_precision

### LINKED ISSUE #296: kvbm-config: workspace test run breaks figment f64 extraction via serde_json arbitrary_precision unification
\`cargo test --release --workspace --lib\` (the documented unit-test command) fails on main:

```
test tests::test_profile_with_defaults_and_overlay ... FAILED
panicked at kvbm/kvbm-config/src/lib.rs:531:89:
InvalidType(Map, "f64") at path ["cache", "host", "cache_size_gb"]
```

## Repro

```bash
cargo test --release -p kvbm-config --lib                       # green (63 passed)
cargo test --release -p kvbm-config -p pegainfer-server --lib   # red — same code
```

## Cause

`vllm-chat` (git dep of pegainfer-server via vllm-server) enables `serde_json/arbitrary_precision`:

```
cargo tree -i serde_json -e features
├── serde_json feature "arbitrary_precision"
│   ├── vllm-chat v0.1.0 (github.com/vllm-project/vllm.git rev 4aaed4c)
```

Workspace feature unification turns it on for every serde_json user. Under `arbitrary_precision`, JSON numbers deserialize as an internal `{"$serde_json::private::Number": "2.0"}` map, so figment's `Json::string(...)` source hands kvbm-config a Map where `cache_size_gb: Option<f64>` expects a float (`kvbm/kvbm-config/src/lib.rs:283` extract, `:531` unwrap in test).

Verified on main 927c7c1; unrelated to any feature branch (single-package runs are green everywhere).

## Acceptance

- `cargo test --release --workspace --lib` green on main.
- Either drop `arbitrary_precision` from the vendored vllm-chat (preferred — nothing in this workspace needs it), or make kvbm-config's figment path tolerant of Number-as-map.

mvanhorn:
## Summary

`kvbm-config`'s `cache_size_gb: Option<f64>` fields now deserialize correctly whether figment hands them a native JSON number or the `serde_json` `arbitrary_precision` "number map" form. This unblocks the documented workspace test command `cargo test --release --workspace --lib`, which was failing on `main`.

## Why this matters

Per #296, `test_profile_with_defaults_and_overlay` panics with `InvalidType(Map, "f64") at path ["cache", "host", "cache_size_gb"]`. The crate is green in isolation but red when co-compiled with `pegainfer-server`: `vllm-chat` (a transitive git dep) enables `serde_json/arbitrary_precision`, and Cargo feature unification turns that feature on for every `serde_json` user in the workspace. Under `arbitrary_precision`, JSON numbers deserialize as an internal map `{"$serde_json::private::Number": "2.0"}`, so figment's `Json` source hands `kvbm-config` a `Map` where a plain `f64` is expected.

Pinning or stripping the feature isn't viable: it's owned by a git dependency and would regress on the next bump. The robust, behavior-preserving fix is to make the float fields tolerant of either representation, so the config-load path stays correct no matter which workspace features get unified on.

## Changes

A small, self-contained `deserialize_opt_f64` serde helper was added in `cache.rs` and applied to `cache_size_gb` on both `HostCacheConfig` and `DiskCacheConfig`. It accepts a native JSON number, the `arbitrary_precision` single-entry map form, or an omitted/null value (yielding `None`). The helper uses only `serde`, so there is no dependency change and `serde_json` stays a dev-dependency.

It also rejects non-finite parses (such as `1e10000` overflowing to `inf`), matching what `serde_json`'s native f64 deserializer already does. Without this guard, a malformed config could smuggle in an infinite cache size that later casts to `usize::MAX` in `compute_num_blocks`.

## Testing

`cargo fmt --check` and `cargo clippy --all-targets` are both clean for the crate, and `cargo test --release -p kvbm-config --lib` passes 65 tests (up from 63). To confirm the fix targets the exact feature-unification path, I forced `serde_json/arbitrary_precision` on the crate's dev `serde_json`: on `main` this reproduces the failure (`test_profile_with_defaults_and_overlay` FAILED), and with this change it passes.

Two new tests cover the behavior. `test_cache_size_gb_arbitrary_precision_map` asserts the map form, a plain integer (`4` -> `4.0`), and a fractional value all extract correctly and that an omitted field yields `None`. `test_cache_size_gb_rejects_non_finite` asserts that an out-of-range value is rejected rather than becoming `Some(inf)`.

Fixes #296

### DIFF
--- a/kvbm/kvbm-config/src/cache.rs
+++ b/kvbm/kvbm-config/src/cache.rs
@@ -10,8 +10,136 @@
 
 use std::path::PathBuf;
 
-use serde::{Deserialize, Serialize};
+use serde::{
+    Deserialize, Deserializer, Serialize,
+    de::{self, MapAccess, Visitor},
+};
 use validator::Validate;
+
+const SERDE_JSON_ARBITRARY_PRECISION_NUMBER: &str = "$serde_json::private::Number";
+
+fn deserialize_opt_f64<'de, D>(deserializer: D) -> Result<Option<f64>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    deserializer.deserialize_option(OptF64Visitor)
+}
+
+struct OptF64Visitor;
+
+impl<'de> Visitor<'de> for OptF64Visitor {
+    type Value = Option<f64>;
+
+    fn expecting(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        formatter.write_str("an optional f64")
+    }
+
+    fn visit_none<E>(self) -> Result<Self::Value, E>
+    where
+        E: de::Error,
+    {
+        Ok(None)
+    }
+
+    fn visit_unit<E>(self) -> Result<Self::Value, E>
+    where
+        E: de::Error,
+    {
+        Ok(None)
+    }
+
+    fn visit_some<D>(self, deserializer: D) -> Result<Self::Value, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        deserializer.deserialize_any(F64Visitor).map(Some)
+    }
+}
+
+struct F64Visitor;
+
+impl<'de> Visitor<'de> for F64Visitor {
+    type Value = f64;
+
+    fn expecting(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        formatter.write_str("an f64 or serde_json arbitrary_precision number map")
+    }
+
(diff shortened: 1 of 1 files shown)


xiaguan (approved):
LGTM. Verified fmt plus kvbm-config lib tests and the kvbm-config + pegainfer-server feature-unification repro locally.

mvanhorn:
Thanks @xiaguan for landing this. f64 extraction behaving under arbitrary_precision closes out that kvbm-config figment trap.

PR #333, opened 2026-06-10
fix(qwen35): chunk prefill so GDR scratch stays bounded (#314)

### LINKED ISSUE #314: qwen35: prefill scratch is unbounded after MAX_SEQ removal (can OOM mid-prefill)
Follow-up from #305 (review), under the memory-bounded admission workstream.

Removing `MAX_SEQ` (#305) also dropped the cap that incidentally bounded the prefill **scratch**. `GdrChunkwiseScratch35::estimate_bytes` scales linearly with `seq_len`, but `weights.rs` only reserves `SCRATCH_ESTIMATE_SEQ = 20_000`, and the scheduler's admission bounds **KV pages, not scratch**.

Consequence: a prompt that fits the KV pool but runs well past ~20k tokens can OOM mid-prefill inside `GdrChunkwiseScratch35::new` — and because this happens **after `kv_state.advance`**, `kv_state` is left advanced on the error path instead of the request being rejected up front.

#252 explicitly defers memory-bounded admission to the roadmap, so this is a follow-up, not part of #305.

Refs #252, #305.

wjinxu:
## Summary

Fixes #314. Qwen3.5 prefill now runs in serial chunks of at most `PREFILL_CHUNK_LEN` tokens instead of one pass over the whole prompt, so the per-pass GDR scratch can no longer exceed the memory reserved at startup.

## Problem

After `MAX_SEQ` was removed (#305), nothing bounded the prefill scratch. `GdrChunkwiseScratch35` scales linearly with the pass length, but startup only reserves scratch for `SCRATCH_ESTIMATE_SEQ` (20k) tokens and the scheduler's admission bounds KV pages, not scratch. A prompt that fits the KV pool but runs past ~20k tokens could therefore OOM mid-prefill inside `GdrChunkwiseScratch35::new` — and because that happened *after* `kv_state.advance`, the KV state was left advanced on the error path.

## Approach

`prefill_forward` splits the prompt into chunks of `PREFILL_CHUNK_LEN` (= `SCRATCH_ESTIMATE_SEQ`, so the existing reservation covers exactly one chunk) and forwards them serially through a new `prefill_chunk_forward` helper:

- Each chunk advances the paged KV and the linear-attention recurrent/conv state in place, so chunk N+1 continues exactly where chunk N left off (recurrent state and conv state are carried in/out; full attention uses the per-chunk `base_pos`).
- Only the final chunk runs the last-token norm + LM head.
- The previous chunk's hidden states are freed before the next chunk allocates, so peak memory stays within one chunk's reservation.
- Position/RoPE range is validated up front for the whole prompt, before any chunk mutates state.

Prompts longer than one chunk now prefill correctly instead of OOM-ing or being rejected. The chunk granularity is bounded, so scratch is bounded.

> Note: KV-pool admission for very long prompts is still out of scope here and
> tracked separately (#252).

## Verification

HF golden gate, comparing chunked vs. single-pass against the HF oracle
(tolerance: mean ≤ 0.06, p99 ≤ 0.20, argmax must stay near HF's best):

| Config | Test | mean | p99 |
|---|---|---|---|
| single pass (chunk = 20000) | long-golden (4097/8192-tok prompts) | 0.018 | 0.064 |
| chunk = 2000 (8192 → 5 passes) | long-golden | 0.022 | 0.067 |
| chunk = 16 (all prompts multi-pass) | golden (seq / batched / slot-compaction) | 0.025–0.027 | ≤ 0.092 |

Chunked output matches single-pass within tolerance (small drift from floating-point reduction-order differences at chunk boundaries; argmax never violated). `e2e_scheduler` also passes at both the default chunk size and a forced chunk size of 4 (sequential / multi-request / concurrent).

## Follow-up (not in this PR)

Prefill scratch is still re-allocated per chunk. If this becomes a hotspot on long-context prefill, scratch reuse should be owned by the execution/scheduler context (alongside the decode graph and sample scratch), not pinned on the model — a model-level singleton would bake in a serial, non-preemptible assumption. Chunk boundaries are intended as clean checkpoints for future KV-pressure handling (preempt/evict another request when the KV pool fills mid-prefill, then resume), where only the per-request recurrent/conv state and KV pages need to survive — the transient scratch does not.

### DIFF
--- a/openinfer-qwen35-4b/src/prefill.rs
+++ b/openinfer-qwen35-4b/src/prefill.rs
@@ -4,8 +4,18 @@ use cudarc::driver::{CudaSlice, DevicePtr, DevicePtrMut};
 /// Sequence length used for conservative prefill scratch reservation.
 ///
 /// This is not an admission cap. Actual prompt admission is governed by the
-/// paged KV pool, RoPE cache coverage, and allocation success.
+/// paged KV pool, RoPE cache coverage, and allocation success. Prompts longer
+/// than this are handled by chunking prefill at `PREFILL_CHUNK_LEN` rather than
+/// being rejected (see `prefill_forward`).
 pub(crate) const SCRATCH_ESTIMATE_SEQ: usize = 20_000;
+
+/// Maximum number of tokens processed in a single prefill forward pass.
+///
+/// Prefill is chunked at this granularity so the per-pass GDR scratch
+/// (`GdrChunkwiseScratch35`, which scales linearly with the pass length) never
+/// exceeds the memory reserved at startup. Kept equal to `SCRATCH_ESTIMATE_SEQ`
+/// so the reservation in `weights.rs` covers exactly one chunk.
+pub(crate) const PREFILL_CHUNK_LEN: usize = SCRATCH_ESTIMATE_SEQ;
 const HEAD_DIM: usize = 256;
 
 use super::prefill_buffers::GdrChunkwiseScratch35;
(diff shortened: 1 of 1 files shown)


xiaguan (approved):
LGTM. Reviewed the prefill chunking state flow and verified locally with qwen35 scheduler/golden coverage.

PR #335, opened 2026-06-10
fix: reserve full lifetime KV blocks at admission to match kvbm peak draw

### LINKED ISSUE #288: qwen3: KV admission under-reserves by 1 block at boundary alignments
## Problem

The qwen3 full-lifetime reservation (#85) computes a request's lifetime KV need as `blocks(prompt + max_tokens - 1)` (`pegainfer-qwen3-4b/src/scheduler.rs:455` `max_active_tokens`). But kvbm's true peak pool draw is `ceil((prompt + max_tokens) / block_size)`: every `apply_decode` — including the final token's — appends the new token to the sequence, and `schedule_decode` provisions its block, even though the final token's KV is never written.

Probed against the real `BlockManager` (block_size 16):

| prompt | max_tokens | peak draw | reserved | |
|---|---|---|---|---|
| 16 | 17 | 3 | 2 | under |
| 1 | 16 | 2 | 1 | under |
| 17 | 16 | 3 | 2 | under |
| 33 | 100 | 9 | 9 | ok |

When `prompt + max_tokens ≡ 1 (mod 16)`, admission under-reserves by 1 block. With the pool genuinely full, the final `schedule_decode` of such a request fails → request killed with an accounting-violation error mid-decode.

## Repro

Probe harness: walk `RequestKv` through `schedule_prefill/apply_prefill` + `schedule_decode/apply_decode` to `max_tokens` against a fresh `BlockPool`, track `base - available_blocks()` peak vs `blocks(prompt + max_tokens - 1)`.

## Fix

Same as kimi-k2 (#239 PR): reserve `ceil((prompt + max_tokens) / block_size)` — see `request_lifetime_blocks` in `pegainfer-kimi-k2/src/runner/scheduler/lifecycle.rs`. One-line change to `max_active_tokens` plus test updates (`scheduler.rs:914` comment encodes the old arithmetic).

## Acceptance

- Admission reserves the empirical peak draw; probe shows no under-reservation across boundary alignments.
- A full-pool e2e with `prompt + max_tokens ≡ 1 (mod 16)` completes instead of erroring on the final decode step.

mvanhorn:
## Summary
KV admission under-reserved: it budgeted blocks for `prompt_len + max_tokens - 1` written tokens, but kvbm's `schedule_decode` provisions the final dangling token's block even though that token's KV is never written. Under tight pool capacity a request could pass admission and then fail mid-decode when the pool could not supply that last block. Admission now reserves full lifetime blocks, `ceil((prompt_len + max_tokens) / block_size)`, with a carve-out for one-token completions, which never call `schedule_decode` and only need their prompt blocks.

## Why this matters
The mismatch in #288 is between two accounting models: admission counted tokens whose KV is actually written, while the block pool's draw is driven by `schedule_decode`'s provisioning, which includes the dangling token. At page boundaries (for example `prompt_len=16, max_tokens=17, block_size=16`) the two models disagree by exactly one block, which is the failure the issue describes. The new `request_lifetime_blocks` helper is the single source for both pending-request admission (`admit_deferred_requests`) and active-request growth reservation (`active_future_blocks`), so the two paths can no longer drift apart.

## Testing
Unit tests cover the boundary cases from the issue: a `16+17` request is rejected with 2 blocks and admitted with 3, one-token page-boundary completions still fit their prompt blocks, and a new `kvbm_peak_draw` probe drives a real `BlockPool` through prefill/decode/release and asserts the reservation matches the pool's actual peak draw for `(16,17)`, `(1,16)`, and `(17,16)`. `cargo fmt --check` is clean; the full test build needs CUDA for `openinfer-kernels`, which this machine lacks, so the scheduler tests are written to run in CI.

Fixes #288

### DIFF
--- a/openinfer-qwen3-4b/src/scheduler.rs
+++ b/openinfer-qwen3-4b/src/scheduler.rs
@@ -641,15 +641,17 @@ fn current_active_tokens(req: &ActiveRequestState) -> usize {
         .saturating_add(req.generated_count.saturating_sub(1))
 }
 
-// Pool blocks a request can draw over its lifetime. kvbm's schedule_decode
-// provisions the final dangling token's block even though that token's KV is
-// never written, so admission must reserve prompt + max_tokens, not just the
-// written-KV token count.
+// Pool blocks a request can draw over its lifetime. One-token completions
+// finish after prefill, so schedule_decode never provisions a dangling block.
+// Multi-token requests can draw that final dangling decode block, so admission
+// reserves prompt + max_tokens for them.
 fn request_lifetime_blocks(prompt_len: usize, max_tokens: usize, block_size: usize) -> usize {
-    prompt_len
-        .saturating_add(max_tokens)
-        .div_ceil(block_size)
-        .max(1)
+    let lifetime_tokens = if max_tokens <= 1 {
+        prompt_len
+    } else {
+        prompt_len.saturating_add(max_tokens)
+    };
+    lifetime_tokens.div_ceil(block_size).max(1)
 }
 
 fn pending_lifetime_blocks(req: &PendingRequest, block_size: usize) -> usize {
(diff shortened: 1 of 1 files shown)


xiaguan (approved):
LGTM. The under-reservation is real: the new peak-draw probe drives the actual BlockPool/RequestKv lifecycle and covers the boundary cases from #288. The max_tokens=1 carve-out also keeps the one-token completion path from over-reserving.

mvanhorn:
Thank you @xiaguan for landing this. Reserving full-lifetime KV blocks at admission keeps kvbm honest about capacity.

Issue #244, opened 2026-06-04
qwen3: mixed-load latency profile + the chunked-prefill decision


xiaguan:
### Background

Qwen3-4B's scheduler admits prefills into the unified step with no per-step token budget: a long prompt arriving mid-decode runs as one uninterrupted prefill while every decoding request waits. Benchmarks against vLLM showed ITL p99 noticeably worse under mixed load, consistent with this. The maintainer position (see `docs/subsystems/scheduler/`) is that chunked prefill is *not* automatically the answer — measure first.

### Current state

- No tracked benchmark profile exercises long-prompt arrivals during steady-state decode, so the tail cost is anecdotal.
- The committed bench snapshots predate prefix caching (#216) and ~190 commits of changes.

### Desired outcome

- A tracked mixed-load profile (long prompts at low QPS over a decode-heavy steady state) reporting ITL p50/p99, refreshed alongside the existing snapshots.
- A data-backed go/no-go on chunked prefill recorded in the docs — implementation only if the measured tail matters for a realistic workload.

### Acceptance boundary

- Decision recorded with numbers (p50 **and** p99); chunked prefill itself is out of scope for this issue.

scatyf3:
I want to solve this 👀

CAICAIIs:
> I want to solve this 👀

just go ahead~

CAICAIIs:
Completed by #368.

The mixed-load ITL profile is now tracked in bench_serving and the results/decision are recorded in docs/benchmarks/mixed-load-itl.md. Chunked prefill itself was explicitly out of scope for this issue; the Qwen3 implementation landed separately in #363, and the Qwen3.5 follow-up is tracked in #375.

PR #340, opened 2026-06-10
refactor(bench): split bench_serving.rs into modules

scatyf3:
## Motivation

To solve Issue #244, we want to add a mixed-load benchmark to bench_serving.rs. But this file is already ~2000 LOC, and the mixed-load code would push it to ~3000(currently in  [my fork]([pegainfer/pegainfer-server/src/bin/bench\_serving.rs at feat/qwen3-mixed-load-itl · scatyf3/pegainfer · GitHub](https://github.com/scatyf3/pegainfer/blob/feat/qwen3-mixed-load-itl/pegainfer-server/src/bin/bench_serving.rs))) — which raises two issues:
1. **Hard to read / review** — we split a file once it grows past ~1000 LOC.
2. **Shared utilities get duplicated** — mixed-load and the existing benchmark
   code drive the scheduler the same way: submit a request, then drain its
   token stream. That loop is already copy-pasted between `timed_generation`
   and `timed_generation_batch`, and mixed-load would copy it three more times
   (`spawn_background_streams`, `run_injector`, `mixed_warmup`). Extracting it
   once lets both old and new code share it instead of duplicating.

so the first step is to refactor `bench_serving.rs`

## Change

refactor `bench_serving.rs` 

| module | LOC | responsibility |
|--------|----:|----------------|
| `main.rs` | 334 | orchestration: model loader, `dispatch`, `main`, tests |
| `cli.rs` | 231 | `Cli` / `Command` / `*Args` |
| `report.rs` | 171 | serde report + metric types |
| `metrics.rs` | 74 | percentiles, stats, token traces |
| `render.rs` | 295 | comfy_table renderers |
| `prompt.rs` | 124 | prompt resolution / synthetic / salting |
| `exec.rs` | 408 | `BenchModel`, `run_timed`, `run_scheduler_stream`, adapters |
| `runners.rs` | 452 | request/matrix/curve drivers + report emission |
| `snapshot.rs` | 379 | snapshot generation + git-baseline compare |

## Verification

`cargo fmt --check` clean; default build clean.

## Next Step

implemment mixed-load ITL based on current refactored codebase


xiaguan:
LGTM. Could u rebase or merge latest main?

PR #368, opened 2026-06-11
feat(bench): mixed-load ITL profile + chunked-prefill decision (#244)

scatyf3:
## Description

Issue #244. Adds a mixed-load ITL profiler — long prompts arriving into a steady-state decode batch — and a data-backed go/no-go on chunked prefill.

What's in it:

- bench_serving mixed — an open-loop driver: N long-lived background decode streams + an injector firing a long prompt every 1/qpss, tagging each background inter-token gap as stall (overlaps an in-flight prefill) or steady, reported vs a decode-only baseline.
- Folded into the snapshot pipeline as a tracked mixed_itl profile in bench_snapshots/{gpu}/{model}.json, refreshed alongside prefill_heavy/decode_heavy. compare prints its p50/p99 delta for context but does not gate on it — the stall tail is thermally/run-to-run noisy.
- Doc: docs/benchmarks/mixed-load-itl.md — the sweep, the severity×frequency model, and the decision matrix.

TL;DR — mixed-load tail cost (ITL p99) is a function of three knobs: prefix-cache hit ratio, QPS, and prefill length.

A prefill admitted mid-decode freezes every active decode for its whole duration (the stalled gap ≈ prefill wall-time: 4k ≈0.5s, 8k ≈1.2s, 12k ≈2.2s); p50/steady stay at baseline — it's a pure tail effect.
It only reaches p99 once stalls exceed ~1% of decode gaps. Low QPS + moderate prompt, or any prefix reuse (warm), keeps p99 at baseline (~15ms). But ≥1 req/s or cold ≥8k prompts blow p99 to 0.5–3.8s (30–250× baseline) — there a hard ITL-p99 SLA can't survive.

## Type of Change

New feature (non-breaking change which adds functionality) and Documentation update


xiaguan (approved):
LGTM — the per-gap stall/steady attribution is exactly the measurement #244 asked for, and the severity x frequency framing makes the tail cost easy to reason about. The tool stays useful well beyond the original question.

One heads-up: for the 0.1.0 push we ended up needing chunked prefill sooner than planned — #362 implements it for Qwen3-4B (per-step prefill token budget, default 1024), which bounds the stall to a single chunk's wall-time instead of the full prefill. Your sweep is the pre-chunking record that motivated that default; once #362 lands we'll re-run the mixed profile and refresh the doc's TL;DR (the decision shifts from "whether to chunk" to "how to pick the budget").

If you're still interested in this area, Qwen3.5-4B would be a great next target for the same chunked-prefill treatment — it's one of the models that matters most for 0.1.0, and its scheduler doesn't have a prefill budget yet. Happy to point you at the relevant pieces if you want to take it.

scatyf3:
Thanks, I'd definitely like to take this on! I've opened [issue](https://github.com/openinfer-project/openinfer/issues/375) to scope out scheduler-level chunked prefill for Qwen3.5 — it captures the current status (KV-cache-level chunking already lands in prefill.rs, but the scheduler still prefills the whole prompt in one step, so it hits the same mixed-load ITL issue this PR measured) and the plan to port the #363 approach over.

Would appreciate any pointers on the relevant scheduler pieces when you get a chance. 🙏

PR #370, opened 2026-06-12
fix(qwen35): reject requests past context window

### LINKED ISSUE #253: qwen35: RoPE cache covers 4096 positions but config admits 262144
### Background

Sibling of #220 (the Qwen3-4B instance of the same bug). The Qwen3.5 full-attention layers precompute a RoPE cos/sin cache at load time with a hard-coded length of 4096 positions, while the model config declares `max_position_embeddings: 262144`.

### Current state

- `pegainfer-qwen35-4b/src/weights.rs` builds the cache from a 4096 literal.
- Positions past the cache are an out-of-bounds read on the kernel side — the same failure shape as #220.
- The checkpoint ships `rope_scaling: null`, so sizing the plain precompute from config is the correct fix for now; rope-scaled checkpoints (#8, YaRN) would need the schedule-aware version.

### Desired outcome

- The cache is sized from config (bounded by a sane serving limit), and admission rejects prompts past the cache length instead of silently corrupting.
- Coordinated with the #220 fix so both crates share the pattern.

### Acceptance boundary

- A >4096-position request either works with correct logits or is rejected with a clear error — never silent garbage.

CAICAIIs:
## Summary

Fixes #253.

This adds the Qwen3.5 admission-side guard for requests that would run past the model position window. The RoPE cache already follows `max_position_embeddings`; this PR makes over-window requests fail closed before prefill/decode.

- reject `prompt + max_tokens > max_position_embeddings` during scheduler admission
- keep context-window rejection separate from KV-budget rejection
- report the Qwen3.5 servable length through `EngineHandle`
- cover the new rejection path in unit tests and the scheduler e2e test
- update the Qwen3.5 roadmap entry

## Verification

```bash
cargo +nightly fmt --check
git diff --check
```

```bash
cargo test --release -p openinfer-qwen35-4b --features qwen35-4b --lib -- --nocapture
```

Result: `23 passed`.

```bash
OPENINFER_TEST_MODEL_PATH=<Qwen3.5-4B> \
OPENINFER_TEST_MODEL_REVISION=851bf6e806efd8d0a36b00ddf55e13ccb7b8cd0a \
```

Result: `2 passed`, including the 4097/8192-token long-prompt HF logits gate.

```bash
OPENINFER_TEST_MODEL_PATH=<Qwen3.5-4B> \
```

Result: `1 passed`, including the new over-context-window rejection path.

## Scope

This PR covers the Qwen3.5 context-window admission path. It does not claim a serving throughput improvement.


xiaguan (approved):
LGTM. Reviewed the scheduler admission chain and ran the Qwen3.5 scheduler unit/e2e gates locally.

PR #376, opened 2026-06-13
feat(bench): prefill + decode scheduler sweeps with KV-capacity guard

xiaguan:
## What

Two new `bench_serving` subcommands that measure **cold-prefill TTFT** and **steady-state decode TPOT** through the cross-model `EngineHandle` — the only abstraction every model line shares (executor types are per-crate; unifying them would be over-abstraction). Plus one minimal shared interface: `EngineHandle::kv_capacity_tokens`.

### `prefill` — TTFT across `prompt_len × batch`
- Emits one token per request (`output_len=1`) so TTFT *is* the prefill latency, no decode mixed in.
- Every prompt is drawn from a **sweep-global monotonic salt** so none repeats — across cells, iterations, or requests. A repeat would hit the prefix cache and inflate throughput (this caught a bogus ~60k tok/s on a 5070 Ti before the fix; physically impossible vs the bf16 TFLOP ceiling).

### `decode` — TPOT across `ctx × batch`
Decode TPOT must exclude prefill. Each cell runs two rounds:
1. **warm** (`max_tokens=1`) — populates the prefix cache with the ctx-length KV, not timed;
2. **measure** — resubmits the same prompts; prefill now hits the cache, so all requests enter decode together and the timed inter-token gaps are pure decode.

Each stream **verifies the hit** via `Scheduled.cached_tokens ≈ ctx` and bails otherwise — a prefill-contaminated TPOT can never be silently reported. TBT skips the first `--warmup-steps` gaps (graph-capture ramp).

### Capacity guard
Both sweeps **refuse to run** if any user-specified cell exceeds the KV pool (error-and-run-nothing, lists the offending cells), so the user never hand-computes per-token KV:
- prefill peak = `prompt_len × batch`
- decode peak = `batch × (ctx + decode_steps)`

Capacity comes from the new `EngineHandle::kv_capacity_tokens`, wired on qwen3 from `available_blocks × block_size` (conservative by the one leaked CUDA-graph padding block). Other model lines report `None`; the guard then downgrades to a warning.

## Validation (RTX 5070 Ti, Qwen3-4B)

decode (cache-hit verified, no bail):
| ctx | bs | tpot_p50 | tpot_p99 | decode_tok/s |
|----:|---:|---------:|---------:|-------------:|
| 512 | 1 | 11.45 | 11.54 | 87 |
| 512 | 16 | 13.15 | 18.42 | 1217 |
| 2048 | 1 | 11.23 | 11.31 | 89 |
| 2048 | 16 | 17.79 | 22.42 | 899 |

Single-stream ~11.2 ms matches the known Qwen3-4B decode TPOT → confirms prefill is excluded. Batch scaling is the memory-bound signature (TPOT ~flat, throughput ~14× at bs16).

prefill (cold, bs=1): 512→10.8k, 2048→10.3k, 8192→8.4k tok/s (drops as O(L²) attention kicks in) — all under the physical TFLOP ceiling.

## Notes
- The cold-prompt salt-draw lives in one helper (`prompt::draw_distinct_prompts`) shared by both sweeps, so the invariant cannot drift.
- p99 + max in every table per bench convention.
- Toxic-reviewer pass: no blocking correctness/perf findings; stale "free == total" comments and the duplicated salt-tiling (both flagged) are fixed in this branch.

## Run
```
cargo run --release -p openinfer-server --bin bench_serving -- \
  --model-path models/Qwen3-4B decode
cargo run --release -p openinfer-server --bin bench_serving -- \
  --model-path models/Qwen3-4B prefill
```

### DIFF
--- a/openinfer-engine/src/engine.rs
+++ b/openinfer-engine/src/engine.rs
@@ -176,6 +176,11 @@ pub fn unix_now_s() -> f64 {
 pub struct EngineHandle {
     inner: Arc<EngineInner>,
     servable_len: Option<u32>,
+    /// Total KV pool capacity in tokens (`total_blocks × block_size`). Lets a
+    /// caller (e.g. the prefill bench) decide up front whether a batch fits,
+    /// instead of computing per-token KV by hand. `None` if the engine did not
+    /// report it.
+    kv_capacity_tokens: Option<usize>,
 }
 
 struct EngineInner {
@@ -224,6 +229,7 @@ impl EngineHandle {
                 join_handle,
             }),
             servable_len: None,
+            kv_capacity_tokens: None,
         }
     }
 
@@ -235,6 +241,19 @@ impl EngineHandle {
 
     pub fn servable_len(&self) -> Option<u32> {
         self.servable_len
+    }
+
+    #[must_use]
+    pub fn with_kv_capacity_tokens(mut self, kv_capacity_tokens: usize) -> Self {
+        self.kv_capacity_tokens = Some(kv_capacity_tokens);
+        self
+    }
+
+    /// Total KV pool capacity in tokens, if the engine reported it. A batch
+    /// whose summed `prompt + generated` tokens exceed this cannot be resident
+    /// at once.
+    pub fn kv_capacity_tokens(&self) -> Option<usize> {
+        self.kv_capacity_tokens
     }
 
     #[allow(clippy::result_large_err)]
--- a/openinfer-engine/src/engine.rs
+++ b/openinfer-engine/src/engine.rs
@@ -172,15 +172,42 @@ pub fn unix_now_s() -> f64 {
         .as_secs_f64()
 }
 
+/// KV pool capacity as the scheduler actually allocates it: whole blocks of
+/// `block_size` tokens. A request of `L` tokens occupies `⌈L / block_size⌉`
+/// blocks no matter how `L` divides, so a fit check must round per request —
+/// summing raw token counts under-counts and can admit a batch that the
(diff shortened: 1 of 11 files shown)


xiaguan:
Both findings addressed in e2904f2.

**P2 — block-round capacity checks (prefill.rs / decode.rs).** Right call: the scheduler allocates whole blocks per request, so the guard now compares `batch × ⌈len / block_size⌉` against `total_blocks` instead of summing raw tokens. To do this without a fuzzy two-Option contract, the engine now exposes `KvCapacity { total_blocks, block_size }` (with `total_tokens()` / `blocks_for()`) in place of the `kv_capacity_tokens` scalar that had erased block granularity; the JSON report still surfaces total tokens. The block math matches the scheduler's `request_lifetime_blocks` reservation exactly for both paths (prefill submits `max_tokens=1`; decode submits `decode_steps`, which folds in the eager boundary block).

Verified: cell `2401×16` (38416 tok ≤ 38496, but 2416 blocks > 2406) is now rejected where the raw-token guard would have admitted it → OOM.

**P2 — decode hit check vs non-reporting schedulers (decode.rs).** Confirmed: kimi `send_scheduled` hardcodes `cached_tokens: 0` (prefix-usage reporting is a follow-up) and qwen3.5 emits no `Scheduled` at all. The check now requires `cached_tokens > 0 && cached_tokens + CACHE_HIT_SLACK >= ctx`, so those lines fail loudly with a clear message instead of letting small `ctx` masquerade as a hit and report a prefill-polluted TPOT. qwen3 (reports cached ≈ ctx > 0) is unaffected.

PR #392, opened 2026-06-16
fix(qwen35): batch step-tail logits and sampling

### LINKED ISSUE #353: Per-request step-tail (lm_head GEMV / sampling / blocking DtoH) in other model crates — same bug class as the Qwen3-4B regression (#345)
## Context

The root cause of the Qwen3-4B #345 regression ("decode doesn't scale past bs≈64") was **not** decode-kernel scaling — the backbone is fully batched and beats vLLM at bs≥32. It was the **step tail**: after the batched forward, logits extraction and sampling ran as a per-request loop:

```
for each request:
    extract_vec        # slice 1 column out of the batch
    rms_norm           # single-row norm
    linear             # lm_head GEMV 1×2560 · 2560×151936 (~0.48 ms each)
    argmax/sample      # single-row sampling
    1 KB pageable DtoH # "fake async" — blocks the stream every iteration
```

nsys attribution at bs≈120: **~57 ms per unified step, ~25 % of total GPU time**, growing linearly with batch size — which is exactly why it presented as "batch doesn't scale". The fix (branch `perf/qwen3-batched-step-tail`, PR to follow) replaces the loop with gather → batched RMSNorm → **one** GEMM → `select_batch_tokens_into` → one DtoH, cutting QPS-10 TPOT from 86.7 ms to 21.4 ms on an RTX 5090.

I audited the other model crates for the same pattern. Findings below, worst first.

## Qwen3.5-4B (`openinfer-qwen35-4b`) — same bug, amplified

- `src/executor.rs` (`execute_decode`, ~L241): per request, `extract_vec` then `token_and_logprob` → `DeviceVec::to_host` copies the **full 151 936-entry f32 vocab row (~600 KB) per request per decode step, blocking**, followed by CPU argmax/log-softmax. At bs=64 that is ~38 MB of synchronous DtoH per step — heavier than the Qwen3-4B tail that caused #345.
- `src/unified_forward.rs` (`unified_step`, ~L88): decode logits are split into per-request `DeviceVec`s (n× DtoD copies) before sampling.
- `src/prefill.rs` (prefill tail, ~L82): per-request extract + norm + lm_head GEMV. Bounded today because GDR chunkwise prefill is serial per request anyway, but it inherits the pattern.
- Note: the pure-decode scheduler path (`src/scheduler.rs` ~L540, `select_tokens_batch_varied`) is already batched and only falls back to host copies when logprobs are requested — the trait-executor path above is the one with the loop.

**Fix**: direct port of the Qwen3-4B batched tail (gather + batched RMSNorm + single GEMM + `select_batch_tokens_into`; host logprobs only for requests that asked for them).

## DeepSeek-V4 (`openinfer-deepseek-v4`) — host-side tail on the direct path

- `src/direct/scheduler.rs` (~L424): `next_logits` is a host `Vec<f32>` (full-vocab DtoH per request per step) and token selection is `argmax_f32` **on the CPU**. Same family (blocking host tail), per-token cost is the largest of all crates. The direct greedy path is research-grade so this is lower priority, but it caps any future batched serving on this model line.

## DeepSeek-V2-Lite (`openinfer-deepseek-v2-lite`) — batched sampling ✓, host-side norm ✗

- `src/runtime/layers.rs` (`sample_tokens`, ~L467): batched lm_head (`gemm_per_token`) + `argmax_batch_bf16_into` — good.
- But `rms_norm_hidden` (~L507) round-trips the hidden batch through **host f32** (`hidden_to_f32` + CPU RMS norm) on every decode step before the lm_head.
- Prefill tail (`sample_last_token`) is per-request GEMV + argmax — bounded while prefill is serial.

## Kimi-K2 (`openinfer-kimi-k2`) — decode tail already batched ✓

- Decode (`src/runner/worker/forward.rs` ~L101): batched `rms_norm_into` → one graph-safe GEMM → `launch_local_top1_batch`. This is the reference implementation; no action.
- Prefill (`src/runner/worker/state.rs` ~L608): lm_head GEMM is computed over **all prompt positions** and only the last row is kept — O(seq_len × vocab_shard) wasted FLOPs per prefill. Compute the lm_head on the gathered last position only.

## Suggested order

1. Qwen3.5-4B decode/unified tail (same fix as Qwen3-4B, biggest serving impact)
2. DeepSeek-V2-Lite host-side RMS norm in the decode tail
3. Kimi-K2 prefill last-position lm_head
4. DeepSeek-V4 direct-path host tail

Refs: #345, #341.

CAICAIIs:
## Summary

Fixes #353.

- Batch the Qwen3.5 step tail: gather each request's last hidden row, run offset RMSNorm in batch, and use one lm_head GEMM.
- Sample decode/unified rows directly from batched logits.
- Keep full-vocab host snapshots only for requests with `logprobs > 0`.
- Add scheduler join ownership to avoid a CUDA/cuBLAS teardown race seen in high-concurrency serving runs.
- Add scheduler e2e coverage for `logprobs=0/1` greedy parity, logprob payload shape, and mixed concurrent logprob/no-logprob requests.
- Update Qwen3.5 docs with concise validation and benchmark boundaries.

## Validation

- `cargo fmt --all --check`
- `git diff --check`
- RTX 5090 / CUDA 12.8:
  - `cargo check --release -p openinfer-qwen35-4b --features qwen35-4b --tests`
  - `cargo test --release -p openinfer-qwen35-4b --features qwen35-4b --test hf_golden_gate -- --nocapture`
  - `cargo test --release -p openinfer-qwen35-4b --features qwen35-4b --test e2e_scheduler -- --nocapture`

Results:
- `hf_golden_gate`: 2 passed
- `e2e_scheduler`: 1 passed

## Benchmark

Same RTX 5090, same model, same seed.

Short-output serving A/B:

`bench_serving request --prompt-len 1 --output-len 1 --concurrency 4 --warmup 5 --iters 20 --seed 353`

- Token hashes matched.
- TTFT avg: `30.531 ms -> 29.221 ms` (`-4.29%`)
- E2E avg: `30.533 ms -> 29.224 ms` (`-4.29%`)
- Request tok/s: `32.751 -> 34.219` (`+4.48%`)

Long-output sanity:

- `output_len=128`, `concurrency=8`: token hashes matched, both exited 0, steady TPOT `9.0303 ms -> 9.0247 ms`.
- `output_len=128`, `concurrency=64`: token hashes matched, branch exited 0 after the scheduler join fix, steady TPOT `9.0412 ms -> 9.0494 ms`.

No meaningful long-decode TPOT speedup is claimed.


openinfer-qwen35-4b/src/logprobs.rs
```diff
@@ -0,0 +1,71 @@
+use anyhow::Result;
+use openinfer_core::engine::TokenLogprob;
+use openinfer_core::tensor::{DeviceContext, HiddenStates};
+
+pub(crate) fn snapshot_requested_logprobs(
```
xiaguan, line 5: It appears that each model currently rolls its own logits module, and sampling is not tightly coupled to the model. Perhaps we could introduce a crate called openinfer-sample to house the core inference sampling logic and kernels.

xiaguan (approved):

Issue #328, opened 2026-06-10
docs(readme): update Qwen3.5 performance against latest vLLM


xiaguan:
Split from #122.

Update the README with current Qwen3.5 performance numbers and a comparison against the latest vLLM baseline we want to target. The README numbers should be traceable to benchmark snapshots or docs, and the section should make clear what workload, GPU, and configuration were measured.

PR #411, opened 2026-06-17
fix(qwen35): add __syncthreads() to close the warp_norms WAR race in GDN decode

### LINKED ISSUE #410: qwen35: GDN decode kernel has a shared-memory write-after-read race on warp_norms (nondeterministic decode under two concurrent CUDA streams)
## Problem

`gated_delta_rule_decode_kernel` in `openinfer-kernels/csrc/qwen35/gated_delta_rule.cu` reuses the `warp_norms` shared buffer for the q-norm and the k-norm. Thread 0 reads `warp_norms[0..3]` for the q-norm (line 73), then each warp's lane 0 overwrites `warp_norms` with the k partial-sums (line 84) — with no `__syncthreads()` between the read and the overwrite. That is a write-after-read (WAR) race: under two co-resident CUDA streams the altered warp schedule can let the overwrite land before thread 0's read, so thread 0 reads a k value as the q-norm → corrupted q-normalization → nondeterministic decode. Single-stream the read always wins, so it is benign in production; the race is UB regardless of stream count (CUDA does not order warps within a block), but only manifests as divergent output under concurrency.

## Behavior

Bit-different decode output across two identical runs when a second `Qwen35Executor` (a second CUDA stream) is co-resident on one GPU; a single stream is deterministic. It surfaces as the `hf_golden_gate` bit-exact replay assert (`assert_eq!(fp1, fp2)`, the same prompt run twice on one executor) failing under `cargo test`'s default parallelism: the SEQ and LONG `#[test]`s each build their own executor, so cargo runs them as two co-resident streams on one GPU. Single GPU (x86_64, sm_89), 12 repeated runs: parallel fails 12/12, serial (`--test-threads=1`) 0/12. It is a timing race — on Dual-GH200 (aarch64, sm_90) the read always wins and the assert passes (an instrumented per-op trace there shows the race does not fire), while sm_89's single-GPU contention shifts the schedule.

## Reproduce (Single GPU, x86_64, sm_89)

```
OPENINFER_TEST_MODEL_PATH=<Qwen3.5-4B> \
  cargo test --release -p openinfer-qwen35-4b --features qwen35-4b --test hf_golden_gate

OPENINFER_TEST_MODEL_PATH=<Qwen3.5-4B> \
  cargo test --release -p openinfer-qwen35-4b --features qwen35-4b --test hf_golden_gate -- --test-threads=1
```

FeathBow:
## Description

Fixes #410 

`gated_delta_rule_decode_kernel` (`openinfer-kernels/csrc/qwen35/gated_delta_rule.cu`) reuses the `warp_norms` shared buffer for the q-norm and the k-norm, with no barrier between thread 0's read of the q partial-sums and the warps' overwrite with the k partial-sums — a write-after-read (WAR) hazard on the reused buffer. Under two co-resident CUDA streams the altered warp schedule lets the overwrite race ahead of the read → corrupted q-normalization → nondeterministic decode. Add one `__syncthreads()` to close the window.

The fix is the kernel barrier, not making the determinism test serial: the assert only fails under parallel execution, so `#[serial]` / a GPU mutex would hide the symptom and remove the only regression guard for this race. Keep `hf_golden_gate` parallel as the guard.

## Verification

Single GPU (x86_64, sm_89), real `assert_eq!`, no instrumentation: before the fix, parallel `cargo test --test hf_golden_gate` fails 12/12; after, 0/12 with the accuracy gate (logits vs HF golden) still green — the fix changes no single-stream value, only guarantees the already-correct q-norm under concurrency.

#328 (the Qwen3.5-4B single-GPU benchmark) runs one executor = one stream, so the race does not manifest there (it is latent UB that single-stream scheduling resolves correctly on the tested hardware); single-stream output is bit-identical. The added barrier is also perf-neutral, MEASURED: `vllm bench serve` (input-len 1024 / output-len 256, n=30, 3 reps each), Single GPU (x86_64, sm_89) — mean TPOT 11.27 ms (fixed: 11.21/11.29/11.31) vs 11.29 ms (unfixed: 11.27/11.30/11.31). The fixed-vs-unfixed delta (0.02 ms) is smaller than the within-config run-to-run noise (~0.1 ms).

## Type of Change

## Checklist

### DIFF
--- a/openinfer-kernels/csrc/qwen35/gated_delta_rule.cu
+++ b/openinfer-kernels/csrc/qwen35/gated_delta_rule.cu
@@ -74,6 +74,11 @@ __global__ void gated_delta_rule_decode_kernel(
         smem_norm[0] = rsqrtf(total + 1e-12f);
     }
 
+    // Reused buffer: fence thread 0's q-norm read of warp_norms above before the warps
+    // overwrite it with k_sq below. Cross-warp; without this, two co-resident streams race
+    // read-vs-overwrite into a corrupted q-norm (nondeterministic decode). Do not remove.
+    __syncthreads();
+
     float k_sq = (j_slice == 0) ? k_val * k_val : 0.0f;
     k_sq = warp_reduce_sum(k_sq);
     if (lane_id == 0) warp_norms[warp_id] = k_sq;


xiaguan:
Thanks.

Issue #236, opened 2026-06-04
kimi-k2: logprobs and echo support


xiaguan:
### Background

`logprobs` and `echo` are part of the completions API surface this server exposes, and eval harnesses (lm-eval-style loglikelihood tasks) depend on them. Qwen3-4B supports both.

### Current state

Kimi-K2 silently returns neither: every emitted token carries `logprob: None`, and prompt tokens are never echoed — a request with `logprobs=5` or `echo=true` gets a response missing those fields with no error. The engine already computes the top-1 value on-device, so the building blocks exist.

### Desired outcome

- `logprobs` requests return real per-token top-k logprobs; `echo` returns prompt tokens — or, until implemented, such requests are explicitly rejected rather than silently stripped.
- Loglikelihood-style evals become runnable against the Kimi endpoint.

### Acceptance boundary

- Verified through `/v1/completions` against the OpenAI response schema.
- Decode TPOT impact of logprob extraction is measured and acceptable (logprobs-off path unaffected).

xiaguan:
TP1/DP8 half landed in #269 (merged as 59fbbb5): the runner computes exact host log-softmax + top-K of the full-vocab logits row when `GenerateRequest::logprobs > 0` (one D2H per token, on-demand only); plumbed through prefill + decode_batch, the DP coordinator, and the TP8 scheduler. `/v1/completions?logprobs=N` works on Kimi via the existing frontend wiring. The accuracy gate now bounds a two-sided |Δlogprob| distribution on every pass.

Remaining:
- **TP8/DP1 logprobs** — each rank holds a vocab shard; needs a cross-rank logsumexp merge. Currently rejected loudly (`runner/executor/tp8_dp1.rs::ensure_no_logprobs_tp8`). Blocked on the #226 TP8 keep-or-demote decision — if TP8 is demoted, this half closes as wontfix.
- **echo / PromptTokens** — `TokenEvent::PromptTokens` is never emitted by the kimi runner; prompt logprobs need teacher-forced prefill logits, which the prefill path already materializes.

xiaguan:
Status after PR #292 (verified through /v1/completions on 8×H200, Kimi TP1/DP8/EP8):

**Done**
- `logprobs=3`: full OpenAI schema — `tokens`, non-null `token_logprobs`, `top_logprobs` with 3 entries per position. (Engine side landed in #269/#285.)
[...]

PR #417, opened 2026-06-17
refactor(sample): converge per-model token selection + logprobs into openinfer-sample

xiaguan:
## What

A new `openinfer-sample` crate becomes the single door from a logits arena to next-token ids and host logprobs, for the models that batch their sampling. It absorbs the two model-agnostic halves of sampling that were previously scattered and duplicated.

## Why

Token selection and host logprob math depend on nothing model-specific except vocab width, yet:

- selection lived in `openinfer-core::ops` (`select_batch_tokens_into`);
- the host logprob math was copy-pasted **three times** (qwen3 `executor.rs`, qwen35 `logprobs.rs`, kimi `runtime.rs`), each drifting on accumulation precision (f32 vs f64) and top-k tie-breaking (opposite orders), with checked vs unchecked indexing.

## What the crate owns

- **`select_batch(...)`** — one next-token id per arena row. Greedy and *effectively-greedy* (`top_p <= 1/vocab`) rows resolve together through one batched indexed argmax; the rest compact into one FlashInfer temperature/top-k/top-p pass. No per-row escape hatch, so a caller cannot regress to `for i { sample(i) }`.
- **`token_logprob_from_row<T: Copy + Into<f32>>`** — one generic host log-softmax + top-k. f64 exp accumulation (a 160k-wide bf16 vocab loses precision summing exps in f32), ascending-token-id tie-break, generic over the row element so qwen feeds `f32` and Kimi feeds `bf16` with **no widening copy**. This formats the logprob of an *already-selected* token — it is not selection.
- **`SampleScratch`** — consolidates six loose scratch buffers; rejects a vocab-width mismatch (**crash early** vs a silent kernel OOB).

## Who routes through it

- **qwen3, qwen35** — fully (selection + logprobs).
- **Kimi-K2** — non-greedy sampler + logprobs. Its **greedy** stays model-side: a *vocab-sharded* local argmax whose top-1 logit feeds a cross-rank DP reduction (#236/#237) that the whole-vocab crate can't express.
- **DeepSeek-V2-Lite / V4** — **do not** route through. Both are greedy-only and pick with a bare argmax: v2-lite on the GPU via `core::ops::argmax`, v4 on the **host** via `argmax_f32`. They need no batched sampling policy.

## core::ops: policy out, primitives stay

`openinfer-core::ops` drops only the sampling **policy** (`select_batch_tokens_into`, `sampling_params_effectively_greedy`). The bare argmax primitives (`argmax`, `argmax_batch_bf16_into`) **stay** in the facade — they are kernel ops, not policy, and DeepSeek-V2-Lite's single-rank greedy still selects through them.

## Layering

`openinfer-sample` sits beside `openinfer-core`, atop `openinfer-engine` (the CUDA-free `SamplingParams`/`TokenLogprob` contract) and `openinfer-kernels` (the CUDA build owner). It does **not** depend on `openinfer-core`.

## Net effect

27 files, +726/−442. Deletes `core::ops::sampling` (policy), three `compute_logprobs_from_cpu`/`host_token_logprob` copies, and qwen3's `SamplingScratch`. Also fixes leftover `bool_to_int_with_if` lints in kernels sampling.

## Verification

| Check | Result |
|---|---|
| qwen3 / qwen35 / kimi-k2 build | ✅ |
| deepseek-v2-lite `cargo check` (consumes `core::ops::argmax`) | ✅ |
| deepseek-v4 | ⚠️ **not built** — host lacks TileLang; its Rust uses an independent host `argmax_f32` and touches no changed symbol |
| `openinfer-sample` tests (3 CPU + 6 GPU model-free) | ✅ 9/9 |
| qwen3 `sampling_behavior` gate (real weights, end-to-end `select_batch`) | ✅ |
| general clippy hook + `-D warnings` on sample/kimi/kernels | ✅ exit 0 |
| `cargo fmt --check` | ✅ |

## Follow-up (non-blocking)

- DeepSeek-V4 still selects greedily on the **host** (`argmax_f32`) — the one CPU token-selection path left in the tree. Greedy-only, so it needs nothing `select_batch` offers today; if V4 grows temperature/top-p sampling, that host argmax is what should move here (needs an 8-GPU box to verify).
- qwen3 reallocates the whole `SampleScratch` when a batch exceeds its decode bucket (pre-existing behavior, not introduced here); qwen35's fixed-`MAX_BATCH` sizing is the cleaner pattern.

See `docs/subsystems/sampling/openinfer-sample.md` for the full design.


xiaguan:
@codex review

Issue #280, opened 2026-06-06
dsv2lite: diagnose why HTTP serving throughput does not scale with concurrency


CAICAIIs:
### Background

The retained status ledger says PegaInfer HTTP throughput did not scale with concurrency in the DeepSeek-V2-Lite snapshot. That is separate from the direct same-prompt diagnostic batch path and should be debugged as serving behavior.

### Current state

- Direct batch 1/4/8 exists for attribution and diagnostic benchmarking.
- The HTTP comparison uses request concurrency pressure, not a proven internal batch size.
- Broader serving parity claims are out of scope until the runtime path proves mixed live requests can share decode work.

### Desired outcome

Isolate the bottleneck that prevents DeepSeek-V2-Lite HTTP serving throughput from scaling under concurrent requests, then land or propose one narrow fix.

### Acceptance boundary

- A reproduction records HTTP benchmark commands, concurrency, prompt/output shape, completed/failed requests, TTFT, TPOT/ITL, and output hashes when available.
- The analysis separates frontend overhead, scheduler admission, decode batching, and NCCL/model execution.
- Any fix is scoped to the measured bottleneck.
- `--max-concurrency` is not described as true batch size unless the engine path proves it.

Related roadmap: #205

PR #421, opened 2026-06-18
perf(qwen35): fuse gate/up MLP and tune decode GEMMs

CAICAIIs:
## Summary

- Fuse Qwen3.5 MLP gate/up projections into one load-time `gate_up_proj`, so prefill/decode use one gate-up GEMM plus fused SiLU*mul.
- Tune Qwen3.5 decode cuBLASLt buckets before CUDA Graph capture.
- Propagate `silu_mul_fused` CUDA launch errors through the Rust wrappers.
- Refresh Qwen3.5 benchmark/docs with the current OpenInfer vs vLLM boundary.

## Benchmark

Same-host Qwen3.5-4B direct OpenInfer A/B:

| Workload | Before | After | Delta |
| --- | ---: | ---: | ---: |
| 1/256 steady TPOT | 6.524 ms | 6.386 ms | -2.1% |
| 1/512 steady TPOT | 6.603 ms | 6.397 ms | -3.1% |
| 1024/256 steady TPOT | 7.338 ms | 7.100 ms | -3.2% |
| 2048/1 TTFT | 97.978 ms | 95.855 ms | -2.2% |

The vLLM comparison is still workload-scoped: OpenInfer is close on prompt-len-1 decode, while vLLM still leads 1024/256 HTTP decode and high-concurrency throughput.

## Validation

- `cargo fmt --all --check`
- `git diff --check`
- `cargo test --release -p openinfer-kernels vstacked_gate_up_gemm_matches_split_mlp_projection -- --nocapture`
- `cargo check --release -p openinfer-core -p openinfer-kernels -p openinfer-qwen35-4b --features qwen35-4b`
- `cargo check --release -p openinfer-deepseek-v2-lite`


xiaguan (approved):
LGTM — clean, honest, well-scoped increment.

What I like:
- The 2-3% direct TPOT win is measured via a same-binary A/B, not just an HTTP delta, so the attribution is real.
- `vstacked_gate_up_gemm_matches_split_mlp_projection` locks the gate/up fusion to the split-projection reference, so the load-time stacking can't silently drift numerically.
- Propagating the `silu_mul_fused` CUDA launch error through the Rust wrappers is the right call — better to surface a bad launch than swallow it.

Follow-up request: please open a tracking issue for Qwen3.5 follow-up performance work so this doesn't stall here. The most actionable finding in the refreshed benchmark is the attribution, not another kernel rewrite:

- At 1024/256, concurrency 16: openinfer **in-process 9.320 ms** TPOT vs **HTTP 15.566 ms**, while vLLM HTTP is **9.823 ms**. So our model-kernel time already roughly matches vLLM's end-to-end HTTP; the remaining ~6 ms/token gap is serving/scheduler/event-sync overhead, not kernels.

The issue should track:
1. Add a measured-only server-side range and split the 1024/256 c16 gap across scheduler wait / event sync / request dispatch / model execution (roadmap item #6).
2. Teach the Qwen3.5 direct decode bench to prove cached-token exclusion before reporting pure decode TPOT.
3. Cross-link the cross-model serving-overhead issues — this is the same root cause surfacing on Qwen3.5: #160 (frontend hot path), #225 (kimi-k2 HTTP-vs-in-process), #280 (dsv2lite concurrency scaling).

Merging the kernel work here is fine; the serving-layer gap belongs in that follow-up issue.

PR #423, opened 2026-06-18
docs(qwen35): add prefix cache design document

### LINKED ISSUE #257: qwen35: prefix caching needs a design for recurrent state (discussion)
### Background

Prefix caching landed for Qwen3-4B (#216) with large warm-TTFT wins. Qwen3.5's hybrid stack breaks the assumption underneath it: 24 of 32 layers are linear attention carrying fixed-size recurrent state, not paged KV. A "prefix hit" must restore both the full-attention KV blocks *and* a recurrent-state snapshot consistent with the matched boundary.

### Current state

- No prefix reuse exists for this crate.
- A recurrent-state snapshot is tens of MB per boundary (vs KB-scale for a KV block), so snapshot granularity (every block? every N blocks? request end only?) dominates the design — naive per-block snapshotting would blow the memory budget.

### Desired outcome

- A design note in `docs/models/qwen35/` settling: snapshot granularity, eviction interaction, and which match lengths are even worth caching given the snapshot cost — before any implementation.

### Acceptance boundary

- This issue covers the design discussion plus a committed note; implementation is a separate follow-on.
- Depends on the paged-prefill migration (the current monolithic prefill has no block boundaries to snapshot at).

Ke-Wng:
## Description

Fixes #257

Add design document for Qwen3.5-4B prefix caching. Qwen3.5's hybrid architecture (24 linear-attention + 8 full-attention layers) breaks Qwen3's KV-only caching assumption—prefix hits must restore both paged KV blocks and recurrent-state snapshots at the same boundary.

**Key design decisions:**
- **Snapshot interval:** 256 tokens (4×GDR chunks), chosen to balance cold prefill overhead (+32ms for 4096-token prompt) vs hit-rate granularity
- **Two-tier pool:** GPU primary (~29 slots on RTX 4090 = 7,424 cached tokens) + CPU eviction backup (unlimited, +2.18ms H2D restore cost)
- **Joint matching:** radix lookup requires both KV and snapshot availability at the same boundary
- **LRU eviction:** independent per-tier, with active-request protection

See `docs/models/qwen35/prefix-cache.md` for full design rationale, quantitative analysis, pitfalls, and implementation roadmap.

## Type of Change

- [x] Documentation update

## Checklist


xiaguan:
Thanks for writing this up, but I don’t think this is ready to land as a committed design doc yet.

The hard part here is not mainly adding snapshot offload. For Qwen3.5, prefix caching needs a real allocation / indexing / lifetime design for linear-attention state. A valid hit has to jointly restore full-attention KV, GDR recurrent state, and conv state at the same token boundary, with the same token hash / adapter salt. That means we need to define snapshot handles, ownership, pinning, eviction, and how radix lookup joins KV blocks with recurrent snapshots.

There is also an implementation gap: the current prefill path does not naturally produce a whole-model snapshot at every 256-token boundary. The GDR chunk scratch is per linear layer, and conv state is only kept as final state, not as per-boundary snapshots. So “D2H copy at GDR boundaries” is not enough as an implementation plan.

I’d prefer to move this to an RFC issue first and narrow the first step to:
- GPU-only recurrent snapshot allocator
- exact snapshot contents: GDR state + conv state
- snapshot key and boundary semantics
- joint KV+snapshot lookup
- request lifetime / eviction protection
- how prefill will actually materialize boundary snapshots

CPU-tier offload and two-tier LRU can be a later optimization once the core indexing/lifetime model is clear.

Ke-Wng:
Thanks for the review. I agree with narrowing the scope first. I drafted an RFC comment under https://github.com/openinfer-project/openinfer/issues/257 that focuses on the first step: a GPU-only recurrent snapshot design for Qwen3.5 prefix caching.

CAICAIIs:
Hi, thanks for your contribution! Could you let me know if you're still working on this PR? If yes, please resolve the merge conflicts.

Ke-Wng:
> Hi, thanks for your contribution! Could you let me know if you're still working on this PR? If yes, please resolve the merge conflicts.

Yes, still on it. Based on our discussion, I'll make the necessary changes and push the updated doc in a few days. Appreciate the reminder!

CAICAIIs:
lgtm, please fix the conflict and this is ready to land from my side

CAICAIIs (approved):

PR #430, opened 2026-06-21
feat(qwen3): add profile-based KV cache sizing

### LINKED ISSUE #422: RFC: profile-based KV-cache sizing to replace the flat 0.85 activation headroom
## Summary

Today the qwen3 KV pool is sized by a flat heuristic: `kv_budget()` in `openinfer-qwen3-4b/src/weights.rs:846` takes **85% of the free VRAM after weights are resident** and leaves a hardcoded **15%** to absorb everything that is neither weights nor KV — activation scratch, CUDA-graph captured buffers, cuBLAS/FlashInfer workspace, sampling buffers, etc.

```rust
let (free_bytes, _) = mem_get_info();      // free after weights load
let kv_budget_bytes = free_bytes * 0.85;   // KV takes 85%, 15% is the catch-all margin
let num_blocks = kv_budget_bytes / bytes_per_block;
```

That 15% is **guessed, not measured**. We can't tell whether it's wastefully generous or one kernel/dtype/batch-config change away from OOM, and it doesn't adapt across models.

This is the residual half of #358. The live OOM there is already gone — per-request prefill chunking (#363) bounds per-step activation, so a long prompt no longer spikes the scratch. What remains is that our headroom is unprincipled. The goal of this RFC is to make "a well-formed request OOMs after admission" *structurally* impossible (headroom derived from a measurement) rather than merely *empirically unobserved* (headroom is a magic number that happens to hold).

This is a hardening RFC, not an urgent bug. Filing it so we converge on a design before anyone writes code.

## How vLLM does it (reference, not a prescription)

vLLM sizes the KV pool from a **measured** peak, not a formula:

- `determine_available_memory()` — `vllm/v1/worker/gpu_worker.py:372` — docstring: *"Profiles the peak memory usage of the model to determine how much memory can be used for KV cache without OOMs."*
- It runs one dummy forward `profile_run()` (`vllm/v1/worker/gpu_model_runner.py:6223` → `_dummy_run(self.max_num_tokens, is_profile=True)`) at the **worst-case step size**, wrapped in a memory-profiling context.
- It reads the allocator high-water mark `allocated_bytes.all.peak` and stores it as `self.peak_activation_memory` (`gpu_worker.py:446`), then:

```python
# gpu_worker.py:431 / 461
non_kv_cache_memory     = non_torch_increase + torch_peak_increase + weights_memory
available_kv_cache_bytes = requested_memory            # total * gpu_memory_utilization (utils.py:410)
                          - non_kv_cache_memory         # weights + measured peak activation + non-torch
                          - cudagraph_memory_estimate
```

Two properties that matter for us:

1. **Measured, not computed.** Because it runs the real kernels, the peak captures cuBLAS/attention workspace, CUDA-graph buffers, and fragmentation — exactly the things a hand-written formula forgets. Analytically predicting activation per request is not tractable, and vLLM does not attempt it.
2. **Bound, then measure the bound, then reserve it.** The runtime peak can't exceed the profiled peak because the scheduler caps every step at `max_num_batched_tokens` (chunked prefill). We already have the "bound" piece (`max_prefill_tokens`, #363); we're missing the "measure + reserve" piece.

vLLM also errors out early if free VRAM is below `total * gpu_memory_utilization` (`request_memory`, `utils.py:405`) rather than discovering it mid-serving.

## What we need

Replace the flat `free * 0.85` with a startup measurement:

- Profile a worst-case step once at startup, read the real peak, and size `num_blocks` from `budget - weights - measured_peak - margin`.
- Expose a utilization knob (our equivalent of `gpu_memory_utilization`) instead of the hardcoded `0.85`.
- Keep **admission a pure KV-block check**. This RFC is explicitly *not* about per-request activation accounting at admission — that's not analytically tractable and isn't where the fix belongs.

## Open design questions — please reply with your approach before implementing

I don't want to assume the mechanism; I'd like the design discussed first. Things to resolve:

1. **Peak measurement in our stack.** Do we have an allocator high-water mark we can read, or do we add one / diff `cuMemGetInfo` around a dummy run? How does that interact with cudarc's default-stream allocations vs. our KV pool (which we'd want to exclude or run the profiling *before* reserving)?
2. **What to profile.** Our worst case is a *unified* step — a prefill chunk up to `max_prefill_tokens` **plus** a decode batch up to `max_decode_batch_size` — not prefill alone. And the decode path is CUDA-graph captured with pre-allocated buffers: does graph capture need to happen before/within the profiling window so its memory is counted?
3. **Scope.** qwen3-only first, or land in the shared runtime so deepseek-v2-lite / kimi-k2 inherit it? TP/EP and multi-GPU change the per-rank picture.
4. **Failure mode & knob.** Utilization default, and what happens when the measured peak leaves too little for a viable KV pool — error out early like vLLM's `request_memory`, or fall back?
5. **Margin.** Peak-allocated != peak-reservable under fragmentation; do we keep a small safety margin on top of the measured peak, and how big?

## Non-goals

- Not changing admission to compute activation per request.
- Not touching the echo all-position-logits path (see #416 discussion — it's unreachable from the API).

Refs: #358 (original report), #363 (chunked prefill that bounded per-step activation), #416 (related, echo path). vLLM source: `vllm/v1/worker/gpu_worker.py`, `gpu_model_runner.py`.

wdlin233:
Add Qwen3 memory options for GPU utilization and KV cache margin, then size the KV pool from a startup profile instead of the old free-memory heuristic.

The profile runs after weights are loaded and before the final KV pool is allocated. It exercises decode buffers, sampling scratch, CUDA graph decode setup, and a bounded unified step with max prefill tokens plus the largest decode batch. Peak sampling now happens while temporary forward buffers are still live, and temporary profile KV memory is excluded from the final budget.

Also plumb the new options through the server CLI and use a checked MiB-to-bytes conversion for the KV cache margin.

## Description

Fixes #422 

Profile-based KV cache sizing replaces the old `free * 0.85` heuristic. Key pieces:

- `Qwen3MemoryOptions` (gpu_memory_utilization=0.90, margin=150 MiB)
  plumbed through server CLI as `--gpu-memory-utilization` and
  `--kv-cache-memory-margin-mib`

- `profile_kv_budget_on_worker`: after weights load, allocates decode
  buffers + sampling scratch + temp KV, runs a unified step at worst-case
  prefill + decode batch, records peak via `cuMemGetInfo` diff, subtracts
  temp KV, adds margin, and sizes the real KV pool from the remainder

- TP: each rank profiles independently, executor takes min block count

- Old `kv_budget()` preserved as fallback for users who want the
  free-heuristic path

Validated on RTX 3090 Ti (24 GB), single-GPU and TP=2.

Addressed the remaining Codex feedback in ef24fb5. The LoRA adapter-load profiling case is intentionally left as a follow-up because LoRA disables CUDA graph in openinfer.

  ## Validation

  - `git diff --check`
  - `cargo test --release -p openinfer-qwen3-4b --lib`：42 passed
  - `cargo test --release -p openinfer-server --lib`：built, 0 tests
  - `cargo test --release --workspace --lib -- --skip gdr_copy_flag_GPU`：passed

  Unfiltered `cargo test --release --workspace --lib` currently fails one unrelated hardware-runtime test on this host:

  - `openinfer-comm-cuda-lib::test_gdr::gdr_copy_flag_GPU`
  - reason: `/dev/gdrdrv` is absent, so `gdr_open()` fails even though user-space `libgdrapi` is available

## Type of Change

- [ ] Documentation update

## Checklist


wdlin233:
@codex review

wdlin233:
@codex review

wdlin233:
@codex merge — LoRA disables CUDA graph in openinfer, so the profiled graph-mode peak is already conservative for the LoRA no-graph codepath. Extending the profile to cover LoRA adapter load is worth a follow-up but out of scope for this pass (echo fix is in ef24fb51).

xiaguan (approved):
LGTM. Verified locally on an RTX 5070 Ti (16 GB) with Qwen3-4B:

- `cargo build --release` (lib + server) and clippy clean; `cargo test --release -p openinfer-qwen3-4b --lib` → 44 passed, including the two new echo-rejection tests.
- Real startup profiling sizes the pool correctly: `initial_used=9651 / peak_non_kv_increase=338 / margin=150 → KV=4114 MiB`; nvidia-smi reports 14044 MiB used ≤ 14253 MiB (90% request). No OOM.
- 8-way concurrent burst (128 tokens each): 8/8 succeed, memory stable, no OOM.
- Early-out path works: `--gpu-memory-utilization 0.62` exits cleanly with `leaves no room for KV cache: requested=9819, non_kv=10139` instead of OOMing mid-serve — exactly the vLLM `request_memory` behavior #422 asked for.

Worst-case profile shape is sound: admission caps `active + prefilling ≤ max_decode_batch_size`, so '1 max-prefill row + remaining decode rows' simultaneously maximizes total tokens and output rows. TP profiling is sequential but safe because `attach_tp_comm` runs after profiling, so the per-layer NCCL all-reduce is a no-op during the profile (and being in-place, it doesn't change buffer sizes).

Nit (non-blocking, optional follow-up): `Qwen3MemoryOptions::validate()` runs three times along the call path (main.rs, from_runtime_with_lora_options, profiled_kv_budget) — one boundary is enough.

PR #431, opened 2026-06-21
feat(qwen35): scheduler-level chunked prefill (#375)

### LINKED ISSUE #375: Implement scheduler-level chunked prefill for Qwen3.5
Here is an follow up for PR#368

Reference pr: [feat(qwen3): chunked prefill + drop bs8/16 buckets (cuBLAS split-K hole) by xiaguan · Pull Request #363 · openinfer-project/openinfer · GitHub](https://github.com/openinfer-project/openinfer/pull/363) 

Current Qwen3.5 chunked-prefill status

1. **KV-cache level** chunking already exists in [[openinfer-qwen35-4b/src/prefill.rs:66-76](vscode-webview://1junfb8ee29m8lp8hpd95pslfhc5taofla5o68p2t3slecd36rsi/openinfer-qwen35-4b/src/prefill.rs#L66-L76)](vscode-webview://1junfb8ee29m8lp8hpd95pslfhc5taofla5o68p2t3slecd36rsi/openinfer-qwen35-4b/src/prefill.rs#L66-L76): a prompt is split into `PREFILL_CHUNK_LEN`-token chunks and appended to the KV cache **sequentially**, bounding the per-pass GDR scratch to avoid OOM.
2. **But at the scheduler level**, all of a prompt's chunks are still prefilled **within a single step** — so Qwen3.5 hits the same mixed-load ITL problem #368 measured on Qwen3.

Scope
1. Let the **scheduler** drive chunked prefill — one chunk per step, interleaved with decode (port the #363 approach to the Qwen3.5 scheduler, which has no prefill budget yet).
2. **Perf report**: chunked prefill vs. not (mixed-load ITL, same profile as #368).

scatyf3:
## Description

Fixes #375

Qwen3.5 already chunked prefill at the **KV level** (`PREFILL_CHUNK_LEN`), but the whole chunk loop ran inside a single scheduler step. A long prompt was therefore atomic to the scheduler: one `Unified` step packed the entire prompt plus every active decode row into one forward pass, stalling all decoding requests for one inter-token gap — the same ITL tail #368 profiled on Qwen3.

This PR moves chunking up to the scheduler. State now lives across steps instead of inside one `prefill` call, and each step prefills at most a **per-step token budget** off the FIFO front, servicing the decode batch between chunks.

No new kernel: `prefill_forward` already reads its base from `kv.seq_len()` and advances KV + recurrent state in place, so calling it with successive prompt slices evolves state bit-identically to a whole-prompt call. This is pure scheduler bookkeeping.

### KV / recurrent-state menagement

Before, `prefill` allocated the full KV/rec state in a temporary, and on completion did a KV-move + rec-copy and dropped the temp. Now state advances **in place** across chunks:
- **In flight:** a new `PrefillingRequest35` owns the growing `KvState` + `RecurrentState` + a `cursor` (prompt tokens prefilled so far) and lives in the `prefilling` FIFO across steps — not as a function local.
- **On completion:** the same ownership handoff as before — KV/rec state moves from `PrefillingRequest35` to `ActiveRequest35`.

### Scheduler (`scheduler.rs`, `scheduler/plan.rs`)

- **`prefilling` queue** — FIFO that owns in-flight KV/rec state across steps.
- **Per-step budget helpers** (pure, unit-tested):
  - `plan_prefill_chunks(remaining, budget)` packs front requests up to `budget`tokens; a request that doesn't fit takes a partial chunk and stays at thefront, so one long prompt is sliced across steps while short prompts behind it still get serviced. 
  - Admission shrinks the budgets passed to `admit_pending_requests` under 2 constrain, kv pages and cuda graph slots
- **`run_step`** mirrors the Qwen3 scheduler: 
- admitted prompts enter the`prefilling` queue → `ExecutionPlan` now threads the prefilling queue → advance scheduled prefill chunks → `decode_step` the active batch → `promote_or_requeue` each scheduled chunk (a finished prompt promotes to decode, an unfinished one re-queues at the front). 

## usage

use `OPENINFER_QWEN35_PREFILL_BUDGET`  to control chunked size

## Evaluation

- **Latency:** improves p99 ITL; slightly raises p50 in a minority of cases (the expected budget trade-off).
- **Correctness:** Qwen3.5 e2e output accuracy unchanged. `cargo fmt`, the 24 scheduler unit tests, and `e2e_scheduler` (at the 1024 default and at budget 4 for heavy slicing / concurrent interleave) all green.

## Type of Change

New feature (non-breaking change which adds functionality)


xiaguan (changes requested):
Verified locally (RTX 5070 Ti, Qwen3.5-4B): 30 lib tests + `e2e_scheduler` green at the 1024 default and at budget 4. Greedy output at budget 4 (273-tok prompt sliced into ~68 chunks) is byte-identical to the whole-prompt run, so the cross-step KV/recurrent advance is exact even with the chunkwise GDN kernel. Correctness LGTM.

Two changes before merge:

1. Drop the `OPENINFER_QWEN35_PREFILL_BUDGET` env var. Wire the existing `--max-prefill-tokens` server flag through to Qwen3.5 instead — it's already an arg (used by Qwen3) but currently ignored on this path. A hidden env var for a serving knob isn't the pattern we want, and it contradicts the 'mirrors Qwen3' framing.

2. #375 asks for an ITL A/B report and the PR doesn't include one. Please add the numbers: mixed-load p50/p99 ITL, chunked vs not, same profile as #368. The ITL win is structurally expected, but it should be measured, not asserted.

xiaguan (approved):
LGTM. Clean scheduler-level chunked prefill that mirrors the proven Qwen3 design — the admission accounting (`slot_budget = max_batch - prefilling.len() - active.len()`, plus `prefilling_future_pages` reserving each in-flight prefill's KV growth) keeps `active.len() + prefilling.len() <= max_batch` invariant, so promotion never runs out of a graph slot, and the state-ownership handoff across steps is symmetric. No new kernel, recurrent/KV state advances in place off `kv.seq_len()`.

One non-blocking nit: the `--max-prefill-tokens` help text in `openinfer-server/src/config.rs` still scopes the flag to Qwen3 ("one **Qwen3** scheduler step", "Echo requests are never split"), but after this PR it also drives the Qwen3.5 chunked-prefill budget. Worth a one-line tweak so the flag's scope reads correctly for both model lines.

Approving.

PR #439, opened 2026-06-23
feat(qwen3-4b-dflash): add Qwen3-4B-DFlash draft model crate

kitty-eu-org:
## Description

Add `openinfer-qwen3-4b-dflash` crate supporting the `z-lab/Qwen3-4B-DFlash-b16` draft model, laying the groundwork for future speculative decoding integration. The crate is behind its own cargo feature and does not affect existing model crates.

**What's included:**

- **New model crate** (`openinfer-qwen3-4b-dflash`): config parsing, exact-key safetensor weight loading, draft forward with HF remote-code parity gate, unified `DFlashDraftCache` (mirrors reference `DynamicCache` lifecycle), draft-only batch executor/scheduler with FCFS exact-shape batching, single-instance batch buffers, bounded `max_caches` eviction with `drop_cache`, and join-on-drop scheduler shutdown aligned with `EngineHandle` / qwen3 `drop_request`.
- **Shared kernels** (`openinfer-kernels`): non-causal dense/ragged-batch prefill attention, K-only norm+RoPE variant, fused `strided_segment_copy`.
- **Correctness gates** (`hf_golden_gate.rs`, 8 tests): HF golden forward parity (uncached/unified-cache/draft-cache), batch-vs-single, executor/scheduler smoke, cache control (fail-closed / `drop_cache` / `max_caches`).
- **Docs & tools**: `docs/models/qwen3/dflash.md`, Python accuracy scripts (golden dump, forward bench, drafter-substitution comparison).

**Performance** (RTX 5070 Ti, CUDA 13.3, `ctx_len=2`, `q_len=16`): bs32 reaches 75,148 draft tok/s (9.0× scaling from bs1), with two perf passes — fused strided copy (1.5× at bs32) and K-only norm+RoPE for context-K.

**Out of scope**: target hidden extraction, target verification, acceptance, fallback token selection, OpenAI serving — these belong to the future speculative-decoding controller.

## Type of Change

## Checklist


xiaguan:
Thanks for this — genuinely nice work. The draft-forward perf passes (fused strided copy, K-only norm+RoPE for context-K, ~9× bs1→bs32 scaling) and the golden-gate parity suite are solid, and the crate is clean and well-scoped.

One timing note so we don't waste your effort: the **end-to-end DFlash speculative decoding integration already landed on `main`** in #436 + #442. It lives inside the existing `openinfer-qwen3-4b` crate (`src/dflash*`, `src/verify_graph.rs`) and covers the full loop — draft model loading + batched draft forward + target "verify" forward (piecewise CUDA graph) + acceptance + KV rollback + OpenAI serving — behind `--dflash-draft-model-path`, with a greedy-losslessness gate and 5070 Ti / 5090 validation (it matches or beats vLLM dflash across c1/c8/c16, and gsm8k strict-match is identical spec-on vs spec-off). See `docs/models/qwen3/dflash-speculative-decoding.md`.

Because of that, the draft half this PR builds — config, weight loading, batched draft forward, draft cache, draft executor/scheduler — substantially overlaps what's already shipped, and the "future speculative-decoding controller" it's groundwork for already exists. So we don't think it makes sense to land this as a separate parallel draft crate.

That said, we'd really like to keep your involvement, and to be fair some of your kernel work is in areas the merged path doesn't cover — e.g. the dedicated non-causal **dense/ragged-batch prefill attention** module, and the **K-only** norm+RoPE variant (we currently only have a combined QK variant). If those measurably beat what's on `main`, they're worth landing on their own as kernel PRs. Concretely, here's open DFlash work that builds directly on what you've already done:

**1. Draft-side piecewise CUDA graph** *(perf — your wheelhouse).* Our tracked next step. The 5-layer draft forward is the remaining ~16% of the decode launch gap; it needs the same pre/attn/post split as the verify graph, with the variable-length contiguous `DFlashLayerCache` handled at the eager-attention boundary. Your non-causal prefill / K-only norm+RoPE / fused strided-copy kernels are likely reusable here.

**2. Speculative *sampling* — MTP for non-greedy requests** *(new capability, highest value).* Today speculation is greedy-only: a request with `temperature`/`top_k`/`top_p` sampling correctly falls back to plain decode (no wrong distribution, but no speedup — and one sampling request in a batch currently disables speculation for its co-batched greedy requests). Adding lossless **speculative sampling** — rejection-sampling acceptance (accept with `min(1, p_target/q_draft)`, resample from `(p_target − q_draft)₊` on reject) so the committed distribution exactly matches the target's — would unlock speculative speedup for sampled requests too. This lives in the verify controller (`src/dflash.rs` + `src/verify_graph.rs`), so it builds right on the merged path.

**3. Smaller, well-defined.** Tighten the DFlash memory reservation (`from_config`, from the conservative per-request upper bound to the lane-level `DFlashBatchScratch` footprint); and make `gemm_lt_pin_tune` a real warmup (actually run a `cublasLtMatmul` so the first matmul doesn't land inside graph capture).

Happy to scope any of these with you — **#2** in particular is a meaty, genuinely-missing piece. Thanks again for digging into DFlash, and sorry the timing overlapped.

PR #438, opened 2026-06-22
fix(qwen3): opt-in split-KV chunk-count reduction-order pin

FeathBow:
## Description

Refs #414, #435 

Follow-up to #428. A request's split-KV decode output depends on which other requests share its batch: the split chunk size is derived from the live batch `max_seq_len`, so co-batching a request with a longer one changes that request's chunk count and therefore the online-softmax rescale order over its split-KV partials. Under the opt-in `NumericPolicy` (default `Tuned`, unchanged from today), `split_chunk_size` keys the chunk *size* on `max_context_tokens`, a process constant, so a request's chunk count, and therefore its rescale order, depends only on its own sequence length.

It covers one measured source from #435 (the split-KV chunk count). It does not address the other sources, such as the batch-dependent choice of decode attention kernel (NonPartition vs SplitKv) or decode/unified routing.

## Behavior

bf16 split-KV decode accumulates per-chunk partials and merges them with an online-softmax rescale whose order is the chunk count. Keyed on the live batch `max_seq_len`, that order can perturb the bf16 result enough to flip a near-tie argmax. Keying the chunk size on the model context limit makes the count request-local; for Qwen3-4B it resolves to `ceil(max_position_embeddings / 64) = 640` tokens, the finest split the 64-chunk workspace allows. The default path (`--batch-invariant` off, `Tuned`) is unchanged.

## Test Env

A new gate, `batch_invariance_decode_splitkv_graph`, prefills request A alone so its KV is fixed, then co-batches it with a shorter then a longer B so that only the decode `max_seq_len` moves: A's Tuned chunk count goes 64 -> 40 across the SplitKv graph's capture and replay. A runtime counter (`split_kv_counters`) lets the gate assert the chunk size directly: it drifts under Tuned (79 -> 126) and is constant under Pin/PerToken (640); A's top-K follows (Tuned drifts, the baseline-drift guard against a vacuous pass; Pin and PerToken bit-identical). Existing no-regression gates also pass. Verified on Single GPU (x86_64, sm_89) and GH200 (aarch64, sm_90).

```bash
OPENINFER_TEST_MODEL_PATH=<Qwen3-4B-base> cargo test --release -p openinfer-qwen3-4b --test batch_invariance_decode_splitkv_graph -- --nocapture
```

```
[...]

Issue #435, opened 2026-06-21
Tracking: batch-invariant Qwen3 inference: remaining sources and e2e bar


FeathBow:
## Description

Tracking issue for the Qwen3 batch-invariance workstream. Origin: #414.

#428 is PR1. It adds the default-off `--batch-invariant` path and covers one source: projection-GEMM N-dependence / reduction order in the tested Qwen3 paths. The rest stays here so #428 can stay scoped.

## PR1 coverage

  - Projection-GEMM reduction order across the tested prefill/decode/unified projection paths.
  - Mechanism: one cuBLASLt algo pinned per `{M,K}` at a fixed representative N, with per-token fallback.
  - Gates: within-path bit equality, `fallback==0` across the swept production N envelope, and baseline-drift guards.
  - Measurements: within-path bit-identical on sm_89 and GH200/sm_90; decode overhead is near-free in the measured envelope.

## Remaining sources

  - **Prefill-side e2e invariance**: #428 pins projection GEMMs in the tested prefill path; the rest of prefill still needs output-level coverage, including chunked prefill spanning unified steps.
  - **Split-KV chunk count**: decode attention at long context / batch <= 32 changes online-softmax rescale order when chunk count depends on live batch shape.
  - **Attention-path selection**: NonPartition vs SplitKv is selected from batch composition, so a request can run a different attention kernel depending on co-scheduled requests.
  - **Decode vs unified forward path**: pure decode and unified mixed prefill+decode use different execution paths; this residual survives the pin and is tracked by the ignored cross-path test.
  - **Sampling, norm, embedding, attention, and TP all-reduce**: audit as needed for batch-composition dependence.
  - **Decode-overlap compatibility**: `--batch-invariant` is currently rejected with `--decode-overlap` because stream override forces the pin path to fall back to per-token.
  - **Cross-cuBLAS / cross-driver determinism**: the pinned algo is selected at runtime per `{M,K}`, so cross-version bit identity is not guaranteed.
  - **Uniform coverage**: the policy lives in the shared GEMM seam, but some paths such as `typed_ops` bypass it and other models lack accuracy gates.

## e2e bar

  The target gate runs under production CUDA graph and deterministic sampling settings:

  - submit a request alone;
  - submit the same request co-batched with varying load: different lengths, mixed prefill/decode;
  - assert identical output token sequences.

  Kernel-level gates are evidence for individual sources; this issue closes on the output-level gate.

xiaguan:
[...]

PR #449, opened 2026-06-24
perf(qwen3): decouple batch-invariant split-KV size from Tuned cap

FeathBow:
## Description

Refs #414, #435. Follow-up to #438.

#438's opt-in `--batch-invariant` (`NumericPolicy::Pin`) keys the split-KV decode chunk size on `max_context_tokens`, so a request's chunk count — hence its online-softmax rescale order — depends only on its own length. Under the 64-chunk bound that resolved to a fixed 640-token split for Qwen3-4B, a very coarse one: a request gets `ceil(seq / 640)` chunks, so at short and mid contexts it launches too few CTAs to fill the GPU, costing decode TPOT.

This splits the single `SPLIT_KV_MAX_CHUNKS_PER_REQUEST` constant into two:

- `SPLIT_KV_TUNED_MAX_CHUNKS = 64` — the default (`Tuned`) path's chunk-count cap and padded-grid width, unchanged from #438.
- `SPLIT_KV_MAX_CHUNKS_PER_REQUEST = 256` — the `Pin`/`PerToken` padded-grid width and the basis for their fixed split size: `ceil(max_context_tokens / 256) = 160` tokens (was 640). `ceil(own_seq / 160)` is still request-local (batch-invariant by construction) but a finer split that better fills the GPU at low batch.

The padded grid is policy-dependent (`Tuned` 64, `Pin`/`PerToken` 256), so the default path's numerics and runtime grid are unchanged from #438; only `--batch-invariant` pays the wider runtime grid. The ~49 MiB larger split-KV workspace is allocated for the decode buffers regardless of policy, but it is a one-time memory cost; the runtime tradeoff is the wider `Pin`/`PerToken` grid measured below.

It addresses the split-KV chunk-count source from #435. `--batch-invariant` is still required — it also pins the decode GEMM-N reduction order, an orthogonal axis the chunk size does not touch — and this does not address decode attention-path selection or decode/unified routing.

## Decode perf

`vllm bench serve`, random dataset, output-len 128, `--temperature 0 --ignore-eos`, median TPOT (ms). Each `(+x%)` is the split-KV decode tax over the same-ctx `Tuned` column (`col / Tuned − 1`). bs=1, the low-concurrency regime `--batch-invariant` targets:

**Single GPU (x86_64, sm_89):**

| ctx | Tuned | #438 (640) | this PR (160) |
|---|---|---|---|
| 1024 | 10.12 | 11.52 (+13.8%) | 10.97 (+8.4%) |
| 2048 | 10.42 | 11.88 (+14.0%) | 11.14 (+6.9%) |
| 4096 | 10.71 | 12.29 (+14.8%) | 11.39 (+6.3%) |
| 8192 | 11.58 | 12.53 (+8.2%) | 12.18 (+5.2%) |
| 16384 | 13.06 | 13.54 (+3.7%) | 13.64 (+4.4%) |

**GH200 (aarch64, sm_90):**

| ctx | Tuned | #438 (640) | this PR (160) |
|---|---|---|---|
| 1024 | 4.33 | 5.12 (+18.2%) | 4.25 (−1.8%) |
| 2048 | 4.36 | 5.16 (+18.3%) | 4.44 (+1.8%) |
| 4096 | 4.44 | 5.55 (+25.0%) | 4.59 (+3.4%) |
| 8192 | 4.69 | 6.06 (+29.2%) | 4.91 (+4.7%) |
| 16384 | 5.13 | 6.13 (+19.5%) | 5.19 (+1.2%) |

At bs=1 the finer split removes most of #438's tax — near Tuned on GH200 (−1.8% to +4.7%), and on sm_89 down to a ~+6% residual consistent with the orthogonal GEMM-N pin cost from #438.

<details><summary>bs=8 and the high-concurrency tradeoff</summary>

At bs≥8 a residual `--batch-invariant` tax remains (GEMM-N pin). The finer split is **not** uniformly better there: on GH200 it still helps, on sm_89 it mildly regresses because the finer split over-subscribes CTAs where the batch already fills the SMs.

bs=8, median TPOT (ms):

| | ctx | Tuned | #438 (640) | this PR (160) |
|---|---|---|---|---|
| sm_89 | 1024 | 14.31 | 17.46 | 16.71 |
| sm_89 | 2048 | 18.76 | 22.90 | 23.00 |
| sm_89 | 4096 | 28.78 | 35.83 | 38.50 (+7.5% vs #438) |
| sm_90 | 1024 | 4.86 | 6.07 | 5.51 |
| sm_90 | 2048 | 5.86 | 8.04 | 6.25 |
| sm_90 | 4096 | 6.71 | 10.46 | 9.93 |

</details>

## Hardware dependence

Measured only on sm_89 and sm_90. The optimal fixed split size is hardware-dependent: a 64-token split is faster still there across all batch sizes — whereas sm_89 regresses with it at bs≥8 / long context. 160 is the cross-architecture value: major bs=1 recovery on both (near/full on GH200, with a residual sm_89 tax); in the measured matrix, the visible regression is the sm_89 bs=8 long-context case. It auto-scales per model (`max_context_tokens / 256`), not per-GPU tuning. If conditions allow, please compare on other architectures.

## Test

`batch_invariance_decode_splitkv_graph` (gate unchanged from #438): co-batching request A with a longer B drifts A's `Tuned` chunk count (top-K differs) while `Pin`/`PerToken` replay bit-identically across the SplitKv CUDA-graph. Verified on Single GPU (x86_64, sm_89): clippy clean; `test result: ok`:

```
RESULT Tuned_chunk_tokens(C=79,R=126) | decode_topk_eq baseline=false pin=true per_token=true | first_token_eq baseline=true pin=true per_token=true
```

```bash
OPENINFER_TEST_MODEL_PATH=<Qwen3-4B-base> cargo test --release -p openinfer-qwen3-4b --test batch_invariance_decode_splitkv_graph -- --nocapture
```

## Checklist


xiaguan:
Correctness looks clean — batch-invariance holds (`pin_chunk_size` keys off the `max_context_tokens` constant, so a request's chunk count stays request-local), and the workspace bound, guard, and graph capture/replay grid are all consistent. The perf tradeoff is fine and clearly disclosed. A few cleanup nits, none blocking:

1. **Doc drift** — workspace is now `bs × 256`, but two docs still describe the old world:
   - `docs/models/qwen3/decode-attention.md:64` still says `bs × 64`
   - `docs/models/qwen3/model-crate.md:393` still describes the single-constant version

2. **Dead workspace on the default path** — `new()` sizes `max_split_slots` at `× 256` for every policy, but Tuned (the default) only ever uses 64 chunks, so ~3/4 (~37 MiB at bs=32) is never touched. You already added `max_split_chunks()` — using it here would size the workspace by policy. Optional.

3. **Naming** — `SPLIT_KV_MAX_CHUNKS_PER_REQUEST` now serves as the absolute bound, the Pin cap, and the `pin_chunk_size` divisor all at once, while the Tuned cap moved to its own constant. A one-line note on why the divisor must equal the cap (so the guard stays tight) would help the next reader.

LGTM otherwise.

xiaguan (approved):
LGTM. The CUDA Graph path stays consistent: split-KV workspace, metadata refresh, and replay grid are all tied to the active startup policy, and the previous docs/workspace cleanup comments are addressed.

PR #458, opened 2026-06-25
fix(qwen3): assert NumericPolicy is immutable post-construction

FeathBow:
## Description

Refs #414, #435. Follows #438

Decode reads the process-global `numeric_policy()` to (a) size the split-KV decode workspace and (b) select the GEMM algo baked into cached decode/verify CUDA graphs. Both are fixed at construction: the workspace is allocated for the construction-time policy; graphs are cached keyed `(bucket, attention_path)` / `(bucket, segment)` WITHOUT policy. A mid-run switch would overflow the workspace (`sync` writes `padded_bs × live-cap` slots into a construction-sized buffer) or replay a stale-policy graph — silently. Production sets the policy once at boot and never mutates it; this guard makes that invariant fail-loud.

Snapshots `policy_at_construction` on `BatchDecodeBuffers` and `VerifyGraphBuffers` and asserts `numeric_policy() == policy_at_construction`:
- at `batch_decode()` entry, unconditional, BEFORE the split-KV sync and the graph branch (covers eager + graph + LoRA decode);
- at the `VerifyGraphBuffers` capture/replay arm (covers the DFlash verify graph).

A trace of the relevant post-construction readers shows the guarded decode path reaches the split-KV readers (`max_split_chunks` / `split_chunk_size`) only after the entry assert, and cached decode/verify graph replay reaches policy-selected GEMMs only after their respective policy guards. Other reads are construction-time setup/snapshot reads, guard reads, or eager prefill/unified/drafter GEMMs that do not touch split-KV workspace or cached policy-keyed graphs.

## Test Env

`cargo test --release -p openinfer-qwen3-4b --test batch_invariance_decode_gemm_graph --test batch_invariance_decode_splitkv_graph`

sm_89 — both pass (per-policy executors build + decode; the entry assert holds, no false-fire):

    test result: ok. 1 passed   # decode_gemm_graph
    test result: ok. 1 passed   # decode_splitkv_graph

`batch_invariance_endtoend` (switches policy on a live executor but only `execute_prefill`s) still passes — the guard is scoped to the unsafe decode/verify paths, not a blanket reject.

## Type of Change

### DIFF
--- a/openinfer-qwen3-4b/src/batch_decode.rs
+++ b/openinfer-qwen3-4b/src/batch_decode.rs
@@ -62,6 +62,13 @@ impl Qwen3Model {
         assert_eq!(bs, kv_views.len());
         assert_eq!(bs, lora_adapters.len());
         assert!(bs > 0);
+        // Before the first live-policy read (sync_paged_meta) and on every decode path; see
+        // `policy_at_construction` for the workspace-overflow / stale-graph trap this guards.
+        assert_eq!(
+            openinfer_kernels::ops::numeric_policy(),
+            bufs.policy_at_construction,
+            "NumericPolicy changed after executor construction (policy-key-trap); build a fresh executor per policy"
+        );
         let lora_slots = self.decode_lora_slots(lora_adapters)?;
         let use_lora = lora_slots.is_some();
         let use_cuda_graph = self.enable_cuda_graph && lora_slots.is_none();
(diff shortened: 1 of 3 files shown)


xiaguan (approved):
LGTM. The policy lifetime guard is scoped to the decode/verify graph invariant and makes the existing construction-time assumption fail loud.

Issue #401, opened 2026-06-16
vllm-frontend: wire real KV cache ready_response fields from openinfer scheduler


xiaguan:
## Context

After bumping vllm git deps to latest `main` (`a7fdfeef`), `EngineCoreReadyResponse` gained three KV-cache metadata fields:

- `block_size`
- `kv_cache_size_tokens`
- `kv_cache_max_concurrency`

Our `LocalEngineBridge` currently sends placeholder values (`block_size: 16`, the two optional fields `None`) because openinfer does not run Python vLLM's KV profiling handshake.

Upstream uses these fields for group-aware KV capacity reporting and Prometheus `vllm:cache_config_info` labels. The Rust vLLM frontend stores them today but does not consume them yet.

## Scope

- Plumb real paged-KV block size and capacity from the openinfer scheduler/engine into `LocalEngineBridge`'s ready response.
- Decide whether `num_gpu_blocks` should also reflect scheduler state instead of the current `0` placeholder.
- Add/adjust tests once the frontend starts reading these fields (metrics or admission UX).

## Related

- Tracking issue for the vllm bump half of #364 (long-prompt 400 via vllm#45286).
- Remaining #294 scope (engine `Rejected` → HTTP 500 without message) stays separate.

xiaguan:
Done in #475 (`fix(vllm-frontend): report real KV capacity for engines that publish it`): the bridge's `EngineCoreReadyResponse` now carries real `num_gpu_blocks`/`block_size` and computes `kv_cache_size_tokens`/`kv_cache_max_concurrency` from `EngineHandle::kv_capacity()`; the 16/None placeholders remain only for engines that don't publish a capacity. The frontend consumes KV state beyond the ready response as of #601 (live `vllm:kv_cache_usage_perc` etc. — tracking in #602).

Issue #221, opened 2026-06-04
[Model] Kimi-K2 roadmap


xiaguan:
## Owner

Proposed owner: @xiaguan 
Crate: `pegainfer-kimi-k2`
Feature gate: `kimi-k2`
Docs:

- `docs/models/kimi-k2/roadmap.md` (source of truth for this issue)
- `docs/models/kimi-k2/tp1-dp8-ep8-performance.md` (active performance ledger)
- `docs/models/kimi-k2/optimization.md` (model card + TP8 history)

Related roadmap: #203
Related work: #204 (decode kernel picks), #216 (Qwen3 prefix cache — the pattern to port), #215 (server-wide abort-on-disconnect)

## Current State

Kimi-K2 (61-layer MLA + MoE, INT4 routed experts) serves on 8× NVIDIA H20 in two shapes: TP8+EP8 NCCL (CUDA-graph decode, history path) and TP1+DP8+EP8 PPLX EP (the active line). Decode performance already beats the vLLM out-of-box baseline on the same hardware: service bs64 output `1336` vs `594 tok/s`, TPOT p50 `47.3` vs `107.2 ms`.

The gap is everything around the decode loop. A 2026-06 review verified ~32 findings against the code; headline state:

| Area | State |
| --- | --- |
| Sampling | `temperature/top_k/top_p` silently ignored — every request decodes greedy |
| Stop conditions | no EOS/stop-token detection; every request generates exactly `max_tokens` |
| Admission | no KV budget; prompts beyond the per-slot KV capacity are not rejected |
| Batching | continuous on TP1/DP8 (`DpCoordinator`); batch-then-drain on TP8/DP1 |
| Accuracy | no gate reproducible from a fresh clone (reference fixtures and the kernel-pick A/B harness are unversioned) |
| Tests | the only model crate without a `tests/` directory |
| TTFT | p50 4.5× / p99 31× behind vLLM on the same hardware |
| Prefix cache | none (fixed per-slot KV arena, no block sharing) |

## Active Contract

- Greedy decode only; correctness evidence is greedy token-id parity against vLLM fixtures, run manually on the serving hardware.
- TP1+DP8+EP8 bs64 numbers are service-level (`/v1/completions`, same client as the vLLM baseline). No claim of: non-greedy sampling, EOS-terminated generation, prompts beyond the per-slot KV capacity, long-context decode correctness, prefix reuse, multi-node.

## Roadmap

[...]

PR #475, opened 2026-06-30
fix(vllm-frontend): report real KV capacity for engines that publish it

FeathBow:
## Description

Refs #401

The vLLM bridge used hardcoded KV-capacity placeholders for every engine. Report real `EngineHandle::kv_capacity()` for the engines that publish it, and wire qwen35 to publish the same metadata qwen3 already does.

## After

- qwen3 and qwen35 advertise real usable KV capacity from `EngineHandle::kv_capacity()`: `num_gpu_blocks` / `kv_cache_size_tokens` are the usable count (pool minus the CUDA-graph padding page), `kv_cache_max_concurrency` is vLLM's `num_gpu_blocks / ceil(max_model_len / block_size)`. qwen35 wires `.with_kv_capacity()`, mirroring qwen3.
- Engines that do not publish `kv_capacity()` keep the previous ready-response placeholders (`0 / 16 / None / None`), unchanged by this PR.

## Tests

Single GPU (sm_89, x86_64), `local engine KV capacity` startup log:
- qwen3 `num_gpu_blocks=15376 block_size=16 kv_cache_size_tokens=246016 kv_cache_max_concurrency=6.00625`
- qwen35 `num_gpu_blocks=61980 block_size=16 kv_cache_size_tokens=991680 kv_cache_max_concurrency=3.782958984375`

`frontend_e2e` (CI gate) covers the bridge path.

## Follow-up

kimi-k2 appears to have a real pool but does not publish capacity metadata yet, so it keeps the placeholder. Tracked as a sub-issue of #221.

### DIFF
--- a/openinfer-qwen35-4b/src/scheduler.rs
+++ b/openinfer-qwen35-4b/src/scheduler.rs
@@ -20,8 +20,8 @@ use crate::logprobs::snapshot_requested_logprobs;
 use crate::recurrent_state::RecurrentState;
 use crate::weights::Qwen35Model;
 use openinfer_core::engine::{
-    EngineHandle as SchedulerHandle, FinishReason, GenerateRequest as SchedulerRequest, TokenEvent,
-    TokenLogprob, TokenSink,
+    EngineHandle as SchedulerHandle, FinishReason, GenerateRequest as SchedulerRequest, KvCapacity,
+    TokenEvent, TokenLogprob, TokenSink,
 };
 use openinfer_core::kv_pool::KvState;
 use openinfer_core::sampler::SamplingParams;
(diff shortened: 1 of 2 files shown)


xiaguan:
Great little wiring PR — the real-value path reads well and the None fallback keeping the old placeholder is the right call.

Two nits, both non-blocking:

---

**nit: avoid re-opening the same `Option` four times**
`openinfer-vllm-frontend/src/bridge.rs:64-76` currently does four separate `kv_capacity.map_or(...)` / `map(...)` calls, each reopening the same `Option` and re-copying `c`. A single `match` constructs all four fields at once and reads a little cleaner:

\`\`\`rust
let (num_gpu_blocks, block_size, kv_cache_size_tokens, kv_cache_max_concurrency) = match kv_capacity {
    Some(c) => {
        let blocks_per_req = u64::from(self.max_model_len).div_ceil(c.block_size as u64);
        (
            c.total_blocks as u64,
            c.block_size as u64,
            Some(c.total_tokens() as u64),
            Some(c.total_blocks as f64 / blocks_per_req as f64),
        )
    }
    None => (0, 16, None, None),
};
\`\`\`

Not blocking — just a readability nit. 🙂

---

**nit: `kv_capacity` Debug is logged twice**
`bridge.rs:78-85` prints `{kv_capacity:?}` (which already serializes `Some(KvCapacity { total_blocks, block_size })`) and then immediately repeats `num_gpu_blocks={}` and `block_size={}` the line after. One of the two is enough; keeping both just makes the line longer than it needs to be. nit, feel free to ignore.

PR #496, opened 2026-07-02
fix(qwen3): disable cuda graph under tensor parallelism

### LINKED ISSUE #481: qwen3: default `--tp-size 2` (CUDA graph on) deadlocks on concurrent decode
### Summary

A Qwen3-4B server started with `--tp-size 2` (nothing else) hangs on the first burst of concurrent decode. A single request completes correctly; under concurrency the engine deadlocks silently — no response, no error, nothing logged after `Scheduler ready`. The separating variable is CUDA graph: `--tp-size 2 --cuda-graph=false` serves the same load cleanly. Since `--cuda-graph` defaults to on, the plain `--tp-size 2` invocation is the broken default.

Trace: the decode-graph capture closure (`run_or_capture` in `batch_decode`) records each layer's two NCCL all-reduces into the graph, with nothing aligning capture across ranks.

### Reproduction

Test Env: 2x GH200 (sm_90, aarch64)

32 staggered concurrent completion requests (temp 0, max_tokens 96), fresh server per config. This is normal serving load: the engine itself advertises ~28 max KV concurrency and queues beyond it, and 16-concurrent (under that ceiling) also hangs.

- Single request, `--tp-size 2`: correct output.
- 32 concurrent, `--tp-size 2`: 0 completions, engine wedged until killed.

### Control matrix

| # | tp | cuda-graph | decode-overlap | load | result |
|---|----|-----------|----------------|------|--------|
| 1 | 1 | on | off | 32 | ok 32/32 |
| 2 | 2 | **off** | off | 32 | ok 32/32 |
| 3 | 2 | **on** | off | 32 | **hang 0/32** |
| 4 | 2 | on | green-ctx | 32 | hang 0/32 |
| 5 | **1** | on | green-ctx | 32 | ok 32/32 |
| 6 | 2 | on | off | **16** | hang 0/16 |
| 7 | 2 | on | green-ctx | 16 | hang 0/16 |

### Acceptance

- [ ] Force the graph off when `tp_size > 1`, with a warn — same pattern as the existing LoRA graph clamp
- [ ] A TP=2 test at production defaults (graph on) under concurrency — the existing tp2 golden-gate pass is eager-only and single-request, so it cannot catch this.

FeathBow:
## Description

Fixes #481 

`--tp-size 2` at defaults deadlocks silently on concurrent decode: `--cuda-graph` defaults to on, and the decode-graph capture closure records each layer's two NCCL all-reduces into the graph with nothing aligning capture across ranks. A single request completes; the first concurrent burst wedges the engine with nothing logged. `--tp-size 2 --cuda-graph=false` serves the same load cleanly.

Two layers, so the default keeps working and lower entries fail closed:

- `launch()` clamps the graph off when `tp_size > 1`, with a warn — the same pattern as the existing LoRA clamp there. The plain `--tp-size 2` invocation downgrades to eager and serves.
- The executor constructor rejects `enable_cuda_graph` with multiple devices — the same `ensure` the KV-offload / KV-events / DFlash guards already use, closing the direct `start_engine*` / `from_runtime*` entries that bypass `launch()`.

## Test

`tests/tp_concurrent_decode.rs`: TP=2 at production defaults (graph requested) under 16 concurrent requests. The drain loop polls with a deadline, so a reintroduced deadlock fails the test instead of wedging the run. Skips without >=2 GPUs or weights. The test waits briefly after dropping the engine so its asynchronous teardown finishes before process exit.

2x GH200 (sm_90, aarch64), A/B against the same origin/main base, one variable.

fire 32 staggered concurrent completions (temp 0, max_tokens 96):

| side | result |
|------|--------|
| base | hang: 0/32 completions, engine wedged until killed, no warn |
| base + fix | 32/32 completed; `Qwen3: CUDA Graph is disabled under tensor parallelism` logged once |

Test red/green — same command on both sides:

```
OPENINFER_TEST_MODEL_PATH=<Qwen3-4B> cargo test --release -p openinfer-qwen3-4b \
    --test tp_concurrent_decode -- --nocapture --test-threads=1
```

Base + test only (red — the 300s drain deadline fires on the deadlock):

```
failures:
    tp2_concurrent_decode_completes

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 311.81s
```

Base + fix (green):

```
test tp2_concurrent_decode_completes ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 11.15s
```

### DIFF
--- a/openinfer-qwen3/src/executor.rs
+++ b/openinfer-qwen3/src/executor.rs
@@ -1074,6 +1074,13 @@ impl Qwen3Executor {
             "KV block events are only supported on the single-GPU path; got {} devices",
             device_ordinals.len()
         );
+        anyhow::ensure!(
+            !enable_cuda_graph || device_ordinals.len() == 1,
+            "CUDA Graph is not supported under tensor parallelism (the captured decode \
+             graph would record NCCL collectives with no cross-rank capture alignment); \
+             got {} devices",
+            device_ordinals.len()
+        );
         // The store cursor announces a block as cacheable the moment it is
         // registered, assuming GPU-resident reuse. With KV offload on, a block
         // can be evicted to the host tier and restored under a different lineage
--- a/openinfer-qwen3/src/lib.rs
+++ b/openinfer-qwen3/src/lib.rs
@@ -156,16 +156,17 @@ pub fn probe_model(model_path: &Path) -> Result<Option<ModelInfo>> {
 /// Server-facing launch knobs for the Qwen3 engine.
 ///
 /// The binary maps raw CLI flags into this struct; [`launch`] then owns the
-/// Qwen3 startup policy — the TP→device mapping and the LoRA↔CUDA-Graph
-/// exclusion — and dispatches to the right low-level entry. That policy lives
-/// with the model instead of leaking into the server.
+/// Qwen3 startup policy — the TP→device mapping and the LoRA↔CUDA-Graph and
+/// TP↔CUDA-Graph exclusions — and dispatches to the right low-level entry.
+/// That policy lives with the model instead of leaking into the server.
 #[derive(Clone, Debug)]
 pub struct Qwen3LaunchOptions {
     /// CUDA device for single-GPU loads (ignored when `tp_size > 1`).
     pub device_ordinal: usize,
     /// Tensor-parallel world size; `> 1` uses devices `0..tp_size`.
     pub tp_size: usize,
-    /// Whether the user requested CUDA Graph. LoRA serving forces it off.
+    /// Whether the user requested CUDA Graph. LoRA serving and tensor
+    /// parallelism force it off.
     pub cuda_graph: bool,
     pub offload: Qwen3OffloadOptions,
     pub no_prefix_cache: bool,
@@ -193,10 +194,16 @@ pub fn launch(model_path: &Path, options: Qwen3LaunchOptions) -> Result<EngineHa
         (0..options.tp_size).collect()
     };
     // LoRA serving repoints adapter weights between steps, which a captured
-    // decode graph bakes in — the two cannot coexist, so LoRA wins.
+    // decode graph bakes in; under TP the graph would capture each layer's
+    // NCCL all-reduces with nothing aligning capture across ranks.
     let enable_cuda_graph = if options.lora.is_some() {
         if options.cuda_graph {
             warn!("Qwen3: CUDA Graph is disabled while LoRA serving is enabled");
+        }
+        false
+    } else if options.tp_size > 1 {
+        if options.cuda_graph {
+            warn!("Qwen3: CUDA Graph is disabled under tensor parallelism");
         }
         false
     } else {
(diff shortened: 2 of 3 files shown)


xiaguan:
Thanks for the precise root-cause diagnosis — the deadlock from NCCL collectives captured without cross-rank alignment is exactly right.

That said, I don't think clamping `cuda_graph=false` under TP should be the landing point. TP + CUDA Graph is a production-default configuration in vLLM and every other serving engine; Qwen3 is the flagship model, and shipping TP meaningfully slower than single-GPU is a regression we'd have to unwind anyway.

The current capture path violates three things NCCL graph capture requires:

1. **No dedicated capture stream.** `run_or_capture` captures on the compute stream (`active_cu_stream`). When one rank enters capture, its NCCL all-reduces become no-op recordings, but a peer still executing eager work on the previous step's tail blocks waiting for a matching collective that never arrives. vLLM uses a dedicated capture stream (`graph_capture` context) so capture never races with in-flight eager NCCL calls.

2. **No cross-rank capture alignment.** Each rank independently decides capture-vs-replay based on `self.exec.is_null()`. First hit on a new bucket → one rank is capturing (NCCL no-ops) while another is still in eager (NCCL blocking) → deadlock. A CPU barrier before entering capture — all ranks agree "we're capturing bucket N now" — closes this. vLLM skips the explicit barrier but guarantees alignment structurally (every rank enters `graph_capture` together); a barrier is cheaper insurance and fails fast instead of hanging silently.

3. **Replay not aligned.** Even after successful capture, each rank independently `cuGraphLaunch`es. `run_step` dispatches concurrently, but scheduling jitter means one rank's graph all-reduce can execute before the peer has launched → deadlock. The same bucket must be captured on all ranks before any rank replays it.

The structural invariant underneath all three: every rank captures the same bucket list, in the same order, with the same model → identical NCCL collective sequences in every graph. NCCL natively supports graph capture (collectives on a capturing stream are recorded, not executed); the fix is about *when* and *where* we capture, not whether NCCL can do it.

Proposed direction:
- Add a dedicated capture stream to `CudaGraphState` (switch to it for capture, back to compute stream for replay).
- Add a CPU barrier in the TP `run_step` path before any rank enters capture for a new bucket (reuse the existing `wait_for_step_ack` pattern).
- Gate replay on all-ranks-captured for a given bucket (track per-bucket capture state across ranks).

Happy to help scope this, or split it: land the deadlock repro test now (it's valuable regardless), and tackle the real fix in a follow-up. But I'd rather not ship the clamp as the fix.

FeathBow:
Agreed on the direction, and happy to take the split. Thanks for your help :)

One gap in the test-only version of it, though: without the clamp the repro test stays red — the default --tp-size 2 keeps deadlocking under concurrent load until the capture work lands. That's what the clamp is for: it turns the deadlock into eager serving, so the default is usable in the meantime and the gate is green. It doesn't slow down anything that works today (graph-on TP never seems to survive concurrency), and the real fix simply deletes both guards.

PR #511, opened 2026-07-02
fix(qwen3): join the scheduler thread on engine drop

FeathBow:
## Description

Follow up #496 

Dropping the last `EngineHandle` only closes the submit channel; the scheduler thread then tears down the executor (TP rank workers, NCCL/CUDA destruction) asynchronously. A process that drops the handle and exits races that teardown — SIGSEGV after the test harness has already reported pass. Production servers never drop the engine before exit, so only tests hit this; the TP=2 gate was the first to.

qwen3's two engine constructors keep the scheduler thread's `JoinHandle` and build the handle via the existing join-owning constructors (`EngineHandle::new_with_join_handle` / `new_with_command_channel_and_join_handle`) — the same wiring deepseek-v2-lite, glm52, kimi-k2, and qwen35 already use. `EngineInner::Drop` then closes the channels and joins the thread, so teardown completes before the drop returns. No engine or executor changes: the handle-side join and the per-worker joins already exist.

`tp_concurrent_decode` drops its post-drop sleep workaround. The handle sits in `ManuallyDrop`: a panic (worst case the deadline, engine deadlocked) leaks the engine instead of joining a wedged thread; the happy path drops (joins) explicitly.

## Verification

Single GPU (sm_89, x86_64), on current origin/main:

- Workspace lib tests pass, including the engine join unit test.
- `sampling_behavior`, `context_window`, `scheduler_robustness`, and `lora_smoke -- --ignored` all pass — every engine drop in these now exercises the synchronous join.

## Type of Change

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3809,6 +3809,7 @@ dependencies = [
 name = "openinfer-engine"
 version = "0.1.0"
 dependencies = [
+ "log",
  "tokio",
 ]
 
--- a/openinfer-engine/Cargo.toml
+++ b/openinfer-engine/Cargo.toml
@@ -5,6 +5,7 @@ version = "0.1.0"
 edition = "2024"
 
 [dependencies]
+log = { workspace = true }
 tokio = { workspace = true, features = ["sync"] }
 
 [dev-dependencies]
--- a/openinfer-engine/src/engine.rs
+++ b/openinfer-engine/src/engine.rs
@@ -1,4 +1,5 @@
 use std::{
+    any::Any,
     error::Error,
     fmt,
     path::PathBuf,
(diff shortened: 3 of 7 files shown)


xiaguan:
Soft follow-up (not in this PR's diff, doesn't block): `EngineInner::Drop` in `openinfer-engine/src/engine.rs:561` does `let _ = join_handle.join();`, which silently discards a scheduler-thread panic. Before this PR qwen3 detached the scheduler thread, so the swallow was moot; now that qwen3 wires the join, a scheduler panic becomes "thread exited, nobody knows." Worth a follow-up adding `log = { workspace = true }` to `openinfer-engine` (matching the rest of the workspace) and:

```rust
if let Some(join_handle) = self.join_handle.take() {
    if join_handle.thread().id() != thread::current().id() {
        if let Err(panic) = join_handle.join() {
            log::warn!("scheduler thread panicked during shutdown: {panic:?}");
        }
    }
}
```

Issue #542, opened 2026-07-03
glm52: close the last ~3 ms/step to the vLLM DP8/EP8 reference — rank stagger + collective latency


xiaguan:
After #TBD (PR5c whole-step decode graph), GLM5.2 DP8/EP8 decode sits at **25.3 ms/step** (jz-38 8×H200, single request; 8-way concurrent identical at 25.0, ~308 tok/s aggregate). vLLM DP8/EP8 reference is ~22 ms. nsys (node-granularity, proportions only) puts the remaining gap in the collectives:

- `dispatch_impl`: med 12.5 µs but sums to ~6 ms/step — wait-inclusive time (per-layer straggler jitter from uneven expert hits + inter-step rank launch stagger; max instances 24 ms sit at step boundaries).
- `combine_impl`: stable 42 µs × 75 layers ≈ 3 ms — already device-bounded via the vendored capacity-sentinel (verified `kDecodeWorstRecvTokens == kNumMaxTokensPerRank×kNumRanks` triggers the psum read); this is the cooperative-launch + NVLink push latency floor of the combine + reduce-epilogue pair.

Follow-ups:
1. Per-rank graph-launch timeline (not kernel summaries) to quantify inter-step stagger; then shave it — single-packet D2H egress (argmax value+index in one 8-byte buffer), tighter coordinator fan-out.
2. Collective SM allocation sweep (`kDecodeNumSms`, `kCombineWarps`) — measure dispatch/combine latency vs SM count at the 8-token decode payload.

Acceptance: single-request ms/step ≤ 22 on jz-38 with the PR5c gate suite green (determinism, 8-way identical, oracle probes, teardown).

Evidence: docs/models/glm52/whole-step-decode-graph.md ("Where the remaining ~30 ms sits" + follow-ups section).

xiaguan:
Data point: `kDecodeNumSms` 32→16 measured **exactly flat** (25.3 ms/step, byte-parity pass) — combine's 42 µs/layer is protocol/NVLink latency, not grid-sync/SM-count-bound. The SM-allocation sweep (follow-up 2) is likely a dead end below 32; the remaining lever is follow-up 1 (rank-stagger timeline + single-packet egress + coordinator fan-out).

xiaguan:
Two more measured dead ends + the scoped next lever (all on jz-38, same probe):

[...]

Issue #541, opened 2026-07-03
glm52: indexer oracle gate reference drift — HF glm_moe_dsa indexer is a moving target


xiaguan:
## Symptom

\`indexer_oracle_gate\` scores **~1585/2048 set-overlap (< 2000 threshold) against any freshly generated oracle**, on main's indexer numerics as well as on \`feat/glm52-decode-graph\`. The #521-era green (2013/2048) is not reproducible: the reference itself moved.

## Evidence matrix (jz-38, ctx 4096, seed 0x5eed604d, layer 0)

| engine | oracle | overlap |
|---|---|---|
| pre-GEMV (8cc3f73 ≈ main numerics) | fresh dev0-fp8sim | **1583** |
| GEMV (d41bef8) | fresh dev0-fp8sim | 1588 |
| GEMV | fresh dev0-gemv | 1591 |
| GEMV | fresh 5.13.0-release fp8sim | 1588 |
| GEMV | **stale #521-era file** (now lost) | 1998 |
| fresh oracles vs each other | | 2000 |

Two independent upstream moves:
1. **transformers 5.13.0 release regressed PR #46842**: the glm_moe_dsa indexer applies non-interleaved RoPE again, contradicting the checkpoint's \`indexer_rope_interleave=true\`. \`tools/accuracy/glm52_oracle.py\` now refuses to run on such builds (source-inspection guard).
2. **\`transformers==5.13.0.dev0\` is a moving git target**: the repo .venv on jz-38 was rebuilt 2026-07-03 00:16 (commit \`0de4bbd1\`), after #521 generated its probes; the indexer changed between dev0 snapshots beyond the rope fix (~430/2048 slot shift for identical inputs — hidden digest matches).

## Acceptance

- Pin the oracle's transformers source to an exact commit whose glm_moe_dsa indexer matches the production reference (vllm \`DeepseekV32Indexer\` semantics the engine was aligned to in #521), or switch the indexer oracle to a vllm-derived reference.
- Regenerate the indexer safetensors + re-establish the overlap threshold from a reproducible reference; gate green on main and on the PR5c branch.
- Note: e2e coverage cannot catch this — at ctx ≤ 2048 top-k = all tokens, so the sparse selection path only runs under this gate.

PR #543, opened 2026-07-03
perf(glm52): whole-step decode CUDA graph + weight-only GEMV — 200 → 25.3 ms/step

xiaguan:
## What

PR5c of the GLM5.2 DP1/EP8 campaign (`docs/models/glm52/dp1-ep8-decode-plan.md`): the per-rank decode step (embed → 78 layers → lm_head → argmax) becomes **one CUDA graph launch per rank per step**, followed by the kernel- and overlap-level cuts its nsys profile pointed at. Targets the fixed ~200 ms/step cost measured after #537.

## A/B (jz-38 8×H200, 133-step greedy request, `/v1/completions` wall)

| stage | ms/step | vs prev |
|---|---|---|
| PR5b baseline (#537) | ~200 | |
| whole-step graph (arena + host-quiet step + capture) | 37.5 | 5.3× — **byte-identical** to the PR5b/PR5a record |
| + weight-only fp8 GEMV for every m=1 projection | 31.3 | numerics change, re-gated (below) |
| + device-bounded, grid-strided MoE recv quant/SiLU | 25.3 | bit-preserving; block *scheduling*, not arithmetic, was the cost |
| + gate\|up packed GEMV + shared-expert ∥ collectives + smem relayout ranges | 23.4 | bit-preserving; overlap = fork/join events inside the capture |
| + closing-add fused with next input-norm + indexer ∥ MLA front | 22.9 | bit-identical `_round` fusion; indexer forks after q_resid, joins before attend |
| + two-tier attention graphs (short-context FlashMLA topk 256) | 22.6 | same attended tokens below seq 256 (top-256 = full set, like top-2048 below 2048), 1/8 the padded index walk — the V3.2 sparse kernel always walks all topk slots |
| + two-stage device argmax (the qwen split kernel) | **22.5** | bit-identical (partials carry global indices, same lowest-index total order); the single-CTA 154k-vocab scan was the last serial kernel in the step; e2e parity vs the v7 reference PASS (24-tok and 128-tok byte-identical) |

8-way concurrency stays free throughout: **22.3 ms/step concurrent, ~346 tok/s aggregate** (measured at the tier stage) — **~8.9× end to end**. The final-state per-kernel profile (where each of the 22.5 ms sits, with the trace-vs-e2e calibration rule) is recorded in `docs/models/glm52/whole-step-decode-graph.md`. Measured against vLLM 0.24.0 GLM5.2 DP8/EP8 on the same node/workload (steady 20.0 ms/step): the remaining 2.6 ms sits mostly in the expert GEMM (their DeepGEMM swapAB ~10.8 µs/instance vs our 64-row-M-tile TRTLLM grouped at 22.7 µs against ~8 real rows) and the collectives' wait structure — scoped in #542 together with the measured dead ends (fp8 dispatch payload proven bit-identical but perf-NEUTRAL — dispatch is rank-arrival-wait-bound, not byte-bound; `kDecodeNumSms` 16/32/64 all flat; `num_sm_parts` reduction regresses at topk 2048; L2 weight prefetch regresses; device self-feeding pointless at a 0.05 ms host gap).

## How

- **Stage 1 — arena** (`Glm52DecodeScratch`): every per-step intermediate allocated once at model build; pointer-stable, allocation-free step. Byte-identical.
- **Stage 2 — host-quiet step**: indexer weights-fold on-device (killed the only 2 mid-step D2H syncs), MLA cache-pack reads its slot from the device `slot_mapping` (traps out-of-window), device argmax egress (6-byte D2H, lowest-index-wins matching the retired host scan). Byte-identical.
- **Stage 3 — capture**: `CudaGraphState::run_or_capture` around the step body; prologue (rope row + slot/seq/token uploads) and egress stay outside. In lock-step all ranks capture on the same global step, so the DeepEP collectives first execute together on every rank's first graph launch — **cooperative+cluster shim launches replay fine under capture** (the kimi decode-graph argument held).
- **Weight-only GEMV**: TRTLLM CUTLASS pads M 1→64 at decode; the m=1 projections + their activation quant + TMA relayout were ~45% of step kernel time. All non-expert projections now run the PP8-campaign weight-only GEMV (bf16 activation, on-the-fly e4m3 block-scale dequant, 1 kernel instead of 3); gate|up pairs are packed into one `[2I, k]` weight at build so a single GEMV writes the exact `[gate | up]` layout the SwiGLU consumes (no concat copies). The EP8 routed-expert chain stays on grouped CUTLASS.
- **Grid-strided quant/SiLU**: the capacity-sized (2080-row) recv-chain launches spent ~60 µs/layer on block scheduling (~100k tiny blocks) that device-bound early-returns cannot recover; the row grid is capped at 256 with an in-block loop bounded by `expert_offsets[n_local]` on device.
- **Shared expert ∥ MoE collectives**: the shared expert only needs the post-attention norm, so it runs on a second stream forked/joined with events around dispatch/grouped-GEMM/combine (the kimi decode-graph overlap pattern) — the cooperative collectives hold a fixed SM slice and mostly wait on peers, so the shared GEMVs fill otherwise-idle SMs. Events recorded at capture become graph edges.
- **Two-tier attention graphs**: a same-node nsys diff against vLLM showed our sparse attention 3× theirs per layer — the V3.2 sparse kernel walks all `topk` index slots (-1 pads run the full load+GEMM, masked only in softmax), so a 139-token context paid the fixed top-2048 walk every layer. While `seq_len <= 256` the DSA top-256 IS the full token set, so each rank keeps one whole-step graph per tier (`topk` 256 / 2048) and picks by position; the DSA indexer's FlashInfer top-k narrows with the attend plan (read from the sched metadata — a mismatch is unrepresentable). `topk` was already a runtime parameter of the vendored kernel (64-alignment is its only assert); the fixed-2048 whitelist was ours. The lazily captured full graph records without executing, so a mid-serving tier crossing holds peers for tens of ms against the ~100 s DeepEP timeout.

## Numerics & gates

Stages 1–3 are byte-identical to the PR5b record (verified). The GEMV switch removes the activation-quant error term and the short tier changes the attend split partition, so outputs are deliberately not byte-identical to the old records; accuracy authority is the #499 oracle harness (`--precision gemv` mode: bf16 activations, fp8 kv-cache + fp8 routed-expert emulation kept):

- MLA oracle gate: **64/64 probes** — **both tiers** (new `mla_oracle_gate_short_tier`: the topk-256 plan must land on the same oracle within the same tolerance; ORACLE_CTX=200 fits the tier)
- decoder-layer gates: dense 64/64; MoE layer-6 62/64 on both Grouped and Gemv expert paths (same tie-flip outliers as before)
- EP8 layer gate (real 8-GPU collective): green
- e2e (v7 suite): determinism ×2, 8 identical concurrent byte-identical, **tier-crossing request ×2 deterministic with the 128-tok output a prefix of the 320-tok one**, **mixed-tier concurrency (1×320 + 7×128)**, disconnect, SIGTERM teardown — all green; new reference recorded on jz-38

**Known-red:** the indexer oracle gate — a pre-existing reference-drift problem, not introduced here: main scores the same ~1585/2048 against any freshly generated oracle because the HF `glm_moe_dsa` indexer is a moving target (the 5.13.0 release also regressed the RoPE-interleave fix). Evidence matrix and acceptance in #541; the oracle script now refuses to run on interleave-regressed transformers builds.

## Follow-ups

- #542 — the last ~2.6 ms to the measured vLLM 20.0: expert-GEMM swapAB-vs-M-tile (the biggest single item), collective wait structure; fp8-dispatch correctness is proven and archived on `wip/glm52-fp8-dispatch`.
- #541 — pin the indexer oracle reference.

Execution record with the full nsys breakdown and measured dead ends: `docs/models/glm52/whole-step-decode-graph.md`.


xiaguan:
Toxic-review pass (internal): **no fatal/hard findings**. Soft findings recorded for follow-up:

1. **Egress is two D2H round-trips, not one** — `argmax_value` and `argmax_index` are separate 1-element readbacks (`model.rs`); packing them into one 8-byte buffer makes it a single sync. Noise at 25 ms/step, but the doc says "6-byte egress" and the implementation should match.
2. **`fp8_mlp_into` gate|up concat is ~156 redundant graph copy-nodes/step** — the two GEMVs could write the halves of `gate_up` directly if the GEMV launcher accepted an output offset/view.
3. **slice-twin wrapper family is growing** (`add_into`, `fused_add_rms_norm_round_into`, `argmax_bf16_into` — same kernels as their `HiddenStates` twins): third occurrence makes it a pattern; the owned-wrapper variants should be reimplemented over the slice variants.
4. **`CudaGraphState::run_or_capture` panic path leaves the stream capturing** (pre-existing; this PR is its heaviest user) — worth a capture guard or at least a comment in `openinfer-core/src/cuda_graph.rs`.
5. `Glm52LayerScratch.attn` carries the residual stream after the fused add — naming tax.

Verified explicitly during review: no host value baked into the captured region; scratch aliasing safe (stale rows provably unread — trap-guarded); loop-internal `__syncthreads` uniform; device argmax tie/NaN semantics case-equivalent to the removed host scan; metadata-before-quant reorder has no hidden dependency.

PR #544, opened 2026-07-04
fix(qwen35): load the untied lm_head instead of reusing embed_tokens

### LINKED ISSUE #516: qwen35: Qwen3.5-9B produces repeated-token output (hidden_size=4096)
## Description

Found while verifying #510 on a GH200 (openinfer `5959f05`, `--features qwen35-4b`).

Qwen3.5-9B loads and serves, but every tested greedy completion is repeated-token output:

```
cargo run --release --features qwen35-4b -- --model-path models/Qwen3.5-9B
# "The capital of France is" -> "锻焊工另一个 Pompe немедленно…Blog simplistic simplistic simplistic…"
# 4201-token input -> "_edges distinction distinction…BackBackBackBack dup…市民市民"
```

`transformers` on the same model dir, same venv, same input ids (`[760, 6511, 314, 9338, 369]`, chat template not applied on either side) produces coherent text, so this is not a corrupt checkpoint or tokenizer; the divergence is somewhere in openinfer's 9B path (the exact kernel/line is not yet located). The layer/head geometry is identical to the working 4B; what changes is `hidden_size` (2560→4096) and its hidden-derived dims (e.g. `intermediate_size`). Head counts being identical means this is not the 27B value-head AOT guard failure — no startup AOT mismatch is reported for 9B. Deterministic (temp 0), reproduced across three independent runs.

Of the existing qwen35-4b tests pointed at 9B, `chunked_prefill`, `sampling_behavior`, `e2e_scheduler` pass on the garbage output (mechanism, not accuracy); `hf_golden_gate` fails (4B-specific fixtures).

FeathBow:
## Description

Fixes #516.

Qwen3.5-9B and 27B set `tie_word_embeddings: false` and ship a top-level `lm_head.weight`. The qwen35 crate unconditionally reused `embed_tokens` as the LM head, so on 9B every completion projects final hidden states through the wrong matrix and collapses into repeated tokens.

Root-cause check: forcing HF 9B to reuse its input embeddings as the LM head reproduces openinfer's broken output token-for-token, while stock HF 9B is coherent. The divergence is exactly the tied/untied choice.

`config.rs` parses `tie_word_embeddings` from the root or `text_config` level and rejects a config with neither. `weights.rs` loads `lm_head.weight` through the loader's fallback-aware path when untied (matching the qwen3 crate — a missing tensor fails loud), validates its shape, and `output_projection()` picks the matrix at the prefill / batch-decode / GEMM-tune sites.

## Test Env

Single GPU (sm_89, x86_64): 4B golden gate green (tied path unchanged). 9B with the fix — the degenerate-output sentinel (10 greedy completions) is green; greedy completions are coherent. Dual-GH200 (aarch64, sm_90): 9B with the fix produces the same coherent greedy output, and a long-context prompt that was garbage on the unfixed build is coherent. The same sentinel is 6/10 degenerate on the unfixed 9B (measurement below).

<details>
<summary>9B greedy completions with the fix (sm_89)</summary>

```
The capital of France is | Paris.
Water boils at | 100°C at sea level. At what temperature does water boil at the top of Mount Everest, where the
The chemical symbol for gold is | Au. What is the origin of this symbol?
In 1969, Apollo 11 | landed on the Moon, and the astronauts planted a flag.
The speed of light is approximately | $3 \times 10^8$ meters per second.
```

</details>

<details>
<summary>Same sentinel on the unfixed 9B — 6/10 degenerate (Dual-GH200, aarch64, sm_90)</summary>

```
DEGENERATE  len=50 distinct_ratio=0.160 max_run=43   prompt: The capital of France is
DEGENERATE  len=50 distinct_ratio=0.340 max_run=18   prompt: Water boils at
DEGENERATE  len=50 distinct_ratio=0.200 max_run=27   prompt: The chemical symbol for gold is
DEGENERATE  len=50 distinct_ratio=0.200 max_run=20   prompt: In 1969, Apollo 11
DEGENERATE  len=50 distinct_ratio=0.200 max_run=28   prompt: The speed of light is approximately
DEGENERATE  len=50 distinct_ratio=0.040 max_run=49   prompt: def fibonacci(n):
```

</details>

## Type of Change

### DIFF
--- a/docs/models/qwen35/optimization.md
+++ b/docs/models/qwen35/optimization.md
@@ -92,7 +92,7 @@ Reference — Qwen3-4B on same GPU: TTFT(2048,1)=213ms, TPOT(1,128)≈10.6ms.
 - hidden_dim: 2560
 - MLP intermediate_size: 9216
 - RMSNorm: (1 + weight) offset variant, eps=1e-6
-- Tied word embeddings (embed_tokens = lm_head)
+- Word embeddings: 4B ties them (embed_tokens doubles as the LM head); 9B/27B are untied with a top-level `lm_head.weight`
 - Vocab size: 248,320
 
 **Full attention (8 layers):**
--- a/openinfer-qwen35-4b/src/batch_decode.rs
+++ b/openinfer-qwen35-4b/src/batch_decode.rs
@@ -319,7 +319,7 @@ impl Qwen35Model {
         )?;
         ops::gemm_into(
             &self.ctx,
-            &self.embed_tokens,
+            self.output_projection(),
             &bufs.normed,
             &mut bufs.logits,
         );
--- a/openinfer-qwen35-4b/src/config.rs
+++ b/openinfer-qwen35-4b/src/config.rs
@@ -32,6 +32,7 @@ struct TextConfig {
     linear_value_head_dim: usize,
     rope_parameters: RopeParameters,
     max_position_embeddings: Option<usize>,
+    tie_word_embeddings: Option<bool>,
     eos_token_id: u32,
 }
 
@@ -39,6 +40,7 @@ struct TextConfig {
 struct RawConfig {
     text_config: TextConfig,
     max_position_embeddings: Option<usize>,
+    tie_word_embeddings: Option<bool>,
 }
 
 /// Qwen3.5 model configuration (text-only).
@@ -71,6 +73,9 @@ pub(crate) struct Config35 {
 
     // Layer layout
     pub(crate) layer_types: Vec<LayerType>,
+
+    /// `false` requires a top-level `lm_head.weight`; `true` reuses `embed_tokens`.
+    pub(crate) tie_word_embeddings: bool,
 }
 
 /// GDN dims the Triton-AOT kernels are built for; a mismatched model is rejected at load.
@@ -84,7 +89,13 @@ impl Config35 {
         let content = fs::read_to_string(&config_path)?;
         let raw: RawConfig = serde_json::from_str(&content)?;
         let root_max_position_embeddings = raw.max_position_embeddings;
+        let root_tie_word_embeddings = raw.tie_word_embeddings;
         let t = raw.text_config;
+
+        let tie_word_embeddings = t
+            .tie_word_embeddings
+            .or(root_tie_word_embeddings)
+            .ok_or_else(|| anyhow::anyhow!("Qwen3.5 config missing tie_word_embeddings"))?;
 
         let layer_types: Vec<LayerType> = t
             .layer_types
@@ -147,6 +158,7 @@ impl Config35 {
             rotary_dim,
             max_position_embeddings,
             layer_types,
+            tie_word_embeddings,
         })
     }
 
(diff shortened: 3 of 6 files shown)


xiaguan (approved):
## Review: fix(qwen35): load the untied lm_head instead of reusing embed_tokens

### Link map

```
config.json
  ├─ text_config.tie_word_embeddings  ─┐
  └─ tie_word_embeddings (root)        ─┴─→ Config35.tie_word_embeddings (bool, crashes if missing)
                                          │
weights.rs:134  if tied  → lm_head=None;            output_projection()=&embed_tokens
              if untied → load "lm_head.weight" + shape check; output_projection()=&lm_head
                                          │
              ┌───────────────────────────┴──────────────────────────┐
              ▼                              ▼                       ▼
  prefill.rs:113  ops::gemm(...)      batch_decode.rs:322           weights.rs:502
  (batch_last_hidden_logits)          ops::gemm_into(...)          gemm_lt_tune(lm_head_samples)
                                       (CUDA Graph capture path)    (cuBLASLt plan tuning)

Embedding lookups (token_id→hidden) still go through embed_tokens:
  prefill.rs:151, batch_decode.rs:246
```

All three lm_head GEMM sites are routed through `output_projection()`; both embedding lookups correctly keep `&self.embed_tokens`. No missed call sites. `unified_forward.rs` / `executor.rs` go through `batch_prefill_logits` → `batch_last_hidden_logits`, covered.

### First-principles

Real bug: 9B/27B ship `tie_word_embeddings=false`; the old code reused `embed_tokens` as the LM head, producing degenerate output (PR shows 6/10 degenerate sentinel completions on unfixed 9B). Fix is minimal and correct — pick the matrix by config flag, validate shape, route through a 1-line helper. Complexity is proportional to the problem. Matches the qwen3 crate's `output_projection()` implementation verbatim (including `pub(super)` visibility). `cargo check` + 30 lib tests green; 4B tied path unaffected.

### Fatal (must fix)
None.

### Hard (will regret)
None. CUDA Graph pointer stability OK (`lm_head` allocated at load, stable thereafter); shape check `[vocab, hidden]` matches `embed_tokens`; missing config field crashes early per repo convention.

### Soft (eyesore)

**S1. Zero automated regression coverage for the untied path.**
`hf_golden_gate.rs` only covers 4B (tied=true). The 9B/27B fix relies entirely on manual verification (sm_89 + sm_90). Consider: (a) synthetic untied config + 4B weights (clone `embed_tokens` as `lm_head.weight`, set `tie=false`) to exercise the untied load path against golden logits; (b) a load-time norm assertion that untied `lm_head` differs from `embed_tokens` (9B does). A correctness bugfix with no regression gate means CI stays green if this regresses.

**S2. `info!` vs `debug!` inconsistency.**
`weights.rs:135,147` use `info!("output projection: ...")`, while embed_tokens loading in the same file (line 129-131) uses `debug!`, and qwen3's equivalent (`weights.rs:374,377`) also uses `debug!`. Two extra info lines per model load is noise for multi-model serving. Align to `debug!` — unless tied/untied is considered ops-critical to surface at info level, in which case align qwen3 too.

**S3. Silent text_config-over-root precedence when both present.**
`config.rs:95-98` `t.tie_word_embeddings.or(root_tie_word_embeddings)` silently picks `text_config` when both exist and disagree. Matches the existing `max_position_embeddings` pattern (local convention), but an `ensure!(t.tie == root, ...)` when both are present would catch hand-edited configs. Low priority.

### Verdict

**Approve (reluctant)** — fix is correct, minimal, aligned with qwen3, manually verified on two architectures. The only real discomfort is zero test coverage on the untied path (S1); doesn't block merge, but a follow-up regression gate would be welcome.

PR #547, opened 2026-07-04
feat(glm52): continuous batching — multi-slot admission + {1,8} batch-bucket graphs (D2)

xiaguan:
## What

D2 of the GLM5.2 EP8 decode campaign: the scheduler admits up to `GLM52_MAX_BATCH_PER_RANK = 8` requests per rank (least-loaded rank first, lowest free slot, FIFO queue beyond 64), and the executor runs one of two batch buckets per step — 1 row/rank while every rank holds ≤ 1 request, the full 8-row batch as soon as any rank holds two. Four whole-step CUDA graphs (2 attention tiers × 2 buckets), lazily captured; the 1-row bucket addresses whichever slot holds the request through a prologue-rewritten block-table row (device data, never a baked slot id). Zero CUDA changes — the weight-only GEMV already supports exactly rows ∈ {1, 8}, everything else takes the row count as a launch parameter.

The admission/step-shape decisions are pure functions over the occupancy (`admission_target` / `plan_step_shapes`, unit-tested); the bucket is a single coordinator decision per step, so the DeepEP collectives always see an agreed global row count (8 vs 64).

## Measured (jz-38 8×H200)

- **Solo decode: 39.2 → 22.4 ms/step** (D1's fixed 8-row batch paid 7 pad rows every step; the PR5c 1-row record was 22.5). ×5 runs dead stable.
- Diverse-prompt closed-loop scaling (non-streaming, `tools`-side harness):

| concurrency | ms/step p50 / p99 | aggregate out tok/s |
|---|---|---|
| 1 | 21.5 / 21.6 | 42 |
| 8 | 28.1 / 28.5 | 256 |
| 16 | 51.1 / 52.3 | 280 |
| 32 | 57.1 / 57.7 | 502 |
| 64 | 62.8 / 63.6 | 911 |

- Poisson soaks (random output 32–224): rate 4 → 600/600 ok, p99 e2e 14.2 s; rate 8 (mild overload) → 600/600 ok, p99 27.0 s.
- Known cliff: c16 ≈ c8 throughput (2 real rows/rank pay the 8-row step) — a middle bucket is the follow-up, to be justified by its own A/B.
- Identical-prompt 64-way reads 1586 tok/s vs 911 diverse (+74%) — degenerate MoE routing; the doc records this as a measurement rule.

## Gates (all green, `docs/models/glm52/continuous-batching.md` for the full table)

- 4 oracle gates, with the EP8 layer gate now replayed at BOTH global-token buckets (64 and 8).
- Solo byte-identical to the PR5c v7 AND D1 records; determinism ×2; tier crossing.
- 8/9/64-way concurrency (all outputs byte-identical to solo), 80-way queueing, mixed tiers/lengths, disconnect, SIGTERM at bucket 8, post-traffic drain.
- **Pinned slot-3/slot-7 parity** (the D1 review's acceptance item): hard-coded admission pin, both the 1-row and forced 8-row buckets, byte-identical to slot-0 refs.

## Known numerics property (root-caused, documented, not a bug)

Batch-1 and batch-8 kernel paths are different FP associations (cuBLAS n=1 vs n=8 kernel selection, FlashMLA SM-part split partitioning), so a request whose lifetime spans a bucket switch can greedy-diverge from its solo replay at a near-tie (measured: first flip at output token 215 on the probe prompt; 9-way runs are mutually byte-identical, 8-way bucket-1 runs match solo exactly). Standard batch-variant numerics, same as every batching engine; determinism holds for a fixed occupancy timeline. Discriminating experiments and the analysis are in the doc.

Toxic-review pass: Approve; both hard findings addressed (shape-indexed graph/plan/scratch arrays kill the 6-site dispatch, `plan_step_shapes` encodes the bucket/slot invariant in one function) plus the queue dead-request skip and the b1 oracle coverage.


xiaguan:
Filled the scaling curve between the bucket boundaries (same non-streaming diverse-prompt harness):

| concurrency | ms/step p50 / p99 | aggregate out tok/s |
|---|---|---|
| 1 | 21.5 / 21.6 | 42 |
| 2 | 24.1 / 24.3 | 75 |
| 4 | 26.6 / 26.8 | 135 |
| 7 | 29.3 / 29.7 | 215 |
| 8 | 28.1 / 28.5 | 256 |
| **9** | **47.1 / 49.2** | **171** |
| 12 | 49.0 / 49.5 | 220 |
| 16 | 51.1 / 52.3 | 280 |
| 32 | 57.1 / 57.7 | 502 |
| 64 | 62.8 / 63.6 | 911 |

Two facts this pins down:

- The 1-row-bucket region (c1–c8) climbs gently (21.5 → 29.3 ms/step) — more active ranks means more real dispatch in the collectives, no cliff.
- The bucket cliff is **non-monotonic**: the 9th concurrent request drops *total* throughput below c7 (171 vs 215 tok/s), and break-even with c8 is ≈ c14 — the whole fleet pays the 64-row step for 9 real rows. This is the measured case for a middle bucket (follow-up, needs its own A/B: 2 more graphs + a batch-4 GEMV instantiation + one more FP-association regime).

Doc table updated in `continuous-batching.md`.

PR #553, opened 2026-07-04
feat(glm52): DSpark speculative-decoding design + span steps (D3 M1)

xiaguan:
## What

First slice of the DSpark speculative-decoding campaign (D3), two parts:

1. **Design doc** (`docs/models/glm52/dspark-mtp.md`): greedy speculative decoding for GLM5.2 using the community `RedHatAI/GLM-5.2-speculator.dspark` checkpoint (not the native MTP layer). Checkpoint facts verified on the dev node: qwen3-architecture 5-layer draft at hidden 6144, `block_size=8` → 7 drafts, verify span 8 = one decode-bucket-8 step; draft `embed_tokens`/`lm_head` sha256-identical to the target's (skipped at load, ~3.8 GB per rank). Anchor-drop layout, bidirectional intra-block attention, and the Markov loop semantics pinned against the `vllm-project/speculators` source.

2. **M1 — span steps**: a step's rows may now repeat a slot as one contiguous run of consecutive positions, so mid-prefill requests batch up to a bucket of prompt tokens through one whole-step-graph decode step (previously strictly one position per step). This is both an immediate prefill win and the exact row shape a DSpark verify step feeds in M3.

## How

- `decode_step`: per-ROW inputs/outputs; the slot-dedup + full-bucket identity asserts become a per-slot contiguous-consecutive-run check; every bucket's block table is dtod-gathered per step (same buffer the captured graphs already read — pointer-stable, graph-safe). Within-step KV causality is structural: per layer every row's cache write launches before any row's attention, and row k's `seq_lens` admits exactly the positions before it.
- `plan_step_shapes`: bool occupancy → per-slot feed wants; every active slot gets a liveness row, leftover bucket capacity extends prefill spans **round-robin** across hungry slots (toxic-review finding: ascending-slot greed starved a co-resident prefill to one row/step).
- `Glm52SlotState`: `feed_want` / `next_input_at` / `advance_span` (mid-prompt row outputs discarded; the boundary row yields the first generated token).
- Drive-by: fix pre-existing `cfg(test)` import drift (`Glm52ProjBytes`) that broke `cargo test -p openinfer-glm52 --lib`.

## Measured (jz-38, 8×H200, 1621-token prompt, ×3 stable)

| metric | main (token-by-token) | M1 (span-8) |
| --- | --- | --- |
| TTFT | 35.2–36.6 s | **9.25 s (3.9×)** |
| prefill steps | 1621 | 203 |
| solo decode | 22.4–22.5 ms/step | 22.5–22.6 ms/step (no regression) |
| bucket-8 one-real-slot step | — | **45.6 ms** (the M3 round-math input) |

## Gates (all green on jz-38)

- 1621-token prompt output **byte-identical** span-path vs main's token-by-token walk (+ determinism ×2); short-prompt outputs identical to main.
- 9-way mutual byte-identity + == solo; queue-80; drain; SIGTERM teardown.
- Mixed: long prefill admitted during 8-way decode completes full length (text divergence vs solo = the known cross-shape FP association property from D2).
- Round-robin path: 10 concurrent 1621-token prefills (≥2 ranks hosting two prefills each) 10/10 full length.
- 17/17 lib unit tests; toxic review: approve, no blockers (fairness finding fixed in `dd9dbdc`).

Known non-M1 quirk surfaced by the gate prompt: a highly repetitive prompt makes GLM emit EOS as its first token, and the frontend then reports `completion_tokens=0` with the prompt echoed in `text` — identical on main, frontend accounting, tracked separately.

Next: M2 (draft lane: backbone forward + Markov propose + aux-hidden capture), then M3 (verify/accept round loop + c1 A/B).


xiaguan:
## M2: draft lane (shadow mode) — jz-38 gate green

The DSpark drafter now loads on every rank behind `--dflash-draft-model-path` in **shadow mode**: the decode path is unchanged; per round the coordinator feeds captured aux hidden to the draft, proposes 7 tokens, and scores them against what plain decode actually emits. One number validates the capture layout, the backbone forward, and the Markov head together.

### Measured (jz-38, 8×H200, temperature 0)

| prompt | rounds | mean accepted (incl. bonus) |
|---|---|---|
| code-gen (CSV parser) | 59 | **3.44** |
| technical prose (B-tree) | 70 | 2.89 |
| narrative prose | 82 | 2.40 |
| predictable counting | 43 | 4.74 |

Checkpoint's own val: 3.967. Zero-accept fraction on the code prompt (17%) matches the published first-position accept of 0.83. A capture-layout bug would collapse all of these to ≈1.x.

- **Output parity**: all prompts byte-identical with the drafter on vs off (shadow perturbs nothing; in-graph capture copies change no results).
- **Cost**: solo decode 22.50–22.56 ms/step with capture copies (baseline 22.4–22.6 — free); 23.32–23.44 with the shadow drafter ≈ **2.2 ms per draft round**, inside the 1–3 ms the round-cadence math assumed.

### Two pins that prevented silent wrongness

1. **vLLM aux-capture ids are "before layer idx" semantics** (pinned from vllm `deepseek_v2.py` + speculators `launch_vllm.py`): the checkpoint's `[8,23,39,55,70]` map to our post-layer capture points `{7,22,38,54,69}`. The qwen3 DFlash checkpoint (DeepSpec-trained) uses the opposite convention — copying its wiring would have been off by one.
2. **head_dim 64**: both draft attention kernels were hard-wired to qwen3's head_dim 128. The dflash qk-norm-rope kernel's cross-warp reduction also read all 4 `warp_sums` slots unconditionally (uninitialized shared memory at 2 warps) — fixed by gating on the actual warp count; FlashInfer prefill got a HEAD_DIM=64 instantiation. The crash-early launcher checks caught this on the first draft round.

M3 (verify/commit round loop + span-4 vs span-8 A/B + c1 A/B per the qwen3 methodology) is next; the measured accepts say span 8 is 1.6× on code but only 1.1–1.3× on prose, so the bucket-4 verify step time is the first thing to measure.

xiaguan:
## M3: verify/commit rounds — speculative decoding is live (jz-38 gates green)

The shadow scaffolding became the real loop: a decoding slot's step span IS the verify span `[anchor, d1..dk]`; `advance_span` runs `accept_greedy` (ported verbatim from `openinfer-qwen3/src/speculative.rs`) over the span's per-row argmax and commits the accepted prefix + correction/bonus. Rejected rows need no KV rollback — `seq_lens` caps reads and the next span overwrites (paged-overwrite semantics). With the drafter off, every span is one row and behavior is unchanged (the plain path is the zero-draft special case of the same code).

### Span A/B: span 4 wins everywhere → `GLM52_DSPARK_SPAN_DRAFTS = 3`

c1, greedy, 200-token completions + 512-token solo ×3 (jz-38, 8×H200):

| | plain | spec span-8 | spec span-4 |
|---|---|---|---|
| solo ms/token | 22.5 | 19.4 (1.16×) | **15.7 (1.43×)** |
| code tok/s | 41.3 | 62.5 | **70.1 (1.70×)** |
| tech prose | 43.2 | 36.5 ⚠️ *slower than plain* | **68.6 (1.59×)** |
| narrative | 43.6 | 33.6 ⚠️ | **59.7 (1.37×)** |
| counting | 43.4 | 95.1 | **97.7 (2.25×)** |
| **sharegpt ×30 aggregate** | **40.2** | — | **61.1 (1.52×)** |

Per-round wall matches the step-time model exactly: span-8 ≈ 46.6 ms/round (M1's measured 45.6 ms bucket-8 step + ~2 ms draft), span-4 ≈ 33.9 ms/round → the bucket-4 verify step is ~32 ms. The cheaper round beats span 8's longer accepted tail on every prompt class; span 8 is strictly dominated and actively loses to plain decode on low-accept prose.

### Closing A/B (qwen3 dspark methodology, user-specified)

30 sampled ShareGPT first-turn prompts, c1 sequential, temperature 0 + ignore_eos: **40.2 → 61.1 tok/s aggregate (1.52×)**, p50 per-request 42.7 → 66.3 (1.55×). Accept incl. bonus 2.489 over 2395 rounds; 32% of rounds accept all 3 fed drafts (span headroom → adaptive span is a future lever).

### Correctness

- 18/30 sharegpt outputs (and 3/5 gate prompts) byte-identical to plain decode; the divergent ones split at a near-tie hundreds of tokens in — the known cross-bucket-shape FP association property (same as D2's diagnosis). Speculation itself is deterministic: the same prompt twice produces identical output.
- EOS inside a committed run truncates and finishes (suppressed-EOS accounting preserved); the max_tokens budget caps the verify span, which also keeps every fed position under the 4096 context cap for free.
- M2 review's H1 addressed: all 8 per-slot draft states (~350 MB each) preallocate at load — a mid-serving draft round never touches the allocator.

PR #540, opened 2026-07-03
feat(qwen3): vLLM-prefill P/D — cross-engine KV compat, verified end-to-end

xiaguan:
## What

Qwen3 decode nodes can now consume KV prefilled by a **vLLM + PegaKVConnector** peer over the pegaflow P2P mesh: vLLM prefills and seals blocks under its own prefix-cache hashes; openinfer-D derives byte-identical keys, discovers the blocks via the MetaServer, RDMA-fetches and restores them, and decodes — zero vLLM/connector changes.

- **`openinfer-kv-offload/src/vllm_hash.rs`** — `VllmBlockHasher`: replicates vLLM's `xxhash_cbor` block-hash chain (hand-rolled canonical CBOR + xxh3_128, `NONE_HASH` from `PYTHONHASHSEED`). Golden vectors captured from a real vLLM process.
- **D-side wait window** — a cold request whose query misses re-queries (5ms/req throttle) until the producer's registration lands (`--kv-pd-miss-wait-ms`, default 5000), bridging the P→D save/publish race. `remote_fetch.rs` decision cascade extended with `wait_on_miss`; 12 unit tests cover the branch matrix.
- **Unhappy path is loud and bounded** (toxic-review hardening): startup fingerprint log (seed/namespace/block_size/`NONE_HASH`/KV geometry — diff it against the P side), warn on every exhausted miss window, and a miss breaker — after 3 consecutive exhausted windows (or 15s hard timeouts) new requests skip the wait until a remote hit re-arms it. Flags validate at parse time (seed must be decimal u32, namespace 8-hex); `miss-wait >= 15s` deadline is rejected at startup.
- **Server flags** — `--kv-pd-vllm-seed`, `--kv-pd-vllm-namespace`, `--kv-pd-miss-wait-ms` (all gated on the P2P mesh flags).

## Compatibility contract (verified on hardware)

[...]

PR #554, opened 2026-07-04
feat(qwen3): make KV page size configurable

### LINKED ISSUE #545: Make the qwen3 KV page size configurable (currently hardcoded to 16)
`openinfer-qwen3/src/weights.rs` pins the paged-KV page size to a constant:

```rust
fn kv_budget_geometry(&self) -> KvBudget {
    let page_size = 16;
    ...
}
```

In vLLM-compat P/D mode (#540) the block hash granularity, seal granularity, and the 1:1 remote-block-to-GPU-page load mapping all derive from `budget.block_size`, so the vLLM prefill peer must run `--block-size 16` to match. That works, but the alignment could equally go the other way if the page size were configurable.

## Proposal

Lift the constant into a config knob (executor option and/or server flag), keeping 16 as the default.

Trade-offs to evaluate before flipping any default:

- Larger pages coarsen prefix-cache reuse (up to `page_size - 1` wasted tokens per boundary in multi-turn) and increase internal fragmentation.
- FlashInfer supports both 16 and 64, but accuracy gates and perf baselines need a re-run per page size.
- 64-token pages quadruple the per-(layer, block) transfer segment (64 KiB → 256 KiB), reducing small-message overhead on the P2P path — not a current bottleneck (16-token blocks measured at 23.7 GiB/s), but relevant if RDMA message rate ever becomes one.

Not urgent: GLM (P/D milestone 2) natively uses 64-token pages, so the granularity mismatch is qwen3-specific.

sparkzky:
## What

Replace the hardcoded `page_size = 16` in Qwen3's `kv_budget_geometry()` with a configurable knob threaded through `Qwen3MemoryOptions`. The new `--kv-page-size` CLI flag (default 16, only 16/64 valid per FlashInfer constraint) flows end-to-end: CLI → config → `Qwen3MemoryOptions` → `kv_budget_geometry()` → `KvBudget.block_size` → `KvCacheManager` → `PagedKvLayout` → FlashInfer kernel.

Closes #545.

## Why

In vLLM-compat P/D mode (#540) the block hash granularity, seal granularity, and the 1:1 remote-block-to-GPU-page load mapping all derive from `budget.block_size`. Lifting the hardcoded 16 into a config knob allows the vLLM prefill peer to align to either granularity — and GLM natively uses 64-token pages.

## Changes

- **`weights.rs`**: `DEFAULT_KV_PAGE_SIZE=16` + `VALID_KV_PAGE_SIZES=[16,64]` constants; `page_size` field on `Qwen3MemoryOptions` with `validate()` enforcing the 16/64 constraint; `kv_budget_geometry(&self, page_size)` now parametric; 3 unit tests (valid/invalid/default).
- **`config.rs` + `main.rs`**: `--kv-page-size` CLI flag (default 16), threaded into `Qwen3MemoryOptions`.
- **`dynamo-backend`**: mirrored `--kv-page-size` arg + `DEFAULT_KV_PAGE_SIZE` import.
- **dflash tests**: updated `Qwen3MemoryOptions::new` call sites for the new 3rd param.

## Verification

- `cargo build --release` — clean (0 warnings)
- 3 unit tests (`page_size` suite) — all pass
- **page_size=16 (default)**: 362 blocks (814 MB), greedy output correct — `" Paris. The capital of Germany is Berlin. The"`
- **page_size=64**: accepted by CLI, passes validation; OOM on 12GB RTX 3060 during profiling (each block is 4x larger → higher temp alloc floor). Expected to work on ≥16GB cards.

## Trade-offs (from #545)

- Larger pages coarsen prefix-cache reuse (up to `page_size - 1` wasted tokens per boundary).
- 64-token pages quadruple per-(layer, block) transfer segment (64 KiB → 256 KiB).
- FlashInfer supports both 16 and 64; accuracy gates and perf baselines need re-run per page size.

### DIFF
--- a/openinfer-dynamo-backend/src/engine.rs
+++ b/openinfer-dynamo-backend/src/engine.rs
@@ -29,11 +29,8 @@ use futures::stream::BoxStream;
 use openinfer_engine::engine::{
     EngineHandle, GenerateRequest, KvBlockEvent, LoadSnapshot, TokenSink, TokenStreamReceiver,
 };
-use openinfer_qwen3_4b::{
+use openinfer_qwen3::{
     DEFAULT_GPU_MEMORY_UTILIZATION, DEFAULT_KV_CACHE_MEMORY_MARGIN_BYTES, DEFAULT_KV_PAGE_SIZE,
-    DEFAULT_MAX_PREFILL_TOKENS, DecodeOverlap, Qwen3LaunchOptions, Qwen3MemoryOptions,
-    Qwen3OffloadOptions,
-};
     DEFAULT_MAX_PREFILL_TOKENS, DecodeOverlap, Qwen3LaunchOptions, Qwen3MemoryOptions,
     Qwen3OffloadOptions,
 };
--- a/openinfer-dynamo-backend/src/engine.rs
+++ b/openinfer-dynamo-backend/src/engine.rs
@@ -29,8 +29,11 @@ use futures::stream::BoxStream;
 use openinfer_engine::engine::{
     EngineHandle, GenerateRequest, KvBlockEvent, LoadSnapshot, TokenSink, TokenStreamReceiver,
 };
-use openinfer_qwen3::{
-    DEFAULT_GPU_MEMORY_UTILIZATION, DEFAULT_KV_CACHE_MEMORY_MARGIN_BYTES,
+use openinfer_qwen3_4b::{
+    DEFAULT_GPU_MEMORY_UTILIZATION, DEFAULT_KV_CACHE_MEMORY_MARGIN_BYTES, DEFAULT_KV_PAGE_SIZE,
+    DEFAULT_MAX_PREFILL_TOKENS, DecodeOverlap, Qwen3LaunchOptions, Qwen3MemoryOptions,
+    Qwen3OffloadOptions,
+};
     DEFAULT_MAX_PREFILL_TOKENS, DecodeOverlap, Qwen3LaunchOptions, Qwen3MemoryOptions,
     Qwen3OffloadOptions,
 };
@@ -71,6 +74,9 @@ struct Args {
     /// Fraction of GPU memory the engine may use (weights + KV cache).
     #[arg(long, default_value_t = DEFAULT_GPU_MEMORY_UTILIZATION)]
     gpu_memory_utilization: f64,
+    /// KV cache page (block) size in tokens. FlashInfer only accepts 16 or 64.
+    #[arg(long, default_value_t = DEFAULT_KV_PAGE_SIZE)]
+    kv_page_size: usize,
 }
 
 pub struct OpeninferBackend {
@@ -112,6 +118,7 @@ impl OpeninferBackend {
         let memory = Qwen3MemoryOptions::new(
             args.gpu_memory_utilization,
             DEFAULT_KV_CACHE_MEMORY_MARGIN_BYTES,
+            args.kv_page_size,
         )
         .validate()
         .map_err(|e| convert::invalid_argument(format!("invalid memory options: {e:#}")))?;
--- a/openinfer-qwen3/src/lib.rs
+++ b/openinfer-qwen3/src/lib.rs
@@ -28,7 +28,8 @@ use openinfer_core::engine::{EngineHandle, EngineLoadOptions, EpBackend, ModelIn
 pub use kernel_plan::kernel_plan;
 pub use scheduler::DEFAULT_MAX_PREFILL_TOKENS;
 pub use weights::{
-    DEFAULT_GPU_MEMORY_UTILIZATION, DEFAULT_KV_CACHE_MEMORY_MARGIN_BYTES, Qwen3MemoryOptions,
+    DEFAULT_GPU_MEMORY_UTILIZATION, DEFAULT_KV_CACHE_MEMORY_MARGIN_BYTES, DEFAULT_KV_PAGE_SIZE,
+    Qwen3MemoryOptions,
 };
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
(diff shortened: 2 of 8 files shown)


xiaguan:
@FeathBow could u help review this? thks

FeathBow:
Thanks for picking up #545 — the direction is sound. What needs work is the state of the branch.

**Blocking — the pushed tree doesn't build** (it's already on post-#495 main, so the hunks need fixing, not a rebase):

- `openinfer_qwen3_4b::` references in `engine.rs`, both dflash tests, and `config.rs` — that crate is `openinfer-qwen3` since #495.
- `kv_page_size` is gated on `#[cfg(feature = "qwen3-4b")]` while the server feature is `qwen3`, so the field is compiled out and `main.rs:220` reads it → the default `cargo build --release` fails with "no field `kv_page_size`".
- Three malformed import blocks — a dangling `};` at `engine.rs:39`, `dflash_speculative_gate.rs:51`, `dflash_speculative_perf.rs:27` (`cargo fmt --all --check` catches these).
- `Qwen3MemoryOptions::new` at `tp_concurrent_decode.rs:59` still passes 2 args — the last call site; your `Default` update already covers the `::default()` ones.
- The flashinfer submodule pointer moved 57ba7eeb → d768c14e — unrelated to this PR and a downgrade; please restore it: `git checkout origin/main -- openinfer-kernels/third_party/flashinfer`.

**The feature itself works:** I applied just the `kv_budget_geometry` 16→64 change on top of `main` and ran it on a 48 GB card — output is token-identical to the 16-page baseline and the engine reports `KvCapacity { total_blocks: 3823, block_size: 64 }`, so `block_size` flows through correctly and your OOM was a VRAM limit, not a functional problem. (I also checked split-KV chunking — it derives from token counts, not pages, so no batch-invariance interaction.) Still owed once it compiles: hf_golden_gate at `--kv-page-size 64` (my run is a smoke test, not a gate). Perf isn't free either — larger pages coarsen prefix-cache reuse and round each request up to whole blocks (up to 63 wasted tokens), and the decode kernel's page-size sensitivity is unmeasured — so keeping the default at 16 is right; treat 64 as opt-in for #540 until a 16-vs-64 bench exists.

On the OOM specifically: the memory profile gives every profiled decode row a full block, so both the profiling transient and the 64-block minimum floor scale ~4× with page size — those block-count constants were tuned for 16-token pages and are worth revisiting while you're here.

**Minor:**

- The load line `KV cache (profiled): N blocks (M MB, ...)` (weights.rs:1069) doesn't print the page size — please add it so `--kv-page-size 64` is visible (a P/D page-size mismatch otherwise only surfaces as silent prefix-cache misses).
- The #540 rationale (hash/seal/1:1 remote-block mapping deriving from `block_size`) isn't exercised by any test at 64 — a good follow-up once this lands.

FeathBow:
LGTM. It seems I'm unable to trigger CI due to a settings issue, so this hasn't been validated by CI. the formal grant is yours @xiaguan :) 

Also, I reviewed on an sm89 GPU (CUDA 12.9).

**Environment:** sm89 · x86_64 · CUDA 12.9 · Qwen3-4B

Startup log (blocks scale with page size, capacity unchanged):

- page 16 → `15296 blocks (34416 MB, page size 16)`
- page 64 → `3824 blocks (34416 MB, page size 64)` — 4x fewer blocks, same MB

The `KvBudget -> KvCacheManager -> PagedKvLayout -> FlashInfer` chain derives correctly from `block_size`.

## Performance — decode (`--no-prefix-cache`, two-point TPOT)

| Metric | page 16 | page 64 | Δ |
|---|---|---|---|
| single-stream TPOT | 9.84 ms | 9.81 ms | -0.3% |
| conc1 throughput | 100.8 tok/s | 101.0 tok/s | +0.2% |
| conc16 throughput | 1378.6 tok/s | 1360.4 tok/s | -1.3% |
| conc16 latency p50 | 1.490 s | 1.494 s | +0.3% |
| conc16 latency p99 | 1.506 s | 1.509 s | +0.2% |
| conc32 throughput | 2378.3 tok/s | 2384.4 tok/s | +0.3% |
| conc32 latency p99 | 1.710 s | 1.705 s | -0.3% |

All within noise — page size does not move decode.

## Performance — prefix cache (cache on)

Re-send same prompt (warm hit):

| prompt_tokens | cached@16 | cached@64 | warm TTFT@16 | warm TTFT@64 |
|---|---|---|---|---|
| 69 | 64 | 64 | 13.1 ms | 12.9 ms |
| 137 | 128 | 128 | 13.0 ms | 12.7 ms |
| 273 | 272 | 256 | 12.6 ms | 12.5 ms |

Two-turn (A -> R, then A+R+suffix, turn2 = 253 tokens):

| Page size | cached | warm TTFT |
|---|---|---|
| 16 | 240 | 13.6 ms |
| 64 | 192 | 13.6 ms |

The granularity cost is real (up to 48 tokens, the "up to page-1 wasted" effect), but the extra prefill is sub-millisecond, so warm TTFT is unchanged.

## Performance — fragmentation (KV util=0.30 -> 5.3 GB pool, short requests, high concurrency)

| Scenario | page 16 | page 64 |
|---|---|---|
| KV pool | 2364 blocks | 591 blocks (both 37824 token slots) |
| 64 concurrent / 128 reqs | 0 rejections · 142.4 req/s | 0 rejections · 142.3 req/s |
| 128 concurrent / 256 reqs | 0 rejections · 207.4 req/s | 0 rejections · 201.5 req/s (-3%) |

Same total token capacity; normal concurrency never reaches KV pressure, no rejections.

## Verdict

64 is correct, accuracy-lossless, and carries no measurable serving cost on any of the three axes. Keeping 16 as the default with 64 opt-in is the right call.

One line worth adding to the flag help text (not a blocker): the memory profile gives each profiled decode row a full block, so the startup profiling transient scales ~4x with page size — that is why 64 ran out of memory on the smaller card. `--kv-page-size 64` wants a >=16 GB card to start; worth documenting so it does not read as a bug.

xiaguan (approved):
LGTM — CPU checks are green (FeathBow's earlier CI-trigger issue resolved itself once the run went through), and @FeathBow already did the detailed pass. Thanks @sparkzky for closing out #545!

PR #558, opened 2026-07-04
perf(glm52): ROWS=4 register tile for the batched weight-only GEMV — bucket-8 step 45.6 -> 37.0 ms

xiaguan:
## Problem

An nsys attribution of the bucket-8 decode step (one slot spanning 8 rows/rank, e.g. DSpark verify or M1 span prompt ingestion) against a bucket-1 control on the same binary showed the step's +22.9 ms over bucket-1 decomposes as: **+15.1 ms in the non-expert weight-only GEMVs**, +3.8 expert grouped GEMM, +1.8 combine, +1.3 MLA, ~0 dispatch. Two-thirds of the cost sat in the kernel designed to be batch-free.

Root cause (ncu): the 1-row/warp batched GEMV issues 16 activation LDGs per weight LDG at batch 8 and saturates the **L1TEX port** (93% L1/TEX vs 12% DRAM). Shared-memory staging cannot help — LDS shares the L1TEX pipe on Hopper (a microbenched smem variant lost on every shape). Registers are the only storage off that port.

## Fix

Generalize the batched kernel to `<BATCH, ROWS, WARPS>`: a warp owns ROWS=4 output rows and reuses the register-held activation chunk across them, cutting per-row activation loads 4x. Batch 2 keeps the 1-row/warp shape (activation traffic is 2:16 against the weight there and it measures faster on 7 of 10 shapes). ROWS=8 spills; WARPS=4 keeps ~3 blocks/SM at 96 regs.

**Bit parity is preserved by construction** — same lane sweep, per-term FMA order, and warp reduction per row as the m=1 kernel; a memcmp-gated H200 microbench verifies byte-identical outputs for all 10 whitelisted shapes x batches {2,4,8}.

## Measured (jz-38 8xH200)

Microbench (batch 8): o_proj 132→66 µs, dense_gu 171→99, dense_dn 99→51, q_b 42→26, shared_gu 32→19. Weighted over the model's per-step shapes: bucket-8 GEMV 23.0→13.9 ms, bucket-4 13.4→9.5 ms.

e2e (dspark stack + cherry-pick, c1 greedy):

| probe | before | after |
|---|---|---|
| bucket-1 solo | 22.5–22.6 ms/step | 22.63 ms/step (flat — b1 path untouched) |
| bucket-8 span step | 45.6 ms | **37.0 ms (−19%)** |
| 1621-token prompt TTFT | 9.25 s | **7.55 s** |

Also benefits bucket-4 (DSpark span-4 verify) and c33+ concurrent decode; not re-measured here.

Attribution record: `docs/models/glm52/whole-step-decode-graph.md` (bucket-8 step attribution section).

### DIFF
--- a/docs/models/glm52/whole-step-decode-graph.md
+++ b/docs/models/glm52/whole-step-decode-graph.md
@@ -97,6 +97,47 @@ Measured dead ends (all recorded in #542): the inter-step host gap is ~0.05 ms (
 
 The same-node vLLM nsys diff (their trace: NCCL AllGather+ReduceScatter naive EP — NOT DeepEP — at 29+36 µs/layer medians, comparable to our 12+47) relocates the remaining 2.6 ms: (1) **expert GEMM** — vLLM's DeepGEMM swapAB runs ~10.8 µs/instance where our TRTLLM grouped takes 22.7 µs (64-row M-tile against ~8 real rows = 8× wasted MMA work; swapAB puts the skinny dimension on N). (2) **collective wait structure** (dispatch is rank-arrival-bound; fp8 payload measured neutral). A calibration lesson from the tier stage: node-granularity nsys inflates small kernels — the attention diff read as ~1 ms in trace proportions but bought 0.3 ms e2e; size expected wins from e2e A/B, not trace ratios.
 
+## Bucket-8 step attribution (jz-38 nsys 2026-07-05)
+
+Why does a bucket-8 step cost 45.6 ms when bucket-1 costs 22.5? Same methodology
+as above (node-granularity trace), two windows on the same binary: bucket-8 via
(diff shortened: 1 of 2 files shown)


xiaguan:
Toxic review: **Approve** — zero correctness defects. Independently verified: indexing/grid exact for all 10 whitelisted shapes (all n%16==0), scale-row sharing valid incl. kv_a n=576 (max n0=572 → scale row 4 of 5, ROWS=4 groups never straddle /128), batch-2 launch geometry byte-for-byte unchanged, ptxas recompile 158/128/40 regs 0 spills, no new alignment assumptions. Two comment corrections requested (stale reg count, stale batched-rpb note) — fixed in the follow-up commit. Non-blocking follow-up recorded: extract a shared `dot16` device helper so the per-term FMA order exists once instead of twice.

Issue #559, opened 2026-07-04
glm52: bucket-8 decode step residual cost — 37 ms vs 22.5 bucket-1 after the GEMV tile fix


xiaguan:
After #558 (batched GEMV ROWS=4 tile) the bucket-8 step is 45.6 -> 37.0 ms measured (jz-38, one-real-slot span steps). The remaining +14.5 ms over bucket-1, from the nsys attribution in `docs/models/glm52/whole-step-decode-graph.md` (bucket-8 step attribution section):

| component | b1 | b8 (pre-fix totals) | lever |
|---|---|---|---|
| batched GEMV | 6.5 | 13.9 (post-fix, weighted) | o_proj still 66 us vs 28.5 weight floor; next: ncu the r4w4 tile, possibly k-chunk ILP |
| expert grouped GEMM | 3.4 | 7.2 | 64-row M-tile padding vs ~2 rows/expert — same swapAB lever as #542 |
| MoE combine | 3.6 | 5.4 | payload-proportional; device-bound spin |
| MLA splitkv | 1.3 | 2.6 | 8 queries/rank; short-tier params tuned for 1 |

Why it matters: bucket-8 is the DSpark span-8 verify step and the M1 long-prompt ingestion step (TTFT), and c33+ concurrent decode pays it too. The span-8-vs-span-4 break-even is step8/step4 <= ~1.15 (accept ratios 1.07-1.17 on real content); currently 37/28 = 1.32, so every ms here directly moves the speculative-decoding frontier (incl. adaptive span).

Acceptance: bucket-8 one-real-slot span step <= 32 ms measured on jz-38 (long-prompt TTFT probe, `~/develop/xingming/glm52_gemv_probe.sh`), or a documented physics floor showing why not.

Artifacts: nsys traces + kern-sum CSVs `~/develop/xingming/b8span* b1ctrl*`, microbench `~/develop/xingming/gemv_bench` (jz-38).

xiaguan:
PR #561 (tensor-core mma path for the batch-4/8 GEMV, stacked on #560) hits the acceptance line: **solo span-8 ingest step 36.9 → 32.2 ms** on jz-38 (same-day A/B via the new in-process `glm52_step_bench`), c64 diverse decode 1046 → 1121 tok/s, buckets 1/2 untouched.

[...]

PR #561, opened 2026-07-04
perf(glm52): tensor-core mma path for batch-4/8 weight-only GEMV — solo span-8 step 36.9 → 32.2 ms

xiaguan:
Stacked on #560 (step bench). Closes the main item of #559.

## Problem

After #558 the bucket-8 step's largest term was still the batched weight-only GEMV (+6.8 ms over bucket-1, post-fix nsys attribution in `docs/models/glm52/whole-step-decode-graph.md`). ncu showed the wall had moved: L1TEX 46% (fixed by #558), DRAM 24%, **Compute 65%** — the `BATCH × 16`-term f32 FMA chain itself. The computed CUDA-core floor for o_proj at batch 8 (~32 µs) sits above the 28.5 µs weight-read floor, and occupancy/cache-policy variants measured flat or worse: the CUDA-core design space was exhausted.

## Fix

`mma.m16n8k16.bf16` retires the FMA chain on the tensor cores (fp8 e4m3 decodes losslessly to bf16). Two design points:

- **No weight repack.** The mma k-slot→column map is a free permutation when A and B agree, so σ(step s, tid, d) = tid·16+4s+d over a k64 super-chunk reads the *original* row-major layout with one 16-byte LDG per owned row per chunk — no second weight copy (a packed fragment layout is ~30% faster on the biggest shapes but needs +15 GB/rank or a bucket-1 numerics change; recorded as a non-goal in the doc).
- **KSPLIT k-slices** (the occupancy lever the bit-parity kernels could never use) write f32 partials to a 12.6 MB per-device scratch — lazily allocated by the existing pre-capture bucket warm-up, so graph capture sees a stable pointer — and a fixed-order epilogue reduces them.

Dispatch is per-shape (`mma_config`): batch 8 routes every whitelisted shape to mma; batch 4 only the k=6144 column (elsewhere the register tile stays faster). Batch 1/2 paths are byte-untouched.

## Numerics contract

Deterministic and replay-stable per bucket, **not** bit-identical to m=1 (mma fixes its own f32 accumulation order). Cross-bucket FP divergence is the already-accepted whole-step contract (the expert GEMM reassociates across buckets since D2). Validation:
- microbench tolerance vs the bit-exact m=1 reference: max_rel < 2e-2 and ≥90% of bf16 outputs exactly equal, all 10 whitelisted shapes × batch {4,8};
- e2e greedy coherence on code/prose/long prompts (jz-38);
- dspark accept-rate parity (verify spans ride bucket 4): code 1.80, prose 0.99 mean accepted drafts — in line with the known per-class distribution.

## Measured (jz-38 8×H200, `glm52_step_bench`, same day A/B)

| shape | before | after |
|---|---|---|
| solo span-8 ingest (spec-verify / TTFT shape) | 36.9 ms/step | **32.2 ms/step (−12.6%)** |
| c64 diverse decode (bucket 8) | 61.2 ms p50, 1046 tok/s | **57.1 ms p50, 1121 tok/s** |
| c32 (bucket 4) | 48.4 | 47.9 |
| c16 (bucket 2) / c8 (bucket 1) | 39.3 / 30.4 | 38.6 / 30.4 (untouched) |

Microbench per shape (batch 8, µs): o_proj 66.5→45.8, dense_gu 99.4→58.6, dense_dn 50.9→33.0, q_b 26.2→16.9, q_a 16.4→10.7, kv_a 13.2→7.7, shared_gu 19.3→13.6, shared_dn 11.5→9.3 (~−66 µs per MoE layer).

Also probed the "NV library" route (vendored TRT-LLM dense fp8 `gemm()`, w8a8): it is a no-op in our AOT setup — the SM90 dense path expects the DeepGEMM JIT runtime. cuBLASLt does not support fp8-weight × bf16-act mixed input.

### DIFF
--- a/docs/models/glm52/whole-step-decode-graph.md
+++ b/docs/models/glm52/whole-step-decode-graph.md
@@ -138,6 +138,48 @@ untouched.** Matches the microbench-weighted −9.1 ms prediction. Artifacts:
 traces + kern-sum CSVs `~/develop/xingming/b8span* b1ctrl*`, microbench
 `~/develop/xingming/gemv_bench` on jz-38.
 
+## Tensor-core GEMV for batches 4/8 (jz-38, 2026-07-05, #559)
+
+Post-#558 re-attribution (same two-window method, fixed binary): the b8−b1
+delta fell +22.9 → +15.2 ms and the GEMV was still the largest term:
+
+| component (per rank-step, totals) | b1 | b8 post-#558 | Δ |
+|---|---|---|---|
+| weight-only GEMV | 6.4 | 13.2 | **+6.8** |
+| expert grouped GEMM | 3.4 | 7.2 | +3.8 |
+| MoE combine | 4.3 | 6.2 | +1.9 |
+| glue | 5.8 | 7.0 | +1.2 |
+| MLA splitkv | 2.1 | 3.3 | +1.2 |
+| dispatch (incl wait) | 5.3 | 5.7 | +0.4 |
+
+ncu on the ROWS=4 tile showed the wall had MOVED: L1TEX 46% (fixed), DRAM 24%,
+Compute 65% — the BATCH×16-term f32 FMA chain itself. Computed CUDA-core
+floor for o_proj at batch 8 ≈ 32 µs vs the 28.5 µs weight-read floor;
+occupancy and cache-policy variants (KernelWiki NVFP4-GEMV levers) measured
+flat or worse. Conclusion: the CUDA-core design space was exhausted — the fix
+is `mma.m16n8k16.bf16` (fp8 e4m3 decodes losslessly to bf16).
+
+Key trick: no weight repack. The mma k-slot→column map is a free permutation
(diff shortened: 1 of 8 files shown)


xiaguan:
Toxic review found a real blocker in the first cut: the lazy per-device scratch was shared between the ctx and aux streams, which the layer forward deliberately overlaps (indexer wq_b/wk on aux vs q_b/kv_a on ctx) — a device-side race baked into every graph replay. Fixed in a19ff3f: the scratch is now caller-owned, one buffer per GEMV-using scratch struct (MlaFront / MlaAttend / Indexer / both Mlp instances), threaded through the FFI; an mma-routed launch without a buffer fails INVALID_VALUE. mma_config is also now an explicit per-shape table (unmeasured future shapes fall back to the register tile).

Re-measured after the fix (jz-38): solo span-8 ingest 32.36/32.44 ms (unchanged), c64 57.46 ms p50 / 1114 tok/s, b1/b2 flat; e2e code/prose/long outputs and dspark accept rates byte-identical to the pre-fix run.

Issue #284, opened 2026-06-06
qwen3/qwen35: migrate sampling to the batched FlashInfer primitive


xiaguan:
## Problem

The kimi TP1/DP8 sampling work (#237) added a batched sampler (`gpu_sample_batch_into`, `pegainfer-kernels/src/ops/sampling.rs`). The qwen crates still use the legacy single-row path, which has both perf and correctness defects:

**Per-row launches + syncs** — batch decode samples one row at a time:
- `pegainfer-qwen3-4b/src/executor.rs:165` — `for (i, req) in requests` → `sample_from_logits` per row
- `pegainfer-qwen35-4b/src/batch_decode.rs:24` — same loop, plus a per-row `ops::extract_vec` (device alloc + D2D copy per row per step)

Each call is O(1 row): kernels + D2H + `ctx.sync()` → O(batch) syncs per decode step.

**Legacy wrapper defects** (`gpu_sample_flashinfer_cuda`, `pegainfer-kernels/csrc/shared/flashinfer_sampling.cu:125`):
- Softmax is a custom `<<<1, 256>>>` single-block kernel over the 151K vocab — 1 SM busy on a 132-SM GPU. The batched path uses FlashInfer's `OnlineSoftmax` vocab-splitting kernel instead.
- `deterministic=false` (vLLM passes true; batched path passes true).
- Hard-coded 1MB FlashInfer row-state scratch (`FLASHINFER_TOPK_ROW_STATES_BYTES`) — brittle dependency on upstream internal layout.
- Seed is `u64::from(random_val.to_bits())` — a host-drawn f32 used only as seed material, ~30-bit effective space.

## Fix

Adopt the kimi structure: greedy rows keep the batched argmax; non-greedy rows go through one `gpu_sample_batch_into` call per step (3 kernels + 1 sync for the whole batch, per-row temperature/top_k/top_p, fresh u64 seed per step). Then delete the single-row wrapper and its custom softmax.

## Acceptance

- Batch decode with mixed greedy/sampled rows issues one sampling pass per step (verify with nsys: no per-row `logits_to_probs_kernel`).
- `gpu_sample_flashinfer_cuda` and the 1MB row-state scratch are gone.
- hf_golden_gate stays green for both models; greedy TPOT unchanged.

Issue #548, opened 2026-07-04
vllm bench serve (0.24) against GLM5.2 serving: client hang at concurrency 8, distorted streaming latencies


xiaguan:
Observed while running the D2 (#547) soak against `target/release/openinfer --model-path <GLM5.2> ` on jz-38, client = `vllm bench serve` from vllm 0.24 (`--backend openai --endpoint /v1/completions --dataset-name random --random-input-len 1 --random-output-len 128 --ignore-eos --temperature 0.0`).

Three independent observations:

1. **Client hang at `--max-concurrency 8`** (`--num-prompts 160`): the server log shows 160/160 requests started AND finished (balanced, no ERROR/WARN), `ss -tnp` shows all client↔server sockets ESTABLISHED with Send-Q/Recv-Q = 0 on both sides (drained keep-alive pool), but the bench process never prints results — main thread parked in `epoll_wait`, 390 threads alive. Concurrency 1 (40 prompts) completes and prints normally.
2. **Streaming latencies are distorted**: at concurrency 1 the bench reports ~39 ms/token E2E, while a streamed `curl` of the same 128-token request against the same server measures 23.0 ms/step (non-streamed 22.5). So the server SSE transport is fine; the inflation is on the client side. An initial TCP_NODELAY/Nagle suspicion was refuted by the curl probe.
3. **Default temperature**: without `--temperature 0.0` the engine fast-rejects (greedy-only bring-up) but the bench counts the rejected requests as successful 1-token completions — the rejection surfaces as an in-stream error event, not an HTTP 4xx (that surface is #294 / upstream vllm#45286).

Repro: `bash /root/develop/xingming/glm52_d2_bench.sh` on jz-38 (the c8 sweep hangs); engine-side numbers were collected with the non-streaming harness `d2_engine_bench.py` instead.

Acceptance: root-cause the c8 hang (client vs our SSE termination under keep-alive — a packet capture of one hung run should settle it); decide whether anything server-side needs fixing or this is purely a bench-client issue to document in the serving playbook.

xiaguan:
[...]

PR #589, opened 2026-07-06
feat(glm52): sampled-verify speculative decoding — lift the DSpark greedy-only gate

xiaguan:
## What

Non-greedy requests now speculate under the DSpark drafter. The verify span's committed rows are **sampled** (the #586 post-graph FlashInfer pass, extended to every span row) and `accept_prefix_match` compares the sampled tokens against the drafts — the same prefix-match rule as greedy, which for a deterministic draft is **lossless speculative sampling**: every committed token is a true sample from the target distribution; the match only decides how many tokens ride one step.

Core diff: `sampling_row() → sampling_rows()` — a verify span samples all rows, row `k` at request-local step `completion + k` with `mix_seed(request_seed, step)`, the exact step plain decode would sample that token at. A seeded request's speculative stream therefore replays its plain stream token-for-token. The greedy-only admission gate and its assert are removed; greedy behavior is untouched.

## Why prefix-match instead of full rejection sampling

Probe-measured before building (branch `probe/glm52-ab-accept`: raw logits dumps + offline eval, 318 rounds/class at the customer-dominant `(temp 1.0, top_p 0.95)`): full rejection sampling (`min(1, p/q)` + residual resampling) buys **≤ +1.5% throughput on code** over this rule — the greedy draft is a confidence concentrator (the Markov head's distribution is noisier than its argmax), so `Σ min(p,q)` underperforms `p(argmax q)`. Not worth draft sampling, q materialization, a residual kernel, and losing byte-level gates. Details in `docs/models/glm52/dspark-mtp.md` § Sampled verify.

## Measured (jz-38 8×H200, c1, temp 1.0 / top_p 0.95, span-4)

| | plain sampled | spec sampled |
| --- | --- | --- |
| solo code ms/token | 20.72 | **8.71 (2.38×)** |
| code / prose / narrative / counting (320 tok) | 20.7 | 10.8 / 13.2 / 15.6 / 7.6 |
| accept incl. bonus (measured vs probe prediction) | — | 2.55/2.67 · 2.06/2.36 · 1.72/1.72 · 3.67/3.81 |

## Gates (jz-38, 2026-07-06)

- **Greedy neutrality**: drafter-on greedy output on this branch is byte-identical to main (P1/P3, 200 tok) — the diff is a greedy no-op.
- **Seeded determinism**: same seeded request twice under spec → byte-identical.
- **Seeded spec-vs-plain**: P2 tracked the plain stream for 35 consecutive sampled tokens before a near-tie flip (~1e-5 probability under a broken seed stream) — divergences are the known D2 cross-bucket FP property, amplified by sampling's CDF-boundary sensitivity.
- **Accept telemetry matches the shadow-probe predictions per class** (narrative exact: 1.72/1.72).
- **Mixed 9-way** (5 sampled + 4 greedy): all complete, 0 server errors.
- 50 unit tests incl. new cross-round seed-step invariant.

Note: seeded requests over HTTP remain blocked by the shared frontend's seed strip (`wire.rs`, qwen3 legacy path — #284); gates used a reverted gate-only passthrough patch.

Toxic review: Approve-after-fixes; all findings addressed (dead `dspark_enabled` param dropped, `accept_prefix_match` rename, stale greedy `--help` docs, cross-round step test).


xiaguan:
Concurrency A/B added (user ask: real gains under load, small-bs focus). vllm-bench Rust client (the Python `vllm bench serve` c8 hang from #548 did not reproduce — 8/8 runs 100% complete), sharegpt first turns, output 200, temp 1.0 / top_p 0.95, same seed both arms, jz-38 8×H200:

| c | tok/s off → on | TPOT p50 (ms) | TPOT p99 (ms) | TTFT p50 (ms) |
| 1 | 40.6 → 57.7 (**1.42×**) | 20.4 → 12.5 | 20.6 → 18.7 | 321 → 320 |
| 2 | 81.7 → 98.7 (**1.21×**) | 21.9 → 15.2 | 25.4 → 20.0 | 121 → 385 |
| 4 | 130.6 → 163.9 (**1.25×**) | 25.7 → 18.2 | 28.0 → 25.2 | 258 → 435 |
| 8 | 225.3 → 281.3 (**1.25×**) | 29.6 → 19.6 | 34.3 → 30.3 | 230 → 511 |

The win holds through c8 (≤1 request/rank, so verify spans don't contend for slots yet): throughput +21–25%, TPOT p50 1.4–1.6×. Two costs live in the tails: TPOT p99 improves less than p50 (round-granular commits jitter ITL — a rejected round pays a bucket-4 step for one token), and the spec arm's TTFT p50 is ~150–280 ms worse under load (drafter forces prefix caching off + verify spans lift the global bucket during co-resident prefill). c>8 multi-slot contention remains unmeasured. Details in `docs/models/glm52/dspark-mtp.md`.

xiaguan:
Two follow-up measurements closing the span-default question:

**Span-8 re-measured under sampling (c1, identical workload/seed as the table above): still dominated.** 53.8 tok/s / TPOT p50 14.13 ms vs span-4's 57.7 / 12.45. Verify cost is no longer the reason — the bucket-8 solo round is ~33 ms post-D4/D5 (the 45.6 ms M1 anchor is stale), only ~21% over bucket-4's 27.3 ms; the extra 4 draft positions simply buy too little (+0.14 tokens/round on sharegpt at temp 1). Accept histograms still show a real all-7 tail on some requests — that's an adaptive-span opportunity, not a fixed-span one.

**vLLM production reference (GLM5.2-FP8, native MTP `num_speculative_tokens=2`, DP16×TP2+EP; read-only metrics pull):** acceptance length 2.19–2.39 incl. bonus, per-position acceptance 0.80 / 0.45, ~13–17 ms/tok at its typical Running=2–3/engine — parity with this PR's sampled span-4 at matched load (12.45 c1 / 15.16 c2). Notably the draft profiles differ: MTP acceptance is front-loaded (0.80 → 0.45) while dspark's chain is flat (~0.72 → ~0.70), so dspark holds up better at deep positions — adaptive span has more headroom on our side. Their K=2 choice mirrors the span-4-over-span-8 conclusion.

Also a metric gotcha worth recording: vLLM v1 observes ITL once per engine output; a spec round emits ~acceptance-length tokens at once, so ITL dashboards read ≈ round wall while TPOT reads round/acceptance — they differ by the acceptance length by construction.

PR #620, opened 2026-07-08
feat(glm52): attention-TP with replicated activations — solo 13.75ms, MTP code 221 tok/s

xiaguan:
## What

M4 of the TP8 low-latency campaign: extend the TP topology from the MoE to the whole layer.

- **Head-sharded attention weights** ×8 (`q_b`/`kv_b`/`o_proj`/indexer-q; 8 heads/rank, W_UK/UV batch 64→8, o_proj input columns split). FlashMLA is untouched: query/latent keep the 64-wide layout (`splitkv_mla.cuh` asserts `h_q % 64 == 0`) with the 8 real heads in slots 0..8 and zero-query pads.
- **Replicated activations**: every rank redundantly computes ALL bucket rows, bit-identically. The MoE AG phase — and the span/owner concept — is deleted; routing and sampling run redundantly on every rank with zero exchange.
- **Two per-layer partial-sum collectives** are the only wire traffic: a new two-shot LL allreduce for the o_proj partial (`glm52_tp8_ar.cu`, 7.7 µs/layer graph-replayed), and the MoE reduce-scatter turned all-reduce.
- **Mirrored scheduler**: one logical rank, eight mirrored executors, single bucket-8 shape; outputs and draft spans asserted bit-identical at join.
- **Device want-mask** (`active_rows`, staged per step like the epoch; actives are a bucket prefix by plan construction): pad rows are excluded from the expert union (solo union 17→9 — the B/C expert GEMMs nearly halve), skip both LL collectives, and get zero-filled outputs. Zero is the pre-capture shape (push nothing, wait on nothing — capture pairs trivially).

## Measured (jz-38 8×H200, same-day A/B, branch f8677a8 vs main 2669bb6, same client)

| workload | main tp8 | this branch | Δ |
|---|---|---|---|
| solo plain ITL p50 (p99) | 15.27 ms (15.38) | **13.75 ms** (13.85) | **−10%** |
| solo MTP code tok/s ×2 | 186.1 / 188.6 | 197.3 / **220.9** | **+17%** |
| solo MTP prose tok/s ×2 | 108.5 / 108.6 | 118.8 / 119.5 | **+10%** |
| c8 diverse ITL p50 ×2 | 20.56 / 20.58 | 22.65 / 22.68 | +10% (known trade, below) |
| EP8+MTP code (regression) | 131–134 (historical) | 132.4 / 134.4 | untouched |

The c8 regression is structural: the old mapping attended 1 row × 64 heads per rank against its own request's KV; replicated activations attend 8 rows × 8 heads (same head-row product) against all eight requests' KV (8× KV bytes/rank) plus the AR wire. Attention-TP still beats EP8 at c8 (22.7 vs 24.6 ms), and tp8 remains the ≤8-concurrency low-latency mode. Solo per-kernel attribution and the remaining lever (the flash-MLA KV-replication tax, +1.2 ms/step) are in `docs/models/glm52/moe-tp8-low-latency.md`.

## Correctness

- Plain DET ×2, MTP DET, **MTP text byte-identical to plain** (lossless speculative contract end-to-end through 78 sharded layers + AR + replicated MoE).
- Oracle gates on 8×H200: attention twin gate (8-head shards + AR vs full 64-head reference, cross-rank bitwise identity), replicated-MoE full-layer gate (62/64 probes + 2 tolerated tie-flips, cross-rank bitwise identity), AR brick gate (tag/parity discipline across steps, want-mask NaN-canary zero-fill assert, 7.74 µs/layer).
- EP8 path untouched; `kv_offload` × tp8 rejected at launch.

Toxic review: approved, zero correctness defects; the two stale comments it flagged are fixed in d7c9340.


xiaguan:
Re the Codex P2 (leftover drafts riding the solo→concurrent transition): verified the scenario — it is real but bounded to exactly one step, and we are keeping the drain behavior deliberately (comment amended in 8af5b48).

- The suppression block stops **new** proposals while `count != 1`; drafts already installed on the solo slot are fed once more as a verify span in the transition step, then no new drafts install until the fleet drains back to solo.
- The newcomer cannot be starved: `plan_step_shapes` gives every active slot its liveness row first, then splits the spare rows round-robin — the new request gets at least half of them in that step.
- Partial spans are already the contract: the planner may grant fewer rows than `feed_want`, the span becomes a prefix, and `advance_span` discards the un-fed drafts. Mixed span+prefill bucket-8 steps are the same shape class EP8 serves routinely.
- Clearing installed drafts at the transition would discard paid-for speculation (those verify rows can commit several tokens for the solo request) and add a drafter-reset path, to speed one step of one prefill by a few rows. Net negative.

PR #629, opened 2026-07-09
fix(qwen3): fail loud at the --batch-invariant envelope boundary

### LINKED ISSUE #628: qwen3: make the --batch-invariant envelope fail loud at its boundary
## Description

1. typed_ops GEMMs bypass the numeric-policy gate silently

`ops::linear` routes every projection GEMM through the numeric-policy match: under `Pin`/`PerToken` the call goes to the pinned or per-token path, or fails loud. `openinfer-kernels/src/typed_ops.rs` has a private `launch_gemm` twin that never reads `numeric_policy()` under any policy: it goes straight to the default `cublasGemmEx` heuristic (`CUBLAS_GEMM_DEFAULT_TENSOR_OP`), not even the `gemm_lt_tune` plan cache. Today this is unreachable in a supported process — the typed_ops consumers are the Kimi-K2 runner only, and the CLI rejects `--batch-invariant` for non-qwen3 models — but the boundary is enforced by nothing except that accident of crate wiring.

2. `--batch-invariant --tp-size 2` starts, with zero evidence for the TP reduction axis

`ensure_batch_invariant_supported` (`openinfer-qwen3/src/lib.rs`) never sees the TP world size, so the combination boots. The all-reduce reduction order under TP is unaudited: since #611 the decode all-reduce runs inside per-bucket captured graphs, while the prefill and unified all-reduces are eager with message sizes that vary continuously with batch composition — and NCCL selects algorithm and protocol by message size, so batch composition can change the reduction order. That is exactly the axis `--batch-invariant` claims to close.

3. The pinned algo's numeric configuration is invisible and unchecked

`gemm_lt_pin_tune` pins `results[0].algo` from the cuBLASLt heuristic and never inspects what it pinned. The two config attributes that decide the GEMM's K-reduction — `CUBLASLT_ALGO_CONFIG_SPLITK_NUM` and `CUBLASLT_ALGO_CONFIG_REDUCTION_SCHEME` — are baked into the pinned algo and replayed at every N, which is exactly why the pin works; but nothing reads them back, so the property the whole feature rests on is invisible at boot and unasserted.

FeathBow:
## Description

Fixes #628 

**1. `typed_ops` GEMMs no longer bypass the numeric-policy gate.** `ops::linear::launch_gemm` opens with `match numeric_policy()`, so under `Pin` a projection GEMM is either served by the pinned algo or fails loud. `typed_ops` has a same-named private twin that reads the policy under no circumstances: it calls `cublasGemmEx` with `CUBLAS_GEMM_DEFAULT_TENSOR_OP` directly, not even the `gemm_lt_tune` plan cache. Both of its launchers now go through one shared `ensure_tuned_policy`, which bails unless the policy is `Tuned`.

This is unreachable in any supported process today, and the fix does not make it reachable. `NUMERIC_POLICY` defaults to `Tuned`; only qwen3 ever writes `Pin`; qwen3 never touches `typed_ops` (whose only consumer is the Kimi-K2 runner, directly and through the `typed_pipeline!` macro). That chain is the whole reason the bypass has never bitten, and it was written down nowhere. It is now a doc comment on the guard, and the guard turns a silent bypass into a startup error the day someone breaks the chain.

**2. `--batch-invariant` is rejected under tensor parallelism.** `ensure_batch_invariant_supported` never saw the TP world size, so the combination booted and armed the pin. It is now rejected at the engine boundary (the load-bearing guard, reachable by any library consumer of `start_engine_with_offload`) and again at the CLI, matching how decode-overlap is rejected at both layers.

The reduction order under TP has never been audited. Since #611 the decode all-reduce runs inside graphs captured per `(bucket, attention_path)`, while the prefill and unified all-reduces are eager with message sizes that vary continuously with batch composition. NCCL selects algorithm and protocol by message size, so batch composition can move the reduction order — the axis `--batch-invariant` exists to close. The honest status is *unaudited*, not *known broken*, and that is what the error text and `tp-design.md` now say.

**3. The pinned algo's numeric configuration is read back and logged.** `gemm_lt_pin_tune` pinned `results[0].algo` and never looked at what it pinned. The two attributes that decide the K-reduction, `CUBLASLT_ALGO_CONFIG_SPLITK_NUM` and `CUBLASLT_ALGO_CONFIG_REDUCTION_SCHEME`, are baked into the pinned algo and replayed at every N — `lt_pin_run` reuses `plan.algo`, `plan.op` and `plan.a` verbatim and rebuilds only the B/C layouts — which is precisely why the pin works. 

Selection is unchanged: the heuristic's first result is still what gets pinned, and `cublasLtMatmulAlgoConfigGetAttribute` is a getter. What is new is one startup line per pinned `(M,K)`, and a boot-time bail if the scheme is outside `cublasLtReductionScheme_t`'s documented set. 

## Before and after

Measured on Single GPU (sm_89, x86_64). 

`--batch-invariant --tp-size 2`, on base:

```
INFO openinfer_qwen3 Qwen3 numeric policy: Pin (--batch-invariant=true)
...
Error: 0: failed to start Qwen3 engine
       1: Failed to set CUDA device 1: cudaError=101
```

Zero rejections at any layer. The pin is armed under TP, and the run ends only because this box has one GPU. With this branch:

```
$ openinfer --model-path <qwen3-4b> --batch-invariant --tp-size 2
Error: --batch-invariant is not supported with --tp-size > 1; enable one at a time
$ echo $?
1
```

Single-GPU `--batch-invariant` still boots, so the reject does not over-fire.

The new startup log, on the same GPU:

```
Qwen3 numeric policy: Pin (--batch-invariant=true)
Qwen3 GEMM pin: m=4096,   k=2560, splitk=2, reduction_scheme=output_type
Qwen3 GEMM pin: m=1024,   k=2560, splitk=8, reduction_scheme=output_type
Qwen3 GEMM pin: m=2560,   k=4096, splitk=3, reduction_scheme=output_type
Qwen3 GEMM pin: m=9728,   k=2560, splitk=1, reduction_scheme=none
Qwen3 GEMM pin: m=2560,   k=9728, splitk=3, reduction_scheme=output_type
Qwen3 GEMM pin: m=151936, k=2560, splitk=1, reduction_scheme=none
```

Four of the six pinned shapes really do use split-K, and every one of them reduces through `OUTPUT_TYPE` — partials to workspace, summed by a separate kernel in a fixed order. That is the property the pin depends on, and until now nobody could see it. These values are specific to this GPU and this cuBLAS version.

The block prints twice per boot, once from the memory-profile thread and once from the serving worker. The pin cache is `thread_local`, so each thread pins independently.

## Tests

None added. The TP reject is observable end to end from the CLI.

Single GPU (sm_89, x86_64).

`batch_invariance_decode_gemm_graph` and `batch_invariance_endtoend` are green on both, unchanged: 83.6s / 27.2s on base, 82.0s / 27.2s here. 

## Type of Change

### DIFF
--- a/docs/models/qwen3/model-crate.md
+++ b/docs/models/qwen3/model-crate.md
@@ -1,8 +1,12 @@
 # Qwen3-4B Model Crate
 
 **Created**: 2026-05-03
-**Last touched**: 2026-06
+**Last touched**: 2026-07
 **TL;DR**: `crates/openinfer-qwen3` now owns Qwen3 config, weights, execution, scheduler, tests, benches, and kernel plan. Root `openinfer` loads Qwen3 through a generic `EngineHandle` and no longer contains `Qwen3Model`, `Qwen3Executor`, `ModelRuntimeConfig`, root Qwen3 tests, or `src/model/qwen3/*`. The old `ModelForward` path has been removed; decode length-limit now emits the final token before `Finished`. Long-context `bs=1` TPOT was traced to non-partition FlashInfer paged decode under-filling the GPU; Qwen3 runtime gates FlashInfer split-K decode on `padded_bs<=32` (the `seq_len>=1024` gate was dropped in #437) with a 64-token chunk floor (`Tuned` capped at 64 chunks; opt-in `--batch-invariant` pins a fixed 160-token split), cutting 4k/64 serving steady TPOT from about `11.7ms` to `6.46ms` on RTX 5090. Qwen3 now keeps a single model-crate bench entry: `qwen3_kernel_snapshot`, a JSON snapshot runner with warm/cold-L2 latency, default-on CUPTI counters, and compare. Correctness/truth is intentionally out of this snapshot for now.
+
+## Determinism scope
+
+`--batch-invariant` is scoped to one process, one toolchain, and one GPU. The cuBLASLt algo is re-tuned on every boot and cached per executor thread (`thread_local`), so it does not promise stability across cuBLAS or driver versions; startup logging records the pinned `splitk` and `reduction_scheme` per `(M,K)`. The builder rejects decode-overlap, DFlash speculative decoding, LoRA, and tensor parallelism.
 
 ## Preparation
 
--- a/docs/models/qwen3/tp-design.md
+++ b/docs/models/qwen3/tp-design.md
@@ -56,6 +56,7 @@ The first TP milestone does not aim to do the following:
 - preserve current CUDA Graph behavior from day one
 - introduce vocab-parallel embedding or vocab-parallel `lm_head` in the first pass
 - optimize every path for peak throughput before the basic design is proven correct
+- serve `--batch-invariant`, which the builder rejects for `tp_size > 1`: the prefill and unified all-reduces are eager and unbucketed, so NCCL can pick a different algorithm and protocol as the message size moves with batch composition, and that reduction order has never been audited
 
 ## Simplifications Allowed In First Pass
 
--- a/openinfer-kernels/csrc/shared/linear.cu
+++ b/openinfer-kernels/csrc/shared/linear.cu
@@ -4,6 +4,7 @@
 
 #include <array>
 #include <cstddef>
+#include <cstdint>
 #include <map>
 
 static constexpr int CUBLAS_STATUS_ERROR_OFFSET = 100000;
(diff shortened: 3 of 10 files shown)


xiaguan:
need rebase

FeathBow:
Great. Done.

PR #201, opened 2026-06-01
test(sim): add front e2e tests using the simulated engine

Mrtroll486:
## Summary

Implements #137. Follow-up from #136.

This adds lightweight CPU-only frontend e2e coverage for `pegainfer-sim`. The test starts the simulated engine behind the real vLLM/OpenAI HTTP frontend and verifies:

- `/v1/models`
- non-streaming `/v1/completions`
- streaming `/v1/completions` reaches `data: [DONE]`

The test uses a tiny local tokenizer/model metadata fixture and does not require CUDA, model weights, or real model correctness.

The CPU CI path now runs the simulated frontend e2e test with `cargo test --release -p pegainfer-sim --test frontend_e2e`.

## Evidence

Static / metadata checks:

- `cargo fmt --all --check`
- `cargo metadata --locked --no-deps --format-version 1`

Simulated frontend e2e:

- `cargo test --release -p pegainfer-sim --test frontend_e2e -- --nocapture`

Result:

- `/v1/models`: passed
- non-streaming `/v1/completions`: passed, returned non-empty output for `max_tokens > 0`
- streaming `/v1/completions`: passed, emitted terminal `data: [DONE]`
- metadata contract test: passed, documents the tiny tokenizer/model metadata required by the simulated frontend path

## Claim Boundary

This PR proves the simulated engine can exercise the real vLLM/OpenAI HTTP frontend in a CPU-only test path without CUDA or model weights.

It does not claim real model correctness, GPU/CUDA coverage, throughput improvement, benchmark validity, or production serving performance.

## Follow-up

Future frontend coverage can extend this path to chat/completions variants, request validation, or benchmark-client compatibility while keeping the simulator separate from real model correctness tests.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)

- [x] New feature (non-breaking change which adds functionality)

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

- [x] Documentation update

## Checklist

- [x] My code follows the style guidelines of this project (see `docs/areas/coding-style.md`).

- [x] I have performed a self-review of my own code.

- [x] I have formatted my commits according to Commitizen conventions.

- [x] I have run the local test suite and all tests pass (see `CLAUDE.md`).

Issue #137, opened 2026-05-17
Add frontend e2e tests using the simulated engine


CAICAIIs:
Follow-up from #136.

After `pegainfer-sim` lands, we can add frontend e2e coverage that exercises the real vLLM/OpenAI HTTP path without requiring CUDA or model weights.

## Goal

Add lightweight e2e tests that start the simulated engine and verify the frontend behavior through HTTP.

## Scope

In scope:
- Start `pegainfer-sim` in a test or test harness.
- Verify `/v1/models`.
- Verify non-streaming `/v1/completions`.
- Verify streaming `/v1/completions` reaches `data: [DONE]`.
- Keep the test CPU-only and independent of real model weights.

Out of scope:
- Real model correctness.
- GPU/CUDA paths.
- Performance claims.

## Acceptance Criteria

- The test runs in the CPU-only CI path.
- The test fails if completions return empty output for `max_tokens > 0`.
- The test fails if streaming does not emit a terminal `[DONE]`.
- The test documents any tokenizer/model metadata required by the simulated frontend path.

Mrtroll486:
Hi! I'm intrested in this project, can I take this issue as a start?

Mrtroll486:
I've looked the updated code, the buggy ci path has been fixed, and my tests works properly, so I think this issue can be closed.

Issue #78, opened 2026-05-05
Preserve streaming usage in the vLLM frontend


xiaguan:
## Context

The vLLM serving benchmark sends streaming completion requests with:

- `stream: true`
- `stream_options.include_usage: true`
- `ignore_eos: true`

The benchmark records output token counts from the final streaming `usage.completion_tokens` field. If that field is missing, `vllm/benchmarks/serve.py` falls back to re-tokenizing `generated_text`, which can inflate token counts for Unicode-heavy generated text.

## Observed behavior

On a Qwen3-4B QPS benchmark with seed 42, `ignore_eos`, and temperature 0:

- Expected output tokens from the random workload: `64,446`
- vLLM reported: `64,446`
- PegaInfer reported: `69,403`

A detailed PegaInfer run showed only 8 out of 500 requests had mismatched `output_lens`. Those mismatched lengths matched the re-tokenized generated text length, not the requested output length, which strongly suggests the benchmark did not receive final usage for those streams.

## Likely source

The scheduler appears to enforce `max_tokens` correctly:

- `crates/pegainfer-qwen3-4b/src/scheduler/resolve.rs`
- `completion_tokens >= req.max_tokens` finishes with `FinishReason::Length`

The more likely issue is in the vLLM frontend bridge:

- `src/vllm_frontend.rs::run_request_stream` receives `TokenEvent::Finished { finish_reason, .. }` and discards `prompt_tokens` / `completion_tokens`.
- `src/vllm_frontend.rs::send_terminal_output` sends a terminal `EngineCoreOutput` with empty `new_token_ids` and a finish reason, but no token count / usage information.

## Suggested fix

Thread `prompt_tokens` and `completion_tokens` from `TokenEvent::Finished` through the terminal-output path and connect them to the usage mechanism expected by the current vLLM frontend/output processor.

A good implementation should verify that streaming completions with `stream_options.include_usage=true` reliably emit final usage under concurrent load.

## Acceptance criteria

- Streaming completion responses include final `usage.completion_tokens` for every successful request.
- The vLLM benchmark no longer falls back to generated-text re-tokenization for PegaInfer responses.
- A Qwen3-4B random QPS benchmark with fixed seed reports total output tokens equal to the generated workload total.

FeathBow:
This issue currently doesn't seem reproducible on the latest main now? I have some logs I'll paste here shortly.

FeathBow:
Re-ran the test described in this issue on current main.

### Test Env

[...]

PR #630, opened 2026-07-09
test(sim): add chat completions e2e tests

bytelazy:
## Description

Add CPU-only end-to-end coverage for `/v1/chat/completions` through the real OpenAI-compatible frontend and `openinfer-sim`.

Previously the simulated frontend suite covered `/v1/completions`, but did not guard the chat response shape or the SSE wire contract.

This adds three chat-completions E2E scenarios:

- Non-streaming response:
  - `object == "chat.completion"`
  - assistant role and non-empty content
  - `finish_reason == "length"`
  - prompt/completion/total token accounting
- Streaming response:
  - `Content-Type: text/event-stream`
  - each JSON chunk and `[DONE]` is a separately dispatched, blank-line-delimited SSE event
  - all chunks use `object == "chat.completion.chunk"`
  - the first delta declares the assistant role
  - streamed `delta.content` is concatenated and checked
  - the finish chunk carries `finish_reason == "length"`
  - exactly one terminal `[DONE]` event, accepting both `data: [DONE]` and `data:[DONE]`
- Streaming usage:
  - exactly one usage-only chunk with `choices == []`
  - emitted immediately after the finish chunk and before `[DONE]`
  - `completion_tokens == max_tokens`
  - `total_tokens == prompt_tokens + completion_tokens`

The tiny tokenizer fixture now includes a minimal deterministic `chat_template` that makes streamed content observable without a real model directory.

The SSE helper also has controlled negative coverage for missing event separators, duplicate mixed-spacing `[DONE]` events, and an undispatched final event at EOF.

## Why

This closes the frontend test gap for chat and agent clients without CUDA, model weights, or a GPU. The tests still exercise the same HTTP frontend used by real serving paths.

## Test plan

- [x] `CARGO_NET_GIT_FETCH_WITH_CLI=true cargo test --release -p openinfer-sim --test frontend_e2e`
  - `13 passed; 0 failed`
- [x] `CARGO_NET_GIT_FETCH_WITH_CLI=true cargo clippy -p openinfer-sim --tests`
  - no `openinfer-sim` warnings
- [x] `cargo fmt --all --check`

Rebased onto upstream `main` at `6b6b6c37`.

## Type of Change

- [x] Test coverage
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change


FeathBow:
Thanks for extending the CPU-only frontend E2E coverage introduced in #201, which implemented #137, from /v1/completions to /v1/chat/completions.

This is a useful addition: it exercises the real OpenAI-compatible HTTP frontend through openinfer-sim without requiring CUDA, model weights, or a GPU. In particular, the non-streaming test provides meaningful coverage of the response shape, assistant content, finish reason, and usage arithmetic.

I checked out the PR head (112dae1) and ran the release E2E suite; all 8 tests passed. I then inspected the actual SSE payloads and ran controlled negative checks against the new assertions.

That validation found two blocking coverage gaps: the streaming test passes without emitting or asserting any delta.content, and [DONE] is checked only for presence rather than as the terminal SSE event. The usage response is currently correct, but its test accepts malformed counters and does not verify the usage-only chunk contract. I also found a lower-severity documentation reproducibility issue caused by an author-local file path.

openinfer-sim/tests/frontend_e2e.rs
```diff
@@ -221,6 +221,178 @@ async fn streaming_completion_emits_terminal_done() -> Result<()> {
...
+            "chunk object must be chat.completion.chunk: {chunk}"
+        );
+    }
+
+    assert_eq!(
+        chunks[0]["choices"][0]["delta"]["role"].as_str(),
+        Some("assistant"),
+        "first chunk must declare assistant role: {}",
+        chunks[0]
+    );
```
FeathBow, line 618: The role assertion itself is valid, but this test never validates any delta.content. On this exact PR head, the SSE response contained only the assistant-role chunk, an empty finish chunk, and [DONE]; no content chunk was emitted, yet the test passed.

Please adjust the fixture/request so that it produces observable streamed content, then concatenate and assert the resulting delta.content. Otherwise the content path named by this test can disappear while the test remains green.

openinfer-sim/tests/frontend_e2e.rs
```diff
@@ -221,6 +221,178 @@ async fn streaming_completion_emits_terminal_done() -> Result<()> {
...
+
+    assert!(
+        stream_text
+            .lines()
+            .any(|line| line.trim() == "data: [DONE]"),
```
FeathBow: This predicate only proves that [DONE] appears somewhere, while the assertion message says the response must end with it. In a controlled negative check, the invalid order role → [DONE] → finish still passed.

Please assert that the final SSE data payload is exactly [DONE], and preferably that it appears exactly once.

openinfer-sim/tests/frontend_e2e.rs
```diff
@@ -221,6 +221,178 @@ async fn streaming_completion_emits_terminal_done() -> Result<()> {
...
+    let chunks: Vec<Value> = stream_text
+        .lines()
+        .filter_map(|line| line.strip_prefix("data: "))
+        .filter(|data| *data != "[DONE]")
+        .map(serde_json::from_str)
+        .collect::<std::result::Result<Vec<_>, _>>()
+        .context("failed to parse streaming chunks")?;
+
+    assert!(
+        chunks.iter().any(|chunk| !chunk["usage"].is_null()),
```
FeathBow: This verifies only that some chunk has a non-null usage value. A malformed {}, incorrect counters, or usage attached to a normal choices chunk would all pass. In a controlled negative check, completion_tokens=999 and total_tokens=0 still passed.

Please assert a single usage-only chunk with choices == [], emitted after the finish chunk and before [DONE], with completion_tokens == 3 and total_tokens == prompt_tokens + completion_tokens. This would also provide regression coverage for the class of streaming-usage failures historically tracked in #78.

docs/subsystems/frontend/chat-completions-e2e.md
```diff
@@ -0,0 +1,57 @@
...
+> **TL;DR:** Simulated frontend now covers `/v1/chat/completions` E2E for non-streaming format, streaming chunks, usage chunks, and the minimal fixture `chat_template`.
+>
+> **Last touched:** 2026-07
+
+## Preparation
+
+- **Read**:
+  - `docs/index.md` — routed this as frontend subsystem work.
+  - `docs/subsystems/frontend/simulated-inference-engine.md` — confirmed `openinfer-sim` is the CPU-only frontend validation harness and its metadata fixture is intentionally minimal.
+  - `/Users/asher/Downloads/openinfer-chat-completions-e2e-pr_20260704_135002.md` — identified the intended change: three `/v1/chat/completions` integration tests plus a fixture `chat_template`.
```
FeathBow: This document references an author-local /Users/asher/Downloads/... file that other contributors cannot access or verify. It also records machine-specific provenance in a committed subsystem document.

FeathBow (changes requested):

openinfer-sim/tests/frontend_e2e.rs
```diff
@@ -633,6 +847,29 @@ fn completion_body(model_name: &str, stream: bool) -> Value {
     })
 }
 
+fn parse_terminal_sse_chunks(stream: &str) -> Result<Vec<Value>> {
+    let payloads: Vec<&str> = stream
+        .lines()
+        .filter_map(|line| line.strip_prefix("data: "))
```
FeathBow: The strengthened content/order/accounting assertions look good, but parse_terminal_sse_chunks still parses individual data: lines rather than SSE events. A full role → content → finish → [DONE] body with the blank event separators removed passes every current assertion, even though SSE clients treat those lines as fields of one event (or discard the pending event at EOF), so they never receive independent chunks or a terminal [DONE].

The helper also ignores the valid no-space form data:[DONE], which means a duplicate DONE event can escape the count. Could this parse blank-line-delimited SSE events, or use an SSE parser, and then assert each JSON payload and [DONE] is a separate dispatched event? Checking Content-Type: text/event-stream would also complete the wire-level contract.

openinfer-sim/tests/frontend_e2e.rs
bytelazy: Addressed. The deterministic chat fixture now produces observable streamed content, and the test concatenates delta.content and asserts the exact simulated output ( alpha alpha). Release E2E passes on the rebased branch.

openinfer-sim/tests/frontend_e2e.rs
bytelazy: Addressed. The SSE helper now requires exactly one terminal [DONE] event and preserves event order. The latest follow-up also parses blank-line-delimited dispatched events and recognizes both data: [DONE] and data:[DONE].

openinfer-sim/tests/frontend_e2e.rs
bytelazy: Addressed. The test now requires exactly one usage-only chunk with choices == [], immediately after the finish chunk and before [DONE], and validates completion_tokens == 3 plus total_tokens == prompt_tokens + completion_tokens.

docs/subsystems/frontend/chat-completions-e2e.md
bytelazy: Addressed. The author-local Downloads path was removed from the committed subsystem document and replaced with PR review provenance.

openinfer-sim/tests/frontend_e2e.rs
bytelazy: Addressed in c8ff4c2. The helper now parses blank-line-delimited SSE events, joins multi-line data fields, accepts optional post-colon spaces, and rejects pending data at EOF. Controlled negative tests cover missing separators, duplicate mixed-spacing [DONE], and an undispatched final event. Both chat streaming tests also assert Content-Type: text/event-stream. Release E2E: 13 passed, 0 failed.

bytelazy:
@FeathBow All five review threads are now addressed, replied to, and resolved.

The second-round SSE wire-contract follow-up is in c8ff4c2:
- parses blank-line-delimited dispatched SSE events rather than individual data lines
- accepts both data: [DONE] and data:[DONE]
- rejects missing separators, duplicate DONE events, and pending data at EOF
- checks Content-Type: text/event-stream

Validation on the rebased branch:
- cargo test --release -p openinfer-sim --test frontend_e2e: 13 passed
- cargo clippy -p openinfer-sim --tests: pass
- cargo fmt --all --check: pass

Could you please re-review when convenient?

openinfer-sim/tests/frontend_e2e.rs
```diff
@@ -633,6 +847,81 @@ fn completion_body(model_name: &str, stream: bool) -> Value {
...
+    }
+    Ok(())
+}
+
+fn parse_terminal_sse_chunks(stream: &str) -> Result<Vec<Value>> {
+    let normalized = stream.replace("\r\n", "\n").replace('\r', "\n");
+    let mut payloads = Vec::new();
+    let mut data_lines = Vec::new();
+
+    for line in normalized.split('\n') {
```
FeathBow: One EOF edge still slips through: split('\n') turns data:[DONE]\n into ["data:[DONE]", ""], so the synthetic trailing item is treated as a blank-line dispatch even though a single LF only terminates the data: line and pending data must be discarded at EOF. split_terminator('\n') (or equivalent) would preserve the SSE dispatch semantics.

docs/subsystems/frontend/chat-completions-e2e.md
```diff
@@ -0,0 +1,85 @@
...
+- **Outcome**: Implemented the chat-completions E2E coverage and addressed both PR #630 review rounds. The streaming tests now fail if content disappears, SSE events are not independently dispatched, Content-Type is wrong, `[DONE]` is duplicated or non-terminal, or the usage-only chunk is malformed, misplaced, or miscounted.
+- **Pitfalls encountered**:
+  - The first `cargo test` attempt failed during git dependency fetch for `dynamo` with `unexpected return value from ssl handshake -9806`.
+  - Retrying with `CARGO_NET_GIT_FETCH_WITH_CLI=true` let Cargo fetch git dependencies through the system git client and proceed.
+- **Lessons learned**:
+  - For this frontend test path, `openinfer-sim` still needs chat-capable tokenizer metadata even though it does not load model weights.
+  - SSE tests must preserve data-event order; filtering `[DONE]` before assertions makes terminal-contract regressions invisible.
+  - SSE wire coverage must model event dispatch, not just scan `data:` lines; blank separators and EOF handling are part of the contract.
+- **Follow-ups**:
+  - All five PR #630 review threads were answered and resolved; maintainer re-review remains.
```
FeathBow: maybe redundant

docs/subsystems/frontend/chat-completions-e2e.md
```diff
@@ -0,0 +1,85 @@
+# Chat Completions E2E
```
FeathBow: Btw this file is mostly PR-local planning, validation output, and review-thread bookkeeping already captured by the diff, CI, and GitHub, so it will quickly go stale; the only durable point about chat-capable tokenizer metadata belongs in the existing simulated-inference-engine.md, so could we drop this file and its index entry?

openinfer-sim/tests/frontend_e2e.rs
bytelazy: Addressed in b46bba6. The parser now uses split_terminator('\\n'), so a single LF only terminates the data line and does not synthesize a dispatching blank line. Added a controlled negative case for data:[DONE]\\n at EOF, alongside the existing no-LF EOF case. Release E2E: 13 passed.

docs/subsystems/frontend/chat-completions-e2e.md
bytelazy: Agreed. The redundant PR-local material was removed by deleting chat-completions-e2e.md in b46bba6.

docs/subsystems/frontend/chat-completions-e2e.md
bytelazy: Agreed and addressed in b46bba6. Deleted chat-completions-e2e.md and its docs/index.md entry. Moved the durable chat_template and observable streamed-content requirement into the Frontend Metadata Contract section of simulated-inference-engine.md.

bytelazy:
@FeathBow The latest review items are addressed in b46bba6:

- switched the SSE parser to split_terminator so data:[DONE]\n at EOF is not falsely dispatched
- added a controlled negative test for the single-LF EOF case
- deleted the PR-local chat-completions-e2e.md and its index entry
- moved the durable chat_template/observable-content requirement into simulated-inference-engine.md

Validation remains green:
- release frontend E2E: 13 passed
- clippy: pass
- cargo fmt --all --check: pass

All review threads are replied to and resolved. Could you please re-review when convenient?

FeathBow:
LGTM.

PR #632, opened 2026-07-10
perf(deepseek-v2-lite): group EP2 expert routes and compute gate logits on CUDA

### LINKED ISSUE #464: deepseek-v2-lite: profile flat HTTP throughput under active decode batches
## Background

PR #463 adds a retained DeepSeek-V2-Lite EP2 benchmark snapshot for OpenInfer host-staged, OpenInfer NCCL, vLLM TP2, and vLLM TP2+EP2.

The retained snapshot shows that OpenInfer can reach non-1 active request sets under HTTP pressure, but output throughput stays nearly flat as concurrency increases. This looks like a scheduler/runtime profiling follow-up, not a benchmark-script issue.

Benchmark snapshot:
`docs/benchmarks/deepseek-v2-lite-vllm-tp2-ep2-2026-06.md`

## Current Observation

OpenInfer HTTP pressure rows completed all `max_concurrency=1/4/8` cells, but output throughput changed only slightly:

| Engine | c1 output tok/s | c4 output tok/s | c8 output tok/s |
| --- | ---: | ---: | ---: |
| OpenInfer host-staged | 27.122 | 26.452 | 26.566 |
| OpenInfer NCCL | 27.376 | 27.892 | 28.509 |

The OpenInfer trace pass shows active requests and decode batches did form:

| Engine | c1 active/decode max | c4 active/decode max | c8 active/decode max |
| --- | --- | --- | --- |
| OpenInfer host-staged | 1 / 1 | 4 / 4 | 8 / 5 |
| OpenInfer NCCL | 1 / 1 | 4 / 2 | 8 / 5 |

In the supplemental FlashInfer-fixed vLLM validation, vLLM HTTP output throughput scaled with concurrency under the same client/workload contract. Those rows are supplemental validation, not a stock-package vLLM claim.

## Goal

Profile why OpenInfer active HTTP requests are not turning into higher effective decode throughput for DeepSeek-V2-Lite EP2.

Keep this as a runtime/scheduler investigation. Do not treat the benchmark snapshot as a production serving claim or a vLLM parity claim.

## Areas To Profile

Likely bottleneck areas:

- serial per-request prefill or admission work;
- same-position-only decode subgrouping limiting effective batching;
- per-layer host D2H/H2D work in norm, attention, routing, or bookkeeping;
- per-token GEMM loops that do not batch across active requests effectively;
- per-route MoE replay overhead;
- NCCL EP2 transport/combine overhead versus host-staged runtime overhead.

## Suggested Evidence

Collect evidence for at least host-staged and NCCL on the retained HTTP contract:

- `input_len=64`
- `output_len=64`
- `num_prompts=32`
- `max_concurrency=1,4,8`
- `temperature=0`
- `ignore_eos=true`

Useful outputs:

- per-step active request count;
- decode batch/subgroup size;
- prefill/admission time;
- GPU step time;
- per-layer host time;
- MoE routing/dispatch/combine timing;
- output tok/s and TPOT from the same artifact path;
- a short conclusion separating scheduler overhead, GPU kernel shape, and EP communication.

## Acceptance Boundary

This issue is complete when we can explain which runtime/scheduler component prevents throughput scaling under active HTTP pressure, with profile evidence from the retained benchmark contract.

It does not need to optimize the runtime in the same issue. If the profile points to a clear fix, open or link a separate implementation issue.

CAICAIIs:
## Summary

- group DeepSeek-V2-Lite host-staged and NCCL expert routes by stable `(owner_rank, global_expert)` while preserving original token-major contribution accumulation;
- compute NCCL gate logits on CUDA while keeping the existing host top-k, softmax, tie-breaking, and route-plan rules;
- extend HTTP trace attribution with request phase timing plus singleton/batched decode-step accounting;
- harden benchmark evidence so missing traces, partial trace coverage, workload drift, missing output hashes, and partial HTTP completion fail closed.

## Root causes

Under the retained 2x RTX 5090 short fixed-shape HTTP contract (`prompt_words=64`, `max_tokens=64`, concurrency `1/4/8`, greedy, ignore-EOS), throughput flattened for two measured reasons:

1. host-staged and NCCL replayed many small expert calls per route;
2. NCCL copied the full hidden state to the host and ran the gate GEMM on CPU.

The trace refresh showed that active sets and batched decode steps do form under HTTP pressure, so scheduler admission was not the final limiter for this retained shape.

## Performance evidence

Host-staged paired HTTP run:

| Concurrency | Baseline | Grouped routes | Change |
| ---: | ---: | ---: | ---: |
| 1 | 20.73 tok/s | 23.72 tok/s | +14.4% |
| 4 | 21.13 tok/s | 24.58 tok/s | +16.3% |
| 8 | 21.41 tok/s | 25.32 tok/s | +18.3% |

Three paired NCCL HTTP runs under the same request contract:

| Concurrency | Median change |
| ---: | ---: |
| 1 | +23.4% |
| 4 | +21.7% |
| 8 | +29.5% |

All NCCL paired cells completed under the corrected 32-request contract: `18` cells / `576` requests, with zero failures, timeouts, or missing traces. Each paired baseline/optimized cell preserved its request output hash set.

## Correctness and verification

- CUDA router logits match the retained host accumulation bitwise.
- Grouped route ordering, row selection, scatter/combine, rollback parsing, trace payload fields, workload metadata, strict trace coverage, and partial HTTP completion have focused tests.
- Local checks:
  - `python3 -m py_compile scripts/bench_dsv2lite_vllm_matrix.py scripts/bench_http_serving.py`
  - `python3 tests/test_bench_dsv2lite_vllm_matrix.py` (`76` tests)
  - `python3 tests/test_bench_http_serving.py` (`15` tests)
  - `cargo fmt --all --check`
  - `git diff --check`
- Remote validation:
  - synced Python benchmark tests passed (`76` + `15` tests);
  - Rust/CUDA narrow gate passed on 2x RTX 5090 + CUDA 12.8 for DSV2-Lite MoE/runtime and scheduler trace tests.

Diagnostic rollback switches:

- `OPENINFER_DSV2_LITE_HOST_STAGED_EXPERT_BATCH=serial`
- `OPENINFER_DSV2_LITE_NCCL_EXPERT_BATCH=serial`
- `OPENINFER_DSV2_LITE_NCCL_ROUTER=host`

These are diagnostic/emergency rollback controls for retained A/B and recovery. They are not production tuning knobs.

## Scope

This PR closes #464 with profiling evidence and measured fixes for the identified MoE replay and NCCL router bottlenecks.

It does not claim mixed/long-prompt scaling, soak or SLO readiness, fully device-resident routing, fewer token-wise GEMM launches, sparse or multi-node EP, production readiness, or vLLM parity.

The remaining attention/KV host-side work is a separate implementation contract because it changes KV/cache ownership, long-context capacity, request retirement, and CUDA Graph pointer-stability assumptions.

Fixes #464


docs/models/deepseek-v2-lite/status.md
```diff
@@ -60,6 +60,66 @@ Latest 2026-06-28 result on 2x RTX 5090:
...
+
+Three paired NCCL HTTP runs used `64/64`, `num_prompts=32`, concurrency `1/4/8`, `temperature=0`, `ignore_eos=true`, and no warmup:
+
+| Concurrency | Baseline output tok/s | Optimized output tok/s | Change | Decode mean |
+| ---: | ---: | ---: | ---: | ---: |
+| 1 | `25.369` | `30.089` | `+18.6%` | `+3.1%`; no c1 decode-speed claim |
+| 4 | `24.486` | `30.782` | `+25.7%` | `-42.25%` |
+| 8 | `24.723` | `32.286` | `+30.6%` | `-30.06%` |
+
+All `18` cells and `144` requests completed with `failed=0`, `timeouts=0`, and `missing_trace=0`; every paired run preserved the same eight-prompt hash set. The retained public artifact basename is `issue464-http-nccl-three-pair-summary.json`.
```
FeathBow: I think there might be some room here to refine the benchmarking descriptions and alignment. This section says the paired HTTP runs used `num_prompts=32`, but the retained accounting says `18 cells / 144 requests` and refers to the same eight-prompt hash set. That is 8 requests per cell; if every cell had used 32 prompts, the total would have been 576 requests.

The committed runner also has two different input-length semantics: the HTTP matrix passes `--random-input-len 64` to `vllm bench serve`, while the trace driver passes `--prompt-words 64` to `bench_http_serving.py`. The latter does not guarantee a 64-token prompt.

The percentages may still be valid for the measured eight-request workload, but the retained evidence does not substantiate the stated 32-prompt contract. Should we rerun the cells with 32 requests and verified lengths, or would it be easier to just relabel this to reflect the actual counts?

scripts/bench_dsv2lite_vllm_matrix.py
```diff
@@ -1017,10 +1053,10 @@ def run_openinfer_trace_pass(args: argparse.Namespace, out_dir: Path) -> list[di
...
                         "completed": payload["summary"]["completed"],
                         "failed": payload["summary"]["failed"],
+                        "timeouts": payload["summary"]["timeouts"],
                         "output_tok_s": payload["summary"]["output_tokens_per_s"],
-                        "trace": payload.get("server_trace", {}),
                     }
-                    cell["missing_trace_count"] = trace_missing_count(cell["trace"])
+                    cell.update(trace_summary_for_payload(payload))
                     row["cells"].append(cell)
             row["passed"] = bool(row["cells"]) and all(
```
FeathBow, line 1461: One thing I noticed while reading the trace pass: a successful HTTP response is currently sufficient for this trace row to pass. If no `openinfer_http_trace` records are emitted or matched, `bench_http_serving.py` still exits successfully with `failed == 0`, while reporting `traced_requests == 0` and `missing_trace_count == completed`. This function then marks the row as passed because it checks only `failed == 0`.

Partial coverage has the same problem: phase and decode-step metrics are aggregated over only the matched traces, so non-empty batching metrics can be reported from a biased subset while the row remains green.

There is no downstream fail-closed check either: `summarize_existing` applies the same rule, and the manifest/regression output preserves the row as passed.

This doesn't affect the run recorded here (it shows `missing_trace=0`), but a future rerun with missing traces would come out green. Would it make sense to require `traced_requests == completed` and `missing_trace_count == 0` for the row to pass, and apply the same rule in `summarize_existing`?

docs/models/deepseek-v2-lite/status.md
CAICAIIs: Thanks, this was a real mismatch. I updated the retained wording and artifact contract so the HTTP evidence now uses `prompt_words=64` / `max_tokens=64` explicitly instead of `64/64`, and it calls out that prompt words are workload-generator input rather than token count.

I also replaced the earlier 8-request text with the retained 32-request contract. The tracked snapshot now records 24 cells / 768 requests across host-staged + NCCL, and the NCCL paired rows are documented as 18 cells / 576 requests. The public status page now keeps this bounded to short fixed-shape HTTP A/B only, with no production, broad-workload, or vLLM-parity claim.

scripts/bench_dsv2lite_vllm_matrix.py
CAICAIIs: Agreed. I made the trace row fail closed now.

`run_openinfer_trace_pass` and `summarize_existing` both route trace artifacts through the same strict validation. A cell now requires the expected workload, full request count, zero failures/timeouts, `traced_requests == expected`, `missing_trace_count == 0`, full output-hash coverage, positive prompt/completion-token coverage, phase attribution, and valid singleton/batched decode-step accounting. Missing or partial trace coverage marks the cell failed instead of preserving a green row.

I also added regression tests for missing traces, partial trace/output coverage, workload drift, invalid prompt/completion token coverage, incomplete decode-step breakdowns, and partial HTTP completion. One additional fix from review: HTTP cells now require `completed == expected num_prompts`, so a 31/32 cell with 31 outputs cannot pass or enter perf medians.

openinfer-kernels/src/ops/deepseek_v2_lite.rs
```diff
@@ -10,6 +10,96 @@ pub struct Dsv2LiteRouterOutput<'a> {
...
 }
 
+pub fn dsv2_lite_router_logits_into(
+    ctx: &DeviceContext,
+    hidden: &HiddenStates,
+    gate_weight: &DeviceMatrix,
+    logits: &mut CudaSlice<f32>,
+) -> Result<()> {
+    ensure!(
+        hidden.hidden_dim == gate_weight.cols,
```
FeathBow, line 67: Thanks — the earlier workload-contract and trace fail-closed fixes both landed cleanly. One soundness item on the two new router ops added here. `dsv2_lite_router_logits_into` validates `hidden.hidden_dim == gate_weight.cols` and the `logits` output capacity, but never checks the backing lengths (`hidden.data.len()` vs `seq_len * hidden_dim`, or `gate_weight.data.len()` vs `rows * cols`). Since `HiddenStates` / `DeviceMatrix` / `HiddenStatesRef` all have public metadata fields (`tensor.rs` L382 / L458 / L659), safe code can allocate one hidden row, set `seq_len = 2` (or larger), pass an honestly-sized `logits`, and every check passes — but `router_logits_kernel` then reads rows that were never allocated.

Measured at this head on an sm_89 GPU: a one-row hidden buffer with `seq_len` inflated to 4096 returns `Ok(())` (async launch, so the bare run is silent), and compute-sanitizer reports 957 × `Invalid __global__ read` in `router_logits_kernel`, with the host stack going straight through `dsv2_lite_router_logits_into`. The sibling `dsv2_lite_accumulate_route_row_into` has the same shape — it checks `source_row < rows.seq_len` (the mutable metadata) but not `rows.data.len()`, then does raw `.add(source_offset)`; the same probe yields `Invalid __global__ read` in `accumulate_bf16_token_scaled_to_f32_kernel`. In the same run, the honest-shape control (`device_router_logits_match_host_accumulation_bitwise`) passes, so this is specifically the inflated-metadata path.

Today's callers construct honest metadata so nothing crashes in production — this is about the safe-API contract (a non-`unsafe pub fn` shouldn't be able to cause UB). Would it make sense to add a capacity check (`data.len() >= hidden_dim.checked_mul(seq_len)?`, and the gate equivalent), take the source row through a bounded `slice` in the accumulate op, and use `i32::try_from` for the FFI dimension casts (L43-45 / L92-94) instead of `as i32`?

FeathBow:
LGTM. Thanks for the quick fix!

PR #454, opened 2026-06-24
feat(qwen3): EAGLE-3 Chain Speculative decoding

scatyf3:
This is an PR for issue #443 

### step1: Chain draft
1. Draft model loading
	1. Model Loader ✅
	2. Memory Allocator ✅
2. Draft model farward pass ✅
	1. Chain(for testing) ✅
3. Target model forward pass: Target feature capture for draft injection  ✅
4. Executor, wire eagle spec decoding to current steps  ✅
5. Reserve memory for both target and draft  ✅
6. Frontend: add EAGLE3 args, wire to existing generation pipeline  ✅

### Change for step 1

Kernel:
1. rope without qk norm `eagle3_rope_kernel`
2. prefill over contiguous KV `single_prefill_nhd_noncausal_cuda`
3. decode over contiguous KV

Model
1. Draft model loader `openinfer-qwen3-4b/src/eagle3/loading.rs`
2. Memory Allocator `openinfer-qwen3-4b/src/eagle3/reservation.rs`
3. Draft Forward pass `openinfer-qwen3-4b/src/eagle3/forward.rs`

correctness check: EAGLE3 python golden gate `eagle3_drafter_golden_gate` from `tools/accuracy/dump_qwen3_4b_eagle3_golden.py`

SD Loop
1. reuse dflash's target feature capture `capture_layer_ids` in `process_all_layers_batch_multi`
2. Draft Executor `openinfer-qwen3-4b/src/executor/eagle3_lane.rs`, maintain draft weight,each request's buffer, state, tmp captured_hidden
3. Target Executor: 
	1. add `execute_eagle3_verify` in `openinfer-qwen3-4b/src/executor.rs`
	2. add eagle3 to  `SpeculativeVerify` and `SpeculativeDraft` in `execute_step_on_lane`
4. For scheduling and sampling, we reuse the dflash util

Correctness check: `openinfer-qwen3-4b/tests/eagle3_speculative_gate.rs`
``
Perf:`openinfer-qwen3-4b/tests/eagle3_speculative_perf.rs` and `openinfer-qwen3-4b/tests/eagle3_gsm8k_perf.rs`

Result

chain sd, k=3, max new=256, rtx5070ti, greedy, 10 prompt from GSM8k

| engine | spec OFF | spec ON   | **speedup** |
| --------- | -------- | --------- | ----------- |
| openinfer | 83.9     | **108.7** | **1.30×**   |
| vLLM 0.21 | 81.9     | 101.1     | 1.23×       |

[...]

Issue #443, opened 2026-06-23
[Roadmap] Speculative decoding (qwen3 first, shared primitives)


xiaguan:
Consolidating the speculative-decoding threads into one place so we coordinate instead of colliding. Grew out of #325. cc @wjinxu @scatyf3 @kitty-eu-org @CAICAIIs — please holler on / claim the lanes you want.

## Where we are

DFlash speculative decoding shipped on `main` (#436 + #442): the full loop — draft model loading → batched draft forward → target "verify" forward (piecewise CUDA graph) → acceptance → KV rollback → OpenAI serving — behind `--dflash-draft-model-path`. Greedy-lossless (lm-eval gsm8k strict-match identical spec on/off), validated on 5070 Ti / 5090, matches or beats vLLM dflash across c1/c8/c16. Design notes: `docs/models/qwen3/dflash-speculative-decoding.md`.

So speculation is a real subsystem now, not an experiment. This roadmap is about growing it into a first-class, multi-method, multi-model one — without each method/model reimplementing the controller.

## Workstreams

Brackets = current lead / interested. Shout if you want to take or share one.

### Methods
- **DFlash** [@xiaguan] — shipped. Next: draft-side piecewise CUDA graph (the remaining ~16% of the decode launch gap).
- **EAGLE / EAGLE3** [@scatyf3, #325] — orthogonal draft method. We deliberately deferred the proposer-trait abstraction to land EAGLE cleanly, so this is the natural driver for the shared proposer interface below.
- **n-gram / prompt-lookup** [@wjinxu, #349] — no draft weights; great for surfacing exactly what the scheduler / paged KV / CUDA graph / sampling actually need from a speculator.
- **Qwen3.5 DFlash** [@CAICAIIs, #434] — extend DFlash to the qwen3.5 line.

### Performance
- Draft-side piecewise CUDA graph. @kitty-eu-org's #439 has draft-forward kernels (non-causal dense/ragged-batch prefill attention, K-only norm+RoPE) that look directly reusable here — if they beat what's on `main`, worth landing as standalone kernel PRs.

### Capability & correctness
[...]

PR #647, opened 2026-07-10
feat(kernels): custom-mask prefill/paged attention for EAGLE-3 drafter

scatyf3:
This PR is an exploration of issue #443  and a split from EAGLE3 sd pr #454 

Add the single-sequence custom-mask attention kernels the EAGLE-3 drafter forward needs (prefill_attention.cu / paged_attention.cu) with their FFI declarations and safe openinfer-kernels / openinfer-core op wrappers.

### changes

| Op (`ops::…_into`) | Kernel entry | Backend |
|---|---|---|
| `single_prefill_nhd_causal_into` | `single_prefill_nhd_causal_cuda` | **FlashInfer** — `SinglePrefillWithKVCacheDispatched<…, MaskMode::kCausal>` |
| `single_decode_nhd_into` | `single_decode_nhd_cuda` | **FlashInfer** — `SingleDecodeWithKVCacheDispatched` (structurally single-query) |
| `eagle3_rope_into` | `eagle3_rope_kernel` / `eagle3_rope_cuda` | Hand-written `__global__` — plain RoPE, no QK-norm 

especially for `eagle3_rope_cuda`, my selection of handwritten kernel is align with dflash's choice. A further optimization can change this 2 handwritten kernel to flashinfer version, but it is out of this PR's scoop.


openinfer-kernels/csrc/shared/paged_attention.cu
```diff
@@ -758,6 +758,135 @@ int single_prefill_nhd_noncausal_cuda_hd64(
   OPENINFER_FFI_GUARD_END(-1)
 }
 
+// Causal variant of single_prefill_nhd_noncausal_cuda: identical NHD token-major
+// layout (q/output [seq, q_dim], k/v [max_seq, kv_dim]) but a causal mask, so the
+// N query tokens at the tail of the cache attend only positions <= their own.
+// Used for EAGLE-3's teacher-forced prefill in one batched forward (query i at
+// absolute position kv_len - seq_len + i, FlashInfer's causal alignment).
+int single_prefill_nhd_causal_cuda(
```
FeathBow, line 766: Thanks for splitting these kernels out of #454 — keeping them aligned with the existing noncausal/dflash pair made this easy to follow. I'd like to add a measured data point to the earlier guard comment, since it affects both new entries (this one and `single_decode_nhd_cuda` at L843): unlike their siblings in this file (`single_prefill_cuda`, `single_prefill_nhd_noncausal_cuda`, `paged_attention_decode_cuda`, …) they aren't wrapped in `OPENINFER_FFI_GUARD_BEGIN` / `OPENINFER_FFI_GUARD_END(-1)`, and the FlashInfer dispatchers they call can throw — `DISPATCH_GQA_GROUP_SIZE` raises `flashinfer::Error` for any group size outside {1, 2, 3, 4, 8}.

I ran a quick check at this PR's head on an sm_89 GPU: driving the safe wrapper `single_decode_nhd_into` with 56 Q heads / 8 KV heads (group 7) took down the whole process (`fatal runtime error: Rust cannot catch foreign exceptions, aborting`, SIGABRT), while the guarded `paged_attention_decode_cuda` under the same condition cleanly returned the -1 sentinel plus a message — the exact contrast `tests/ffi_exception_catch.rs` was added to pin down ("Before the guard this aborted the whole process").

Would you consider wrapping both function bodies in the guard macros, and having the two new Rust wrappers append `crate::ops::ffi_exception_message(result)` to their bail messages like `single_prefill_nhd_noncausal_into` does, so the caught message actually surfaces? Happy to share the small repro test if that helps.

openinfer-kernels/csrc/shared/paged_attention.cu
```diff
@@ -758,6 +758,135 @@ int single_prefill_nhd_noncausal_cuda_hd64(
...
+    void*    v_cache,
+    int32_t  num_qo_heads,
+    int32_t  num_kv_heads,
+    int32_t  head_dim,
+    int32_t  kv_len,       // positions to attend: [0, kv_len)
+    int32_t  max_seq_len,  // allocated cache rows (validation parity with the *_nhd_* pair)
+    float    sm_scale,
+    void*    stream)
+{
+    if (q == nullptr || output == nullptr || k_cache == nullptr || v_cache == nullptr ||
```
FeathBow, line 860: This one is subtler — I only noticed it while testing the guard case, so sharing in case it's useful. The validation here (and the causal twin at L780) accepts any positive head counts, but FlashInfer's dispatch computes `group_size = num_qo_heads / num_kv_heads` with integer truncation, so a non-divisible pair never throws — it silently serves fewer heads and still returns `cudaSuccess`.


Measured at this PR's head on an sm_89 GPU with 33 Q heads / 8 KV heads (truncates to group 4), output pre-poisoned with NaN and K/V zeroed:

- `single_decode_nhd_into` returns `Ok`, writes heads 0–31, and head 32 keeps all 128 poison values;
- `single_prefill_nhd_causal_into` (4 tokens, kv_len 8) returns `Ok`, and the surviving poison shows exactly 128 contiguous head-rows written from the buffer start (4 tokens × 32 heads, i.e. a 32-head stride) — every token after the first lands at a shifted offset, so the prefill corruption isn't limited to the missing head.

Because nothing throws, the FFI guard alone wouldn't cover this case. It may well be that the EAGLE-3 configs this PR targets always have divisible head counts — but since these are generic entry points, would it make sense to reject `num_qo_heads % num_kv_heads != 0` in these checks (or assert it in the Rust wrappers next to the existing shape asserts, `openinfer-kernels/src/ops/attention.rs` L941 / L990)? The same gap exists in the pre-existing siblings, so extending it there is probably follow-up material.

openinfer-kernels/src/ops/attention.rs
```diff
@@ -779,6 +779,62 @@ pub fn dflash_qk_norm_rope_into(
...
+            cos_ptr as *const ffi::Half,
+            sin_ptr as *const ffi::Half,
+            num_q_heads as i32,
+            num_kv_heads as i32,
+            head_dim as i32,
+            q_seq_len as i32,
+            k.seq_len as i32,
+            q_start_pos as i32,
+            k_start_pos as i32,
+            (cos_cache.data.len() / head_dim) as i32,
```
FeathBow, line 878: One small soundness note: `cos_max_pos` is derived from `cos_cache` alone, and nothing checks `sin_cache` against it — neither here nor in `eagle3_rope_cuda`'s host validation. Since this is a safe `pub fn` (re-exported via `openinfer-core::ops`) and the kernel indexes both caches with the same `pos * head_dim + d`, two individually valid buffers of different lengths can reach the GPU.

I checked at this PR's head on an sm_89 GPU: cos sized for 100 positions, sin for 1, rotating at position 99 — the call returns `Ok` and a bare run is completely silent, but compute-sanitizer reports 2054 × `Invalid __global__ read of size 2` in `eagle3_rope_kernel`, with the host stack going through `eagle3_rope_into`.

Would an `assert_eq!(sin_cache.data.len(), cos_cache.data.len())` next to the existing asserts make sense? For what it's worth, `dflash_qk_norm_rope_into` shares the same latent gap, so the assert may belong in both — totally fine as a follow-up, I'm just mentioning it for context.

openinfer-kernels/src/ops/attention.rs
```diff
@@ -858,6 +924,133 @@ pub fn single_prefill_nhd_noncausal_into(
...
+    assert_eq!(k_cache.hidden_dim, num_kv_heads * head_dim);
+    assert_eq!(v_cache.hidden_dim, k_cache.hidden_dim);
+    assert_eq!(v_cache.seq_len, k_cache.seq_len);
+    assert!(kv_len <= k_cache.seq_len);
+    assert!(
+        q_seq_len <= kv_len,
+        "causal prefill q_seq_len {q_seq_len} exceeds kv_len {kv_len}"
+    );
+    assert!(
+        row_offset + q_seq_len <= q.seq_len,
```
FeathBow: Thanks for the quick turnaround on the guards — the group-divisibility check and the sin/cos assert both look good. One more soundness item on the row-offset path, which the divisibility fix doesn't cover: `row_offset + q_seq_len <= q.seq_len` here (and in `eagle3_rope_into` at L802), plus the row-to-byte offset calculations at L1022/L820 and the subsequent pointer additions, use unchecked `usize` arithmetic. The workspace does not override Cargo's release default (`overflow-checks = false`), so these operations wrap silently.

Measured at this head on an sm_89 GPU: `single_prefill_nhd_causal_into` with `row_offset = usize::MAX` and `q_seq_len = 1` wraps the sum to 0, passes the assert, and wraps the Q/output device pointers to 8192 bytes before their allocations. The call returns `Ok(())`; because the launch is asynchronous, the failure surfaces only at the next synchronization as `CUDA_ERROR_ILLEGAL_ADDRESS`. `compute-sanitizer` reports 262 memory-access errors, with the displayed samples being invalid `__global__` reads. The error is therefore reported at a later synchronization point rather than at the offending wrapper boundary.

Would it make sense to compute the row and element bounds with `checked_add` / `checked_mul`, returning an `Err` on overflow, and then derive the pointers from bounded `CudaSlice::try_slice` / `try_slice_mut` views? The pre-existing `dflash_qk_norm_rope_into` and `single_prefill_nhd_noncausal_into` share this pattern; those may be follow-up material, but the two new safe wrappers should reject the invalid range before entering FFI.

openinfer-kernels/src/ops/attention.rs
```diff
@@ -858,6 +924,133 @@ pub fn single_prefill_nhd_noncausal_into(
...
+    q: &HiddenStates,
+    k_cache: &HiddenStates,
+    v_cache: &HiddenStates,
+    output: &mut HiddenStates,
+    num_q_heads: usize,
+    num_kv_heads: usize,
+    head_dim: usize,
+    kv_len: usize,
+) -> Result<()> {
+    assert_eq!(
```
FeathBow, line 986: Related soundness note on the same wrappers: the shape checks validate the public metadata against the arguments and one another, but never against the backing `data.len()`. Since all fields of `HiddenStates` are public (`tensor.rs` L459–461), safe code can allocate a small buffer and then inflate its logical shape before calling these safe wrappers.

Measured at this head on an sm_89 GPU: with K/V each backed by only 4 rows, both `seq_len` values changed to 4096, and `kv_len = 4096`, `single_decode_nhd_into` returns `Ok(())`; even the post-call sync returns `Ok(())`. The run is silent without instrumentation, while `compute-sanitizer` reports 518 invalid `__global__` reads. In this run there is no `ILLEGAL_ADDRESS` for the caller to catch.

One thing to watch in the fix: exact equality between `data.len()` and `hidden_dim * seq_len` would be incompatible with the established capacity-backed `HiddenStates` semantics. For example, `batch_decode_buffers.rs` allocates tensors at `max_batch_size` and rewrites `.seq_len` to the active batch size on every step. A capacity check seems more faithful: compute the required logical extent with `checked_mul`, then require `data.len() >= required` across Q/K/V/output as applicable.

Making the metadata private could encode this invariant more robustly, but would require replacing existing direct `.seq_len` writes with a checked reshape/view API, so that seems broader than this PR.

scatyf3:
Thank you for the incredibly thorough review, @FeathBow. Two new issue you mentioned are now handled:

1. **Prevent offset-arithmetic overflow**: add `checked_row_offset` (checked `offset+span` and byte offset) to the two row-offset wrappers `eagle3_rope_into` and `single_prefill_nhd_causal_into`, returning `Err` instead of wrapping a device pointer past its allocation (which only surfaced later as `CUDA_ERROR_ILLEGAL_ADDRESS`).
2. **Validate backing capacity**: add `ensure_hidden_capacity` (require `data.len() >= hidden_dim * seq_len`) across the Q/K/V/output buffers of all three new wrappers (`eagle3_rope_into`, `single_decode_nhd_into`, `single_prefill_nhd_causal_into`), so a safe caller can't inflate `.seq_len` past a small allocation into silent OOB reads.

FeathBow:
LGTM. Appreciate the fast turnaround.

scatyf3:
Also thank your quick review 👍

Issue #492, opened 2026-07-02
[Roadmap] 2026-H2


xiaguan:
> **TL;DR:** Second-half 2026 plan. Supersedes the five-workstream roadmap issue (#203), most of which has landed (prefix cache, KV offload, speculative decoding) or drifted. Priorities: make the website the real product surface (recipes-style per-model pages, CLI reference, hardware matrix), formalize model maturity tiers with Qwen3-4B as the first Stable line, and finally wire observability. GLM5.2 large-MoE bring-up is the active mainline and doubles as the boundary test for shared MoE infrastructure. P/D disaggregation (NIXL-compatible) is design-first, Later.
>
> `direction.md` (why per-model engines, where the shared boundary sits) and `execution.md` (what is in flight this week) remain the day-to-day source of truth. This file is the half-year view above them. Entries carry **Now / Next / Later** and move or get deleted as reality changes. No dates.

## Who this is for

openinfer's near-term market is narrow and deliberate: **small-to-mid models, single node, coding-agent backend, fast startup, small footprint**. That matches both our measured strengths (771 MB binary vs multi-GB images; ~3 s vs ~70 s cold start; Qwen3-4B beating vLLM across the QPS sweep on consumer Blackwell) and our contributor base, which is individual developers with one good consumer GPU. The large-MoE / multi-node story is being built right now through the GLM5.2 mainline — but until that line ships, we position the project by the individual-developer lane.

---

### 1. Website as the product surface

Today the site has three pages of real content while the engine's actual capabilities live scattered in repo docs. The website (openinfer-project/website) becomes the canonical user-facing surface, in the style of [recipes.vllm.ai](https://recipes.vllm.ai/): one verified, copy-pasteable recipe per model line, kept current as flags change.

- **Per-model recipe pages.** For each served line: exact launch command, required hardware, the CLI flags that matter for that model (`--kv-offload`, `--dflash-draft-model-path`, TP flags, …), expected TTFT/TPOT on the tested GPUs, and a benchmark snapshot. Only Qwen3-4B has a page today; Qwen3.5 and DeepSeek-V2-Lite are next. Every command on a page is run before it is committed — same rule as repo docs.
- **CLI / server reference.** A maintained page for `openinfer-server` args and env vars (`OPENINFER_CUDA_SM`, `OPENINFER_TRITON_PYTHON`, …). This must not rot: once the page exists, a flag change without a website update is an incomplete PR.
[...]

Issue #584, opened 2026-07-06
frontend: empty completion (suppressed first-token EOS) returns the whole prompt as choices[0].text


xiaguan:
Observed while validating #583 on GLM5.2 (8×H200), but the mechanism is engine-independent.

## Repro

```
curl http://HOST:8300/v1/completions -d '{"model": "...", "prompt": "<'The quick brown fox jumps over the lazy dog. ' x 700>", "max_tokens": 32, "temperature": 0}'
```

GLM5.2 greedily emits EOS as the first generated token on this degenerate prompt; the engine suppresses it (counted, not emitted) and finishes with `stop` after zero emitted tokens. The response then contains the **entire 7001-token prompt** as `choices[0].text`:

```
usage {'prompt_tokens': 7001, 'total_tokens': 7001, 'completion_tokens': 0} finish stop
text == prompt (6300 words), no echo requested
```

## Where

`vllm-server`'s non-stream collector (upstream rev 489abadf, `rust/src/server/src/routes/openai/completions.rs` ~line 173) only prepends the prompt when `echo` is set — we send `echo: false` and GLM5.2 rejects `echo: true` at validation. So `collected.text` itself equals the prompt while `collected.token_ids` is empty: the leak is in the detokenizer/collector seam (likely the incremental detokenizer primed with prompt tokens flushing its buffer on finish with zero generated tokens), either upstream or in `openinfer-vllm-frontend`'s bridge. Needs isolation before deciding to fix here or upstream (precedent: #294 → vllm#45286).

## Acceptance

A zero-emitted-token completion returns `text: ""` with `completion_tokens: 0` (or counts the suppressed EOS consistently with the engine contract), for both stream and non-stream paths.

xiaguan:
Upstream fix proposed in vllm-project/vllm#47707 — `DecodeStream::flush()` now calls `seed_prefix()` before decoding when no `push_token()` has run, so zero-generated-token completions return empty text instead of leaking the prompt. Once merged and bumped here, this resolves the detokenizer-seam root cause described above.

PR #653, opened 2026-07-11
feat(sim): scripted-token completion mode + tool-call round-trip test

scatyf3:
## What & why

Closes part of #492 (2026-H2 roadmap): a **real tool-call round-trip** integration
test through the OpenAI `/v1/chat/completions` frontend, driven entirely on CPU via
`openinfer-sim` — no CUDA, weights, or model crate. Streaming and non-streaming both
covered.

## 1. Deterministic replay in the sim engine (`openinfer-sim/src/lib.rs`)

The simulated engine could only cycle prompt tokens; a round-trip test needs to emit a
*specific* byte stream. Added a scripted-completion mode:

- `SimulatedEngineConfig::with_scripted_completion(ids)` — a token-id sequence replayed
  verbatim. Empty (the default) keeps the old prompt-cycling behavior byte-for-byte, so
  no existing test changes.
- `run_simulated_request` replays it bounded by `max_tokens`:
  - fully replayed (`script.len() <= max_tokens`) → `FinishReason::Stop`
  - truncated (`script.len() > max_tokens`) → emit the leading `max_tokens` ids →
    `FinishReason::Length`
- Unit tests: `scripted_completion_replays_ids_and_stops`,
  `scripted_completion_truncated_by_max_tokens_is_length`.

## 2. Integration test (`openinfer-sim/tests/tool_call_roundtrip.rs`)

`ToolCallSimServer` wires the scripted engine behind the real vLLM/OpenAI frontend. A
fake model dir supplies just enough metadata, no weights:

- `fuse_tokenizer_json()` — a tiny `WordLevel` + `Fuse` tokenizer whose vocab entries
  *are* the tool-call chunks, so replaying ids `[1, 2, 3]` detokenizes to exactly:

  ```
  <tool_call>
  {"name": "get_weather", "arguments": {"location": "SF"}}
  </tool_call>
  ```

  Vocab: `<unk>`=0, CHUNK_1=1, CHUNK_2=2, CHUNK_3=3, `<think>`=4, `</think>`=5,
  `<|endoftext|>`(EOS)=6.

Both tests emit the **same** 3-chunk script; they differ only in response mode:

- `non_streaming_tool_call_round_trips_to_openai_tool_calls` — the frontend assembles the
  full response; assert `message.tool_calls[0]` is `get_weather` with `{"location":"SF"}`
  (`assert_tool_call_is_get_weather_sf`).
- `streaming_tool_call_reassembles_across_chunks` — the same payload delivered as
  incremental SSE deltas. The chunk boundaries fall *inside* the JSON (`argu|ments`,
  `S|F`), so no single delta is valid JSON; the test reassembles the tool call across
  ≥2 deltas and asserts it parses to the same result.

## Relationship to #630

Complementary, not overlapping. #630 adds generic chat-completions E2E (response shape,
role/content, usage, SSE `[DONE]`). This adds the tool-call round-trip and the engine
change it requires. Different files (no `frontend_e2e.rs` / `lib.rs` collision), so the
two land in any order. Both independently rediscovered the tiny fixture needs a
`chat_template`.

## Test plan

- [x] `cargo test --release -p openinfer-sim` — lib 5, frontend_e2e 9,
  tool_call_roundtrip 3; all pass.
- [x] Based on latest `main`.

## Type of Change

- [x] Test coverage
- [x] New feature (sim scripted-completion mode)

### DIFF
--- a/openinfer-sim/src/lib.rs
+++ b/openinfer-sim/src/lib.rs
@@ -47,6 +47,7 @@ impl SimulatedEngineConfig {
     }
 
     /// Replay `ids` verbatim as the completion for every request
+    #[must_use]
     pub fn with_scripted_completion(mut self, ids: Vec<u32>) -> Self {
         self.scripted_completion = ids;
         self
--- a/openinfer-sim/src/lib.rs
+++ b/openinfer-sim/src/lib.rs
@@ -6,12 +6,15 @@ use openinfer_engine::engine::{
 };
 use tokio::sync::mpsc;
 
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Debug)]
 pub struct SimulatedEngineConfig {
     base_ttft_ms: f64,
     prefill_tokens_per_ms: f64,
     tpot_ms: f64,
     fallback_token_id: u32,
+    /// Explicit completion token-id sequence to replay verbatim. Empty (the
+    /// default) keeps the legacy behaviour of cycling the prompt tokens.
+    scripted_completion: Vec<u32>,
 }
 
 impl SimulatedEngineConfig {
@@ -39,7 +42,14 @@ impl SimulatedEngineConfig {
             prefill_tokens_per_ms,
             tpot_ms,
             fallback_token_id,
+            scripted_completion: Vec::new(),
         })
+    }
+
+    /// Replay `ids` verbatim as the completion for every request
+    pub fn with_scripted_completion(mut self, ids: Vec<u32>) -> Self {
+        self.scripted_completion = ids;
+        self
     }
 
     fn ttft(&self, prompt_tokens: usize) -> Duration {
@@ -58,6 +68,7 @@ impl Default for SimulatedEngineConfig {
             prefill_tokens_per_ms: 100.0,
             tpot_ms: 12.0,
             fallback_token_id: 0,
+            scripted_completion: Vec::new(),
         }
     }
 }
@@ -66,7 +77,7 @@ pub fn start_engine(config: SimulatedEngineConfig) -> EngineHandle {
     let (submit_tx, mut submit_rx) = mpsc::unbounded_channel();
     tokio::spawn(async move {
         while let Some(req) = submit_rx.recv().await {
-            tokio::spawn(run_simulated_request(req, config));
+            tokio::spawn(run_simulated_request(req, config.clone()));
         }
     });
     EngineHandle::new(submit_tx)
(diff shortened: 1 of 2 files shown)


xiaguan:
Walked through this locally (all 12 sim tests green). One real finding under the trailing EOS pad, plus a doc-comment correction request.

## The trailing EOS in the script masks a real contract gap

The script `[1, 2, 3, 6]` only round-trips because it encodes the **vLLM engine contract** (stop token emitted as the last token id), which is *not* what openinfer engines do today:

- Every openinfer engine suppresses EOS: on a stop token it sends `Finished(Stop)` and never emits the EOS as a `TokenEvent::Token` — qwen3 `scheduler/resolve.rs:200-205` (decode), `:69-77` (speculative), qwen35-4b `scheduler.rs:662-675`, same pattern elsewhere.
- Upstream `vllm-text` (`text/src/output/decoded.rs`, `suppress_terminal_stop_token`) unconditionally drops the **last token id of the Stop-carrying output** from visible text — "Match Python V1 token-stop detokenization". It assumes the stop token is that last id.
- The bridge coalesces queued `Token` + `Finished` events into one output (`bridge.rs`, `dispatch_burst` / "a trailing terminal rides the same output").

So with the engine-shaped stream (no EOS token), the last *content* token's text gets eaten. Repro inside this very test: change the script to `[1, 2, 3]` and `non_streaming_tool_call_round_trips_to_openai_tool_calls` fails with

```
tool call arguments are not valid JSON: "{\"location\": \"S"
```

`</tool_call>` (id 3) is dropped by the upstream stop-token suppression. That's the token-stream shape production engines actually produce, which means:

- **speculative decode** (`EmitManyAndFinish(Stop)` sends tokens + Finished back-to-back → same burst): last accepted content token's text is lost near-deterministically on EOS stops;
- **single-token decode**: same loss whenever the bridge lags one step and the last token coalesces with `Finished(Stop)` — timing-dependent.

Same root cause as #584 ("suppresses it (counted, not emitted)"); this is the non-zero-token symptom of that contract gap.

## Suggested direction

Align the engines with the vLLM contract: emit the stop token as the final token id and let the frontend suppress it. `completion_tokens` already counts the suppressed EOS, so the metadata contract is already halfway there. Once that lands, this test's `[1, 2, 3, 6]` script becomes exactly the engine-shaped stream — no change needed here. I'd track the engine-side fix in #584 or a sibling issue rather than in this PR.

## Ask for this PR

The two comments attributing the EOS pad to the incremental detokenizer are a misdiagnosis — upstream *does* flush held-back text on finish (`decoded.rs`, `decoder.flush()` in the finish branch). The token isn't held back; it's actively dropped by `suppress_terminal_stop_token`. Please reword:

- the file-level doc ("the incremental detokenizer holds the final token back"),
- the "Trailing EOS terminator" comment at the vocab entry / script construction,

to say the EOS pad exists because a Stop-finished output has its last token id suppressed by the upstream detokenize layer, so the suppressed token must be the EOS itself (matching the vLLM engine contract, which openinfer engines don't implement yet — see #584). Otherwise the next reader "fixes" the pad away and gets a green-looking flake instead of the real story.

scatyf3:
Thanks for the review! I've added a comment explaining this contract gap, and reworded the module doc and the inline comment to attribute the trailing EOS to the upstream terminal-stop-token suppression (rather than incremental detokenizer buffering), with a reference to #584 for the engine-side alignment.

PR #662, opened 2026-07-12
feat(eagle3): drafter config, weight loading, and memory sizing

scatyf3:
This PR is an exploration of issue https://github.com/openinfer-project/openinfer/issues/443 and a split from EAGLE3 sd pr https://github.com/openinfer-project/openinfer/pull/454 

## What's new

This PR implement a full EAGLE model loader:

1. utils in `openinfer-core/src/weight_loader.rs`: `load_tensor_i64_host` and `load_tensor_bool_host` for eagle and target model vocab convert
2. EAGLE3 Config loader in `openinfer-qwen3/src/config.rs`: `Eagle3Config`
3. EAGLE3 Model loader in `openinfer-qwen3/src/eagle3.rs` and `openinfer-qwen3/src/loading.rs`
4. memory allocator in `openinfer-qwen3/src/eagle3/reservation.rs`,  allocate GPU memory for EAGLE's weight/kv cache 

## Test

Correctness is tested by: 

1. `eagle3_reservation_pins_geometry`, where it asserts
	1.  per-token KV bytes is 1  layer × 2 (K+V) × (num_kv_heads 8 × head_dim 128) × 2B(bf16) = 4096
	2. weights consumed memory is in ∈ [470MB, 500MB)
	3. `r.fixed_bytes > weights_only`, which garentee prefill_scratch + decode_scratch + capture_scratch is count in memory allocatoin
2. `aux_layers_reject_tiny_target` is align with EAGLE official implementation( [EAGLE/eagle/model/modeling\_qwen3\_kv.py at main · SafeAILab/EAGLE · GitHub](https://github.com/SafeAILab/EAGLE/blob/main/eagle/model/modeling_qwen3_kv.py) line 583)


openinfer-qwen3/src/eagle3.rs
```diff
@@ -0,0 +1,99 @@
...
+    );
+    Ok([low, mid, high])
+}
+
+#[cfg(test)]
+mod tests {
+    use super::aux_hidden_state_layers;
+
+    #[test]
+    fn aux_layers_qwen3_4b() {
```
xiaguan: the test is useless?

openinfer-qwen3/src/eagle3.rs
```diff
@@ -0,0 +1,99 @@
...
+
+/// Number of tokens the EAGLE-3 chain drafts per speculative round (γ; the verify
+/// span is `EAGLE3_CHAIN_LENGTH + 1`: the current token plus γ drafts). v1 is a
+/// fixed top-1 **chain** — *not* a tree. γ=3 is the measured chain optimum on a Qwen3-4B GSM8K A/B (RTX 5070 Ti)
+pub(crate) const EAGLE3_CHAIN_LENGTH: usize = 3;
+
+/// The three *target* layers (low/mid/high) whose post-layer hidden states EAGLE-3
+/// captures and fuses via `fc`.
+pub(crate) fn aux_hidden_state_layers(target_num_layers: usize) -> Result<[usize; 3]> {
+    let low = 2;
```
FeathBow: I think these IDs mix two different indexing contracts. SafeAILab’s Qwen3 implementation appends hidden_states before executing decoder layers 2 / N/2 / N-3; vLLM likewise checks idx before calling the layer and uses (2, N/2, N-3) as its defaults. Those tensors are therefore the outputs of layers 1 / N/2-1 / N-4.
 
OpenInfer’s capture API is explicitly post-layer: batch_prefill runs forward_layer_batch_paged first and then copies hidden when the same zero-based layer_idx matches. The verify path has the same ordering, and the split source in #454 passed this helper directly into those capture paths without an index conversion.

For the 36-layer Qwen3-4B target, the official features are therefore post-layer [1, 17, 32], while the current helper captures post-layer [2, 18, 33] — one block late for all three fc inputs. Could this return [1, N/2 - 1, N - 4] for the existing post-layer contract, or alternatively keep [2, N/2, N - 3] and add a true pre-layer capture path?

scatyf3: Openinfer's target model perform forward then hidden state capture, thus the feature inject layer index's meaning is differ from EAGLE3 source code(capture → forward). To simplify the code and align with vllm's **post layer hidden state capture(forward → capture ) ** ([vllm/vllm/model\_executor/models/llama.py at main · vllm-project/vllm · GitHub](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/llama.py) line 423), the better approach is **use shared hidden state capture util from current dflash code**. Thus our target hidden capture layer index should be `[1,17,32]` instead of `[2,18,33]`.

openinfer-qwen3/src/eagle3.rs
scatyf3: Yes this is a constant-true assertion. The intent was to align with DFlash's `downloaded_dflash_config_matches_qwen3_4b`, which pins the target feature-injection layers:
`assert_eq!(dflash.target_layer_ids, vec![1, 9, 17, 25, 33]);`. However, EAGLE3's  do not have `target_layer_ids` in config.json, so there's no external value to validate against, and asserting the formula's own output is just an echo. I'll delete that and let future draft golden gate check the end to end correctness.

openinfer-qwen3/src/eagle3/loading.rs
```diff
@@ -0,0 +1,144 @@
...
+            &weight_map,
+            "midlayer.self_attn.k_proj.weight",
+        )?;
+        let v_proj = load_tensor_2d(
+            ctx,
+            &shards,
+            &weight_map,
+            "midlayer.self_attn.v_proj.weight",
+        )?;
+        let q_dim = q_proj.rows;
```
FeathBow: validate_for_target validates the config geometry, but most checkpoint tensors are never checked against that config. Here q_dim / kv_dim are derived from Q/K, and DeviceMatrix::vstack only requires equal column counts. For example, Q [4096, 5120], K [1024, 5120], and V [2048, 5120] can pass this loader even though the configured V projection has only 1024 rows; later slicing still uses kv_dim = 1024.

Gate/up have the same shape, while O/down projections, all norms, and lm_head are also unchecked; fc is currently the only fully validated matrix. The published checkpoint has the expected geometry, but an incompatible or malformed artifact can still construct an internally inconsistent Eagle3DraftModel. Could the loader validate each tensor’s expected shape from Eagle3Config before stacking or returning the model?

openinfer-core/src/weight_loader.rs
```diff
@@ -267,6 +267,43 @@ pub fn load_tensor_1d_f32(
...
+/// Load a 1D I64 tensor into a host `Vec<i64>`.
+///
+/// For small integer lookup tables that live on the host (e.g. EAGLE-3's `d2t`
+/// draft→target vocab offset map), not weights destined for a GEMM.
+pub fn load_tensor_i64_host(
+    shards: &[SafeTensors],
+    weight_map: &HashMap<String, usize>,
+    name: &str,
+) -> Result<Vec<i64>> {
+    let tensor = find_tensor(shards, weight_map, name)?;
```
FeathBow, line 279: These typed loaders validate the raw byte length, but not the safetensors dtype or rank. An F64[32000] d2t has a byte length divisible by eight and is silently decoded as 32,000 i64 bit patterns; a U8[vocab] t2d is accepted with every non-zero byte mapped to true. Shapes such as [1, N] are also flattened even though both functions document a 1D tensor.

The EAGLE loader subsequently checks only the resulting vector lengths, so none of these mismatches is rejected later. Could these functions require Dtype::I64 / Dtype::BOOL and tensor.shape().len() == 1 before interpreting the data?

openinfer-qwen3/src/eagle3/reservation.rs
```diff
@@ -0,0 +1,141 @@
...
+        let midlayer = BF16
+            * ((q_dim + 2 * kv_dim) * (2 * hidden) // qkv_proj (2*hidden input cols)
+                + q_dim * hidden                    // o_proj
+                + (2 * inter) * hidden              // gate_up_proj
+                + inter * hidden                    // down_proj
+                + 3 * hidden); // input_layernorm + hidden_norm + post_attention_layernorm
+        let fc = BF16 * hidden * fc_in;
+        let head = BF16 * (config.draft_vocab_size * hidden + hidden); // lm_head + final norm
+        let weights = midlayer + fc + head;
+        let weights = weights + weights / 10;
```
FeathBow, line 66: The loader always allocates two BF16 RoPE caches of shape [max_position_embeddings, head_dim], but the reservation has no term depending on max_position_embeddings and relies on the 10% weight slack instead.

That happens to cover the published 40,960-position config: the caches use 20,971,520 bytes versus 43,649,024 bytes of slack. However, validation permits any draft limit greater than or equal to the target’s. With otherwise identical Qwen3-4B geometry and max_position_embeddings = 131072, validation passes and the caches use 67,108,864 bytes, exceeding the slack by 23,459,840 bytes.

Could fixed_bytes include the exact checked term 2 * max_position_embeddings * head_dim * sizeof(bf16) rather than treating a config-dependent allocation as generic weight slack?

scatyf3:
Hi @FeathBow , I already update my PR according to your coment

For weight shape check, previously I just check the tensor shape from config.json, now I add:
1. use tensor shape from config to check the weight we loaded
2. define `check2d` and `check1d`

The possible next step is convert  `check2d` and `check1d` to some shared util and check every model's weight using this. This could be a good first issue for beginner

For RoPE Cache reservation, I add rope allocation according to draft's `max_position_embeddings`
```rust
let rope = 2 * config.max_position_embeddings * config.head_dim * BF16;
```

For DType Check, I add dtype check in weight loader
```rust
if tensor.dtype() != Dtype::BOOL {
	return Err(anyhow::anyhow!(
		"BOOL tensor '{}': expected dtype BOOL, got {:?}",
		name,
		tensor.dtype()
	));
}
if tensor.shape().len() != 1 {
	return Err(anyhow::anyhow!(
		"BOOL tensor '{}': expected 1D, got shape {:?}",
		name,
		tensor.shape()
	));
}
```

FeathBow:
LGTM — thanks for addressing the review comments.

btw shared weight-shape validation could be useful, but it may need some scoping first since some loaders intentionally flatten or shard weights, while others already have model-specific validators. If you’d be willing to investigate the uncovered loaders and propose a bounded first slice, such as DFlash/DSpark, I’d be happy to mark it as a good first issue or follow-up work.

PR #664, opened 2026-07-13
perf(glm52): min-latency router logits GEMV replaces the cublas splitK plan

xiaguan:
First cut of the GB300 kernel-count campaign. The EP4 bucket-1 whole-step graph is 2,913 nodes (2,700 kernels) and solo decode is launch-bound at ~8.7µs per node — width-independent (EP4→EP32 p50 flat at ~23.3-23.6ms) and only 8% off the same weight-only chain on H200, so the lever is node count, not bandwidth.

**This PR: the router gate GEMM.** `cublasGemmEx` on the [256,6144]×[6144,tokens≤8] gate GEMM picks a splitK plan at decode row counts, costing 4 graph nodes per MoE layer (GEMM + splitKreduce + workspace alloc/free) — 300 of the 2,913 nodes. Replaced with the min-latency router GEMV from TensorRT-LLM's dsv3MinLatencyKernels (via SGLang sgl-kernel / vLLM, Apache-2.0): one block per expert row, f32 fixed-order accumulation (deterministic run-to-run, and per-row order is independent of the token-count instantiation — stronger than the M-dependent cublas plan), PDL launch attributes gated on cc≥9. The cublas path is deleted.

**Measured (GB300 tray, EP4, 96 steps):**

| bucket | main p50/p90/p99 | PR p50/p90/p99 | Δp50 |
|--------|------------------|----------------|------|
| 1 | 23.42 / 23.48 / 23.60 | **22.77** / 22.83 / 23.50 | −2.8% |
| 8 | 41.21 / 41.56 / 82.79 | **40.69** / 41.14 / 82.02 | −1.3% |

−650µs for −300 nodes ≈ 2.2µs/node, consistent with the launch-tax model. (The bucket-8 p99 ~2× bimodal tail is pre-existing and tracked separately.)

**Gates:**
- New device test `glm52_router_smoke`: logits vs f64 host reference (tol 5e-3), top-k selection exact against the GPU logits, all dispatch instantiations padded_tokens 1..=8, active<padded coverage. Green on sm_120 and sm_103.
- `layer_moe_ep4_oracle_gate` green on GB300 (g32/g16/g8/g4, 63/64 probes in-tol + 1 within the tie-flip allowance — expected from the changed GEMM accumulation order).

Toxic review: Request-Changes → fixed (test now covers the real bucket instantiations 1..8 instead of {3,16}; unreachable 9..16 cases deleted — `GLM52_MAX_BATCH_PER_RANK` pins the bound; PDL gated on cc≥9 mirroring argmax.cu; fence-free trigger rationale documented).

### DIFF
--- a/openinfer-kernels/csrc/glm52/glm52_router.cu
+++ b/openinfer-kernels/csrc/glm52/glm52_router.cu
@@ -1,10 +1,16 @@
 #include "../common.cuh"
 
-#include <atomic>
-
 #include <cuda.h>
 #include <cuda_runtime.h>
 #include <math_constants.h>
+
+// GLM5.2 is FP8 weights + FlashMLA: nothing below Hopper can run the model,
+// so a sub-sm_90 compilation target is a build misconfiguration, not a
+// support tier. Fail here instead of shipping kernels with the PDL calls
+// silently compiled out.
+#if defined(__CUDA_ARCH__) && (__CUDA_ARCH__ < 900)
+#error "GLM5.2 kernels require sm_90+ (check OPENINFER_CUDA_SM / detected GPU targets)"
+#endif
 
 namespace {
 
@@ -125,9 +131,7 @@ __global__ __launch_bounds__(128, 1) void glm52_router_logits_gemv_kernel(
   float acc[kNumTokens] = {};
   __shared__ float sm_reduction[kNumTokens][kNumWarps];
 
-#if (defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 900))
   cudaGridDependencySynchronize();
-#endif
 
   for (int ki = 0; ki < kIters; ++ki) {
     const int k_base = ki * kElemsPerIter + tid * kVpt;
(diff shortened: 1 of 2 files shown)


xiaguan:
Pushed 8060d02 per review discussion: sub-sm_90 compilation targets now fail the build with `#error` instead of silently compiling the PDL intrinsics out, and the cc<9 host-side fallback launch is deleted — GLM5.2 is FP8 + FlashMLA, nothing below Hopper can run the model, so that path was a support tier for hardware that cannot exist. Verified: sm_120 and sm_103 build + smoke green, sm_80 target fails with the directive; EP4 bucket-1 re-run at tip p50 22.71 ms (win holds).

xiaguan:
Measured the select-kernel thread-count question (512 was inherited, threads past n_experts only idled): ncu on GB300, grid(1) bucket-1 shape — 512 threads 10.11µs → 256 threads 10.02µs, e2e bucket-1 p50 22.68ms (flat). The kernel is bound by the per-thread serial rank-count loop (256 iterations over smem), not by the idle upper half, so this lands as hygiene, not perf. Kept at one thread per expert (c9aff6b).

Worth noting for a later PR: at ~10µs × 75 calls this select kernel is ~750µs/step (3.3%) — parallelizing the selection itself (warp-cooperative ranking instead of the per-thread serial loop) is a real candidate, but out of scope here.

PR #667, opened 2026-07-13
feat(qwen35): add transactional speculative verifier

CAICAIIs:
## Summary

This PR is the target-verifier slice split out of #626. It makes Qwen3.5 speculative verification reviewable independently from the draft model and serving integration.

It verifies `[current token, draft tokens...]` in one target forward pass, then commits the longest greedy-matching prefix plus one target token. Qwen3.5 hybrid state is handled as one transaction: paged KV, recurrent state, convolution state, and CUDA Graph slot state.

## What changed

- Adds batched target verification and reusable verification buffers.
- Adds commit, partial replay, and rollback for Qwen3.5 hybrid state.
- Uses checked projection calls so launch failures reach rollback.
- Covers accept-all, reject-first, prefix acceptance, KV page boundaries, injected failures, and replay of a graph captured before the transaction.
- Keeps the verifier on one CUDA stream and rejects stream overrides before mutation.

## Validation

On one RTX 5090 with Qwen3.5-4B:

- Qwen3.5 feature check and Clippy passed.
- Speculative verifier tests: 11 passed.
- HF golden gate: 2 passed, including the 4097/8192-token fixtures.
- Scheduler E2E: 1 passed.
- Local formatting and diff checks passed.

## Scope

This PR does not add the DFlash draft model, scheduler/server opt-in wiring, sampling support, or a performance claim. Those parts remain in follow-up PRs split from #626.


CAICAIIs:
Follow-up split plan:

- Drafter PR: add DFlash draft-model loading and draft-forward correctness gates on top of this verifier.
- Serving integration PR: wire the opt-in scheduler/server flow, explicit fallback and admission rules, then add same-host benchmark and profile evidence.

Each PR will keep the default Qwen3.5 path unchanged unless DFlash is enabled.

Issue #668, opened 2026-07-13
perf(glm52): GB300 decode roadmap — right-size masked MoE GEMM for small buckets (b1 at 29% roofline, ~4 ms/step on the table)


xiaguan:
## Context

Same-machine (4×GB300, one NVL72 tray) comparison against vLLM 0.25.0 on GLM-5.2-FP8, `vllm bench serve` random 1024/256, c1, plain decode (no MTP), `VLLM_USE_DEEP_GEMM=0`:

| topology | vLLM 0.25.0 ITL p50 | openinfer | gap |
|---|---|---|---|
| TP4+EP4 | 11.61 ms | — | — |
| DP4+EP4 (= our topology) | 19.60 ms | 22.62 ms | **1.15×** |
| DP8+EP8 (2 trays) | 20.81 ms | 23.61 ms | 1.13× |

Decomposition: DP→TP attention sharding is worth 1.69× at bs=1 (topology choice, separate discussion); the engine-side gap at equal topology is **15%, ≈3 ms/step**. nsys node-mode slices (locked 1965 MHz, EP4, 60 steps) show where it lives.

## Per-family GPU time, ms per rank-step

| family | b1 | b4 | b8 | b8/b1 |
|---|---|---|---|---|
| **moe_ep_wo_masked_mma** | **8.88** | **13.17** | **17.27** | 1.9× |
| fp8 weight-only GEMV (dense/MLA) | 5.59 | 8.12 | 7.57 | 1.4× |
| DeepEP dispatch/combine | 4.38 | 6.27 | 6.48 | 1.5× |
| attention (sparse MLA + combine) | 2.20 | 2.35 | 2.18 | 1.0× |
| norm family (fused-add + standalone) | 0.99 | 0.97 | 0.96 | 1.0× |
| router select | 0.82 | 0.72 | 0.76 | 0.9× |
| glue / quant / other | 1.00 | 1.10 | 1.17 | ~1.1× |

Expert bytes grow ~5.4× from b1→b8 but masked_mma time only 1.9×: the kernel carries a fixed shape tax that dominates small buckets.

## masked_mma roofline (per layer, W13+W2 pair)

| bucket | live experts/rank | ideal (bytes/8TB·s) | measured | roofline % | fixed tax |
|---|---|---|---|---|---|
| b1 | ~7.5 | ~36 µs | 122 µs | **29%** | ~86 µs |
| b4 | ~25 | ~119 µs | 181 µs | 66% | ~62 µs |
[...]

PR #671, opened 2026-07-13
perf(glm52): GB300 split-K table + fused reduce-SwiGLU for the batched GEMV chain

xiaguan:
## Summary

Right-sizes the GLM5.2 batched weight-only GEMV chain for GB300 (sm_103) EP4 decode:

1. **GB300 batch-8 split-K table** (`mma_config`): every whitelisted decode shape re-swept on GB300 over the full (KSPLIT × NTILES) grid with a locked-clock standalone harness. The H200 picks do not transfer (the wall is bytes-in-flight per SM, an arch property) — e.g. q_a moves 16/2 → 48/2 (8.1 → 7.1 µs pair), kv_a 16/2 → 16/1 (6.7 → 5.5 µs). Gated on `arch_is_blackwell()`; Hopper tables byte-for-byte untouched.
2. **Fused reduce-SwiGLU** (`glm52_gemv_reduce_silu_mul`): the packed gate|up mma path stops at its f32 k-slice partials and the SwiGLU absorbs the fixed-order reduce — bit-identical (both sums round to bf16 before the SiLU math), one launch instead of two, and the `[rows, 2*inter]` bf16 round-trip disappears. Wired through `fp8_mlp_into` (dense MLP + MoE shared expert).
3. A `mma_config` entry whose (ksplit, ntiles) has no instantiated dispatch case now fails the launch at capture instead of silently falling back to the register tile.

## Measured (4×GB300, EP4, locked clocks, `glm52_step_bench` 60 steps)

| config | b8 step p50 / p99 (ms) | b1 step p50 / p99 (ms) |
| --- | --- | --- |
| main | 40.68 / 41.6 | 22.62 / 22.7 |
| this PR | **40.03 / 41.31** | 22.57 / 22.68 |

Attribution: the split-K table is the bulk of the −0.65 ms; the SwiGLU fusion contributes ~−0.15 ms (per-site launch + bf16 round-trip removal). Correctness: dense-layer, EP4 full-layer, and bookend oracle gates green on GB300 at the perf commit (the follow-up commit only adds a host-side dispatch-drift guard, a Rust wrapper dedupe, and docs; a re-gate at the final SHA is queued and will be posted as a comment).

Hopper note: sm_90 keeps its measured tables, but bucket-4/8 graphs now route gate|up through the fused epilogue there too — bit-identical numerics and strictly fewer bytes, not re-benched on H200.

## Falsified along the way (documented in `docs/models/glm52/ep4-gb300.md`)

- reduce→RMSNorm fusion (q_a norm + post-attn add+norm absorbing the reduce): bit-identical impl, oracle green, but b8 regressed 40.03 → 40.84/40.72 — norm kernels are block-per-row (grid = 8 blocks), so the ksplit=16-48 reduction collapses onto 8 latency-bound blocks. Reverted; a fusion only wins when the consumer's grid can host the producer's parallelism.
- Five structural GEMV alternatives (smem staging, launch-bounds squeeze, cp.async, Triton split-K, DeepGEMM masked fp8) all measured slower or unavailable on SM100; a pure-read control at the shipping grid geometry reaches 6.4-6.8 TB/s, so the kernel is at its W8A16 structural optimum. Deeper levers are tracked in #668.

### DIFF
--- a/docs/models/glm52/ep4-gb300.md
+++ b/docs/models/glm52/ep4-gb300.md
@@ -1,6 +1,6 @@
 # GLM5.2 EP4 on GB300
 
-> **TL;DR:** DP4/EP4 (`--moe-topo ep4`) brings the EP high-throughput layout to 4×GB300: 64 whole routed experts per rank (~188 GiB weights/rank of 284 GiB), a second DeepEP shim instantiation (`glm52_ep4_deepep_*`), and a new **weight-only routed-expert chain** replacing the sm_90a-only DeepGEMM masked GEMMs — bf16 recv rows × on-the-fly fp8 dequant on `mma.m16n8k16`, running directly on the DeepEP aligned receive layout (no activation re-quant, no masked relayout, no output remap). Bring-up gates green 2026-07-12: 4-bucket whole-step graph capture, greedy byte-determinism ×2, 8-way concurrent greedy (per-shape identical), sampled fallback, **layer-6 EP4 oracle 63/64 × 4 buckets** (after fixing the FlashMLA sm100 UE8M0 KV-scale contract — see lessons doc — which also un-broke attention accuracy for every heads-64 GLM5.2 path on GB300). First numbers (untuned): bs=1 TPOT `23.1 ms`; **c=32 output `559 tok/s`, TPOT p50 `39.05 ms` / p99 `39.18 ms`**. Needs `EP_DISABLE_GIN=1` on NIC-less nodes.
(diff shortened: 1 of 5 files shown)


xiaguan:
Re-gate at the final SHA (ebfabd7) on 4×GB300, locked clocks, `glm52_step_bench` 60 steps:

| bucket | p50 / p90 / p99 (ms) |
| --- | --- |
| b8 | **40.03 / 40.69 / 41.18** |
| b1 | 22.58 / 22.65 / 24.84 |

b8 reproduces the perf-commit numbers exactly; b1 p50/p90 flat vs main (22.62), the single p99 outlier step is host noise (p90 22.65). Dense-layer, EP4 full-layer, and bookend oracle gates all green at this SHA.

PR #676, opened 2026-07-14
fix(qwen35): restrict token selection to the frontend-decodable vocab

### LINKED ISSUE #673: frontend: a sampled id in the padded-vocab region fails the whole request; HF decodes the same id to `''`
## Description

Once across 800+ random-prompt bench requests at Qwen3.5-9B (`ffb959c4`, GH200 sm_90): `ERROR vllm_server completions.rs:411 completion stream failed error=tokenizer error: decoding failed: unknown token ID: 248257`, followed by the engine auto-aborting the request on the dropped stream.

The gap is measurable: Qwen3.5-9B's config `vocab_size` (lm_head width) is 248320 while `tokenizer.json` defines ids only up to 248069 (transformers adds seven config-only audio/TTS placeholders up to 248076), so a couple hundred padded ids are legal argmax outputs with no tokenizer entry. Under random-token prompts the model can land there (padded rows are trained toward low probability, not zero). Emitting the id is model behavior, not the defect — HF's tokenizer decodes the same id to `''` and generation continues; openinfer turns it into `finish_reason="error"` for the whole request. Aggravator: vllm-bench counts that error-finish stream as successful, so the failure is also invisible in bench artifacts.

Low frequency and OOD-prompt-biased, but a legal inference step should never produce a failed request.

FeathBow:
## Description

Fixes #673 

Qwen3.5 checkpoints pad `vocab_size` past the tokenizer's serialized vocab: `tokenizer.json` across the line defines ids up to 248069 while `vocab_size` is 248320. Every id in the padded tail is a legal argmax output — the rows exist in the output projection — but the frontend cannot detokenize them: the stream fails with `decoding failed: unknown token ID` and the engine aborts the request. Observed once in 800+ random-prompt bench requests on Qwen3.5-9B. HF transformers decodes such ids to an empty string and continues, so this was an availability gap unique to our serving path.

At load, the selection width mirrors the pinned frontend's merge with its exact failure semantics: `tokenizer.json` vocab and added_tokens through the same typed added-token schema and u32 id contract (fatal on parse failure — the frontend cannot serve without it), plus `tokenizer_config.json` `added_tokens_decoder` as a whole-file typed parse where any invalid entry drops all decoder tokens with a warning and unparseable numeric keys are skipped per entry (248077 across the Qwen3.5 line vs the padded 248320). All three output-projection sites go through the checked row-range GEMM so launch failures keep flowing to the scheduler as `Result`s. The merged ids must form a dense prefix — a row-range bound cannot mask holes — so a sparse id space fails the load instead of silently truncating the model's output space; the published artifacts are dense 0..=248076. The output-projection GEMMs run over that row prefix via the existing checked row-range GEMM, and the logits arena plus sampling scratch are sized with the same width. Weights stay untouched, so tied and untied checkpoints take the same path with no extra memory or copies.

The parsed width is additionally validated against the checkpoint `vocab_size`; a missing `tokenizer_config.json` only shrinks the bound, which is the safe direction. Unit tests pin the merge arithmetic, the sparse-id-space failure, the all-or-nothing decoder drop, and the per-entry key skip.

### Semantics

The cutoff is the frontend-decodable vocab (248077 across the line), so nothing the frontend could render is masked — including the audio/TTS placeholder ids (248070–248076) that live only in `added_tokens_decoder`. Every generated special token sits far below the cutoff (`<tool_response>` 248066/67, `<think>`/`</think>` 248068/69). Prompt-side ids are unaffected: embeddings keep the full width. Teacher-forced logprobs shift by the padded tail's softmax mass, which is negligible for real text — the size-keyed golden gates stay green (below). A greedy divergence vs HF is only possible at steps where HF's top-1 lands in the padded tail, which is exactly the failure this PR removes.

### Evidence (GH200 120GB, aarch64 sm_90, single GPU)

- Load line on every size, tied and untied alike: `output projection: selection bounded to decodable vocab 248077 (checkpoint pads to 248320)`.
- Bounded-GEMM path validated on the final tree at 4B (tied) and 9B/27B (untied): `hf_golden_gate` green at all three sizes (4B mean 0.018–0.026 / max 0.190; 9B mean 0.020–0.023 / max 0.138; 27B bit-identical to the pre-change run across all five replay surfaces, mean 0.020–0.022 / max 0.206), `e2e_scheduler` / `chunked_prefill` / `sampling_behavior` green at 4B, serving probe returns coherent greedy output (`<think>` still generates) and a well-formed tool call with `unknown_token=0`.
- Perf anchors on the final tree within the tune band: 9B c=1 126.4 tok/s / TTFT p50 50.6 ms / TPOT p50 7.58 ms, qps=8 701 tok/s; 27B c=1 46.9 / c=8 192.6 / qps=8 323.8 / c=48 315.3 tok/s, error scans all-zero.
- Full 9B + 27B sweeps (c={1,4,8}, qps={1,2,4,8,10,12,16}, long-context in 4097, c=120 overload; 27B additionally c=32/48): server-log scan `unknown_token=0` everywhere; throughput/latency at previously-clean points within noise (27B within 1.5%; 9B within known cross-process GEMM-tune variance).
- HF greedy comparison (6 prompts): same match pattern as the pre-fix run (9B 5/6, 27B 4/6, flips at the same positions).
- Gates on the fixed tree: `hf_golden_gate` green at 9B (mean 0.018–0.024, max 0.122) and 27B (bit-identical stats to pre-fix); `e2e_scheduler`, `chunked_prefill`, `sampling_behavior` green at both sizes; config unit tests 3/3.

### DIFF
--- a/openinfer-core/src/ops.rs
+++ b/openinfer-core/src/ops.rs
@@ -32,13 +32,13 @@ pub use openinfer_kernels::ops::{
 #[cfg(not(feature = "kernel-call-trace"))]
 pub use openinfer_kernels::ops::{
     embedding_batch, fused_add_rms_norm_batch_into, gemm_into, gemm_rows_into,
-    gemm_token_range_into_checked, qk_norm_rope_batch_decode_into, rms_norm_batch_into,
-    silu_mul_fused_batch_into,
+    gemm_rows_into_checked, gemm_token_range_into_checked, qk_norm_rope_batch_decode_into,
+    rms_norm_batch_into, silu_mul_fused_batch_into,
 };
 pub use paged_plan::PrefillPagedPlan;
 #[cfg(feature = "kernel-call-trace")]
 pub use traced::{
     embedding_batch, fused_add_rms_norm_batch_into, gemm_into, gemm_rows_into,
-    gemm_token_range_into_checked, qk_norm_rope_batch_decode_into, rms_norm_batch_into,
-    silu_mul_fused_batch_into,
+    gemm_rows_into_checked, gemm_token_range_into_checked, qk_norm_rope_batch_decode_into,
+    rms_norm_batch_into, silu_mul_fused_batch_into,
 };
--- a/openinfer-core/src/ops/traced.rs
+++ b/openinfer-core/src/ops/traced.rs
@@ -54,6 +54,18 @@ pub fn gemm_rows_into(
     x: &HiddenStates,
     out: &mut HiddenStates,
 ) {
+    gemm_rows_into_checked(ctx, weight, row_offset, num_rows, x, out)
+        .expect("GEMM row-range launch failed");
+}
+
+pub fn gemm_rows_into_checked(
+    ctx: &DeviceContext,
+    weight: &DeviceMatrix,
+    row_offset: usize,
+    num_rows: usize,
+    x: &HiddenStates,
+    out: &mut HiddenStates,
+) -> Result<()> {
     if call_trace::is_enabled() {
         let label = call_trace::current_label("gemm_rows");
         call_trace::record_call(gemm_rows_call::<OutDim>(
@@ -65,7 +77,7 @@ pub fn gemm_rows_into(
             x.seq_len,
         ));
     }
-    openinfer_kernels::ops::gemm_rows_into(ctx, weight, row_offset, num_rows, x, out);
+    openinfer_kernels::ops::gemm_rows_into_checked(ctx, weight, row_offset, num_rows, x, out)
 }
 
 pub fn gemm_into(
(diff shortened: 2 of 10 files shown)


xiaguan (approved):
LGTM.

FeathBow:
The frontend side is being fixed upstream. vLLM's Rust tokenizer wrapper takes a byte-level fast path for GPT-2-style BPE (the Qwen families) that fails the whole decode on an undefined id, while its other two backends and HF's tokenizer skip it. https://github.com/vllm-project/vllm/pull/55240 makes the fast path skip undefined ids too, so a sampled padded id becomes an empty string instead of a dropped stream. Until that lands and our pin moves past it, this engine-side bound stays the only protection.

Issue #470, opened 2026-06-29
qwen35: measure mixed-load ITL after scheduler-level chunked prefill


CAICAIIs:
Parent: #249
Crate: `openinfer-qwen35-4b`
Related: #375, #313, #244
Reference: `docs/benchmarks/mixed-load-itl.md`

## Background

Qwen3.5 scheduler-level chunked prefill has landed, and the resumed/chunked `base_pos > 0` regression is closed. The next question is whether this actually improves the serving tail that matters: long cold prompts arriving while other requests are actively decoding.

Qwen3-4B already has a mixed-load ITL benchmark note showing how long prefills can freeze active decode and how to interpret ITL p99 as a frequency-and-severity problem. Qwen3.5 needs the same post-chunking evidence because its 24 linear-attention layers and recurrent state make the prefill tradeoff model-specific.

## Goal

Measure Qwen3.5 mixed-load behavior after scheduler-level chunked prefill and record whether chunking bounds ITL tail under long-prompt injection.

This issue should answer:

- how bad the active-decode stall is for cold long prompts;
- whether warm/repeated prompts behave differently once prefix reuse is still absent;
- whether chunked prefill changes ITL p99, TTFT, and TPOT in the expected direction;
- whether there is a remaining bottleneck that deserves a separate optimization issue.

## Suggested benchmark shape

Start from the Qwen3 mixed-load method:

- background: several long-lived decode streams, `ignore_eos`, timestamp every token;
- injector: one long-prompt request every fixed interval, short output;
- metric: inter-token gaps for background streams, tagged as steady or overlapping injected prefill;
- compare against decode-only baseline;
- record cold-prompt behavior first.

Suggested cells, adjusted for available VRAM:

- background concurrency: 4 or another stable value that fits Qwen3.5;
- injector prompt lengths: 4k / 8k / feasible upper bound;
- injector rates: low and medium QPS cells that do not immediately saturate;
- optionally compare different `max_prefill_tokens` / scheduler chunk budgets if the CLI exposes them.

Record:

- TTFT for injected requests;
- background ITL p50/p95/p99/max;
- stall vs steady ITL buckets;
- completed and failed requests;
- prefill saturation / OOM / admission rejections if any;
- GPU, commit, model revision, and exact flags.

## Acceptance criteria

- A Qwen3.5 mixed-load benchmark report is added to `docs/benchmarks/` or `docs/models/qwen35/`.
- The report compares mixed-load ITL against a decode-only baseline.
[...]

PR #693, opened 2026-07-16
[feat] qwen35: measure mixed-load ITL after chunked prefill + bench --max-batch (#470)

bbirdxr:
Answers #470: does scheduler-level chunked prefill bound the Qwen3.5 ITL tail when long cold prompts arrive into steady-state decode?

Tooling
- bench_serving: add `--max-batch` (default 4, backward-compatible). The qwen35 branch hardcoded scheduler capacity to 4, so `mixed --bg-concurrency 4` filled every slot and the injector never overlapped a full decode batch — the real cause of the old "p99-immune" reading. `mixed` now warns (into the report `warnings`) when `bg_concurrency >= max_batch`, kept runnable as the starvation negative control.
- Env-gated `ITL_STEP` scheduler log + `scripts/itl_step_agg.py` attribute the true per-step decode stall, independent of the coarse injection-window bucketing.

Report (docs/models/qwen35/mixed-load-itl-470.md)
- Full cold qps x prompt x chunk{on,off} matrix on RTX 4090, 24/24 valid, plus a max_batch=bg=4 starvation negative control.
- Qwen3.5 is not immune: a cold long prompt freezes active decode for ~the whole prefill wall (4k~0.29s ... 16k~1.18s).
- Chunked prefill bounds per-step/max stall but raises headline p99 in low-QPS cells (~14 -> ~80-92ms) and does not reduce total prefill work (+~15% TTFT).
- qps=1.0 12k/16k are a throughput wall (qps*prefill_s >= 1), kept qualitative only (single-run, small sample).
- Corrects the old "p99-immune" claim in benchmarks/mixed-load-itl.md to slot starvation. Raw per-cell data is kept locally outside the repo (untracked).

Scope: no prefix cache, no new scheduler policy, no production-readiness claim. The adaptive chunk-budget / SLA-aware policy is deferred to a follow-up issue after this lands.


CAICAIIs:
thanks, an the retained #470 report currently uses `--max-batch 5 --bg-concurrency 4` as the main valid matrix, but the Qwen3.5 loader only accepts decode bucket capacities `{1,2,4,8,16,32,64}` and `bench_serving` passes the CLI value directly into `openinfer_qwen35_4b::start_engine`. So the documented reproduction command with `--max-batch 5` cannot actually load Qwen3.5 on this PR, which invalidates the retained evidence as written.

Please either rerun/update the retained matrix with a legal capacity such as `--max-batch 8 --bg-concurrency 4`, or split scheduler admission capacity from graph/buffer bucket capacity so `5` maps to an allocated bucket safely.

Two smaller fixes should come with that:
- Record `max_batch` and `max_prefill_tokens` in the mixed JSON config, since both are core to reproducing this report.
- Gate the slot-starvation warning to Qwen3.5 or use the model’s real scheduler capacity; today Qwen3 can get a false warning because it ignores the new CLI `--max-batch`.

Also `cargo fmt --check` currently fails on `openinfer-qwen35-4b/src/scheduler.rs`.

bbirdxr:
Thanks for the detailed review! Addressed all points in the latest revision.

CAICAIIs (approved):
lgtm

PR #695, opened 2026-07-17
fix(qwen3): order cross-stream prefill temporaries in the decode-overlap path

FeathBow:
## Description

Closes a cross-stream buffer use-after-free in the SplitConcurrent (`--decode-overlap green-ctx`/`stream`) path, identified by data-flow analysis. The natural fault is allocator-timing-dependent.

Under decode-overlap, a step launches its prefill asynchronously on the prefill stream and decode on the decode stream (SM-partitioned in `green-ctx`, shared in `stream`). Those streams honor kernel launches (via the thread-local override) but not the `ctx.stream` memory ops — allocation, H2D copy, memset — that produce the kernels' inputs. So for every prefill temporary the stream-ordered allocator's two ordering requirements are both violated:

1. **alloc → use is unordered.** `clone_htod`, `HiddenStates::zeros`, and the paged-plan H2D all run on `ctx.stream`; the gather/GEMM/norm/attention kernels that read or write those buffers run on the override stream. An override-stream kernel can touch a buffer before its `ctx.stream` allocation or copy has completed. The pre-existing `ctx.sync()` before the override is activated covers only allocations made *before* the prefill — not the prefill's own.
2. **use → free was only partly ordered.** The old thread-local `DEFERRED_DROPS` queue deferred `token_ids` (conditionally), `hidden`, the plan, and the layer `bufs`, but never `normed`, the gather `indices`, the logits-head `gathered`, or the LoRA `rank_out`/`delta`/`compact_input` — those freed immediately on `ctx.stream` while override-stream kernels still read them. And the deferred set was cleared by an explicit drain that an error return skipped, so on the error path the missed temporaries free while the prefill kernels are in flight, and any already-deferred entries instead linger to the next drain.

Freeing (or reading before allocation completes) memory a live kernel touches can poison the CUDA context and corrupt re-admitted KV pages. It bit bring-up as an Xid 31/43; it is narrow and timing-dependent — it surfaces only when the allocator reuses the memory before the kernel reaches it.

## Fix

Enforce both orderings explicitly, and only under a stream override: the producer fence returns early when there is no override, and the bin's parking window is opened only on the split path — so with overlap off `park()` just drops in place. The one refactor that also touches the non-overlap path — hoisting the logits-head allocations so they precede a single fence — is behaviour-equivalent there.

## Verification — Single GPU (sm_89, x86_64), Qwen3-4B

The natural bug is timing-dependent — allocator reuse decides whether it corrupts — so a passing run under normal load proves nothing. The check is a static lifetime/ordering proof plus a deterministic A/B that forces the race. What is established:

- **alloc → use holds by construction.** Every `ctx.stream` producer in the prefill path is separated from its override-stream consumer by a `fence_producers_before_override`; the temporaries those override kernels touch are `park()`ed — including `normed`, `indices`, `gathered`, and the LoRA scratch the old queue missed.
- **use → free is ordered on the checked paths.** The parked temporaries free only after `synchronize()` has waited for the prefill stream — called explicitly on the success path, or via `Drop` on a `?`-return (including a GEMM launch failure, now a checked `Err` instead of a panic that skipped the park). A failed sync aborts (fail-closed) rather than free unsynchronized. `Drop` also runs on a panic, but the ordering is only guaranteed for `?`/`Result` errors, not an arbitrary panic that unwinds past an unparked buffer. The decode stream is drained symmetrically by `DecodeStreamGuard` (success/`?`/panic) before any page release.
- **Deterministic A/B on the alloc→use race.** Same binary under a temporary, uncommitted env-gated instrumentation overlay (not part of this PR): a producer-side delay on `ctx.stream` (a memset before the token_ids H2D) forces the override embedding to read before the copy completes; a second toggle disables the fence. Under `--decode-overlap green-ctx`, greedy, with a concurrent decode load forcing a split step — the control (no delay) and fence-on arms produce byte-identical output with zero CUDA errors, while the fence-off arm faults with `CUDA_ERROR_ILLEGAL_ADDRESS`, surfaced at the prefill-stream sync, which fail-closed aborts. That the fence-off arm faults confirms the override path fired in all three arms (the delay is gated on `has_stream_override()`). Forcing the race turns the fence's effect from luck into a clean signal: off it is an illegal access, on it is byte-identical correct output.

## Scope

Decode-overlap `green-ctx`/`stream` (the SplitConcurrent async-prefill path). With overlap off there is no added cross-stream synchronization and the success-path kernel math is unchanged. The producer fence creates/records/destroys one event per phase boundary — a device-side dependency with no host block, active only under an override. 

## Type of Change


FeathBow:
This PR has some performance issues. It is currently under review.

FeathBow:
## Performance / cost

- **No new per-step memory.** This PR adds no LoRA workspace; LoRA prefill scratch stays per-projection (freed per projection, no accumulation), and the base-model prefill buffers are unchanged.
- **Overlap off (default): zero added synchronization.** The producer fence returns early with no override and `park()` drops in place, so the non-overlap path is behaviour-equivalent to before.
- **Overlap on: device-side only.** The fences are `cuStreamWaitEvent` dependencies with no host block; the added waits are the existing per-step decode drain and the single prefill-stream drain at resolve — no per-layer or per-projection sync.
- **Checked GEMMs and the execution-boundary LoRA guard are negligible** — a launch status check and a one-time adapter scan per step.

PR #705, opened 2026-07-18
fix: eliminate the scheduler-thread stall on bulk KV restore (#704)

### LINKED ISSUE #704: Token-readback DtoH blocks the step loop behind bulk KV-restore H2D traffic (P/D co-resident decode hiccup)
## Symptom

When a bulk KV restore (pegaflow P2P fetch + H2D injection, ~2.5 GiB) lands on a decode node while other requests are decoding, every co-resident stream sees one synchronized ~110 ms ITL hiccup. Reproduced deterministically on Qwen3-14B, 1P+1D on 2x H200 (8 concurrent decode streams + one 16k-token cold restore).

At heavy multi-turn load this is the dominant residual tail after prefix-affinity routing (novitalabs/pegaflow#405): D-side post-eviction restores line up with decode and lift ITL p99.

## Mechanism (nsys-verified, not a guess)

- GPU never idles: kernel occupancy stays ~95% through the hiccup (launch-ahead keeps the queue fed). No kernel gap >10 ms. It is NOT HBM-bandwidth or SM contention.
- The stall is on the executor step thread: `select_batch`'s greedy token readback — `clone_dtoh(&scratch.argmax_out)` at `openinfer-sample/src/lib.rs:184` — goes from sub-ms to a flat **23.6 ms per call, 8 calls back-to-back** (one per step while the restore flood drains). 24 slow DtoH calls in a 60 s capture, all inside the 3 restore windows, p50=p90=23.6 ms.
- Why: `clone_dtoh` copies into pageable memory, so the async copy has synchronous semantics — the call blocks until the copy completes, and the copy waits behind the restore's H2D flood (pegaflow Direct mode issued **~14k coalesced ranges averaging 63 KiB** for 848 MiB — a fragmented queue the small DtoH must drain through).
- Net effect: the step loop slows ~8 consecutive steps x 23.6 ms while the GPU stays busy; token delivery for every active stream pauses ~110 ms, then bursts.

## Fix directions

1. **openinfer (primary)**: make the per-step token readback immune to copy-queue congestion — persistent pinned readback buffer + a dedicated (high-priority) stream, or event-polled readback instead of a blocking pageable DtoH. Touches the sampling hot path, so it needs the accuracy gates re-run.
2. **pegaflow (secondary)**: restore injection issues thousands of tiny coalesced ranges; batching them into larger copies (or the Kernel backend for fragmented batches, or a rate-limited/low-priority injection stream) would shrink the congestion window for any host-side reader.
3. M3 layer-wise push (#528) amortizes the same traffic and shrinks the exposure window structurally.

## Evidence

- Client-side: 8/8 decode streams spike 106–112 ms at the same millisecond, exactly at restore completion; reproduced across runs; restores that land with no co-resident decode produce no spike.
- nsys: capture with kernel busy-per-20ms-bucket ~19 ms throughout the window; `cuMemcpyDtoHAsync_v2` p50 23.632 ms on the kernel-launching thread only inside restore windows.
- Full campaign context: `docs/benchmarks/qwen3-14b-pd-vs-mix-h200.md` (PR #701).

xiaguan:
## Problem (#704)

Under P/D disaggregation with heavy cross-node KV traffic, every active decode stream froze for ~70ms whenever a bulk restore landed. nsys showed the GPU 95% busy with no kernel gaps — the stall was CPU-side, on the scheduler thread:

1. `commit_loaded_blocks` registers restored blocks inline at ~70µs each (registry radix insert + event hook + frequency touch + store lock). A 1000+ block restore = ~70ms with token delivery frozen for all streams.
2. The greedy-token readback used a pageable D2H copy, which has synchronous-copy semantics: under concurrent bulk copy traffic the sub-ms readback flattened to 23.6ms per step.

## Fix

- **Chunked commit**: restore registration becomes a `Committing` prefetch phase drained at 64 blocks/tick (~4.5ms), paced only while live decode rows exist. Pure-prefill ticks (P/D prefill node, idle decode node) commit inline — pacing there multiplies restore latency by the prefill tick interval for no benefit.
- **Pinned readback**: greedy tokens land in a pinned host buffer; the reader blocks only on the copy's own event.
- Also fixes a load-failure stranding bug: `settle_prefetch`'s `Err` path now reports the request immediately prefill-eligible.

Alternatives measured and rejected (details in commit message): off-thread commit worker (store-lock contention, RR heavy 437→258 tok/s), Kernel transfer backend (copy kernels steal SMs from decode — two ~110ms stall waves vs Direct's one).

## Validation (Qwen3-14B 2P+2D on 8×H200, multi-turn vllm-bench, cold start)

| config | output tok/s | TTFT p50/p99 (ms) | ITL p99 (ms) |
|---|---|---|---|
| main, round-robin router | 249 | 1563 / 12810 | 86.5 |
| this PR, round-robin router | 259 | 1267 / 11717 | 86.0 |
| this PR, prefix-affinity router | 268 | 1911 / 18095 | **22.8** |
| main, warm rerun | 244 | 952 / 23805 | 83.2 |
| this PR, warm rerun | **298** | 779 / 14702 | 85.9 |

- Micro-probe (8 decode streams + 4×16k cold restores): ITL spikes >40ms went from every-restore to **zero**.
- Warm rerun (restore-heavy): +22% throughput over main.
- Greedy byte-identical P/D vs direct-D baseline across page-boundary prompt sizes (GATE1-PASS, RDMA fetch confirmed).
- `hf_golden_gate` (Qwen3-4B) passes; unit tests + clippy clean on touched crates.

Note: an earlier report of a RR-heavy regression on this branch traced to a contaminated baseline — the 437 tok/s reference was measured on a progressively warm stack (fixed bench seed → identical prompts → full prefix hits). Cold A/B above shows no regression.

Closes #704.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2455,6 +2455,7 @@ dependencies = [
  "bincode",
  "bytes",
  "criterion 0.5.1",
+ "dashmap",
  "derive_builder",
  "dynamo-tokens",
  "futures",
--- a/kvbm/kvbm-logical/Cargo.toml
+++ b/kvbm/kvbm-logical/Cargo.toml
@@ -10,6 +10,7 @@ license.workspace = true
 repository.workspace = true
 
 [dependencies]
+dashmap = { workspace = true }
 dynamo-tokens = { workspace = true }
 
 anyhow = { workspace = true }
--- a/kvbm/kvbm-logical/src/registry/handle.rs
+++ b/kvbm/kvbm-logical/src/registry/handle.rs
@@ -87,7 +87,7 @@ impl Drop for BlockRegistrationHandleInner {
         // during `drop_in_place` the inner allocation is still live (the
         // implicit weak from the strong refcount is released after `Drop`
         // returns). Only remove if the entry still points to us.
-        let map = registry.prefix(&self.seq_hash);
+        let mut map = registry.prefix(&self.seq_hash);
         let should_remove = match map.get(&self.seq_hash) {
             Some(weak_ref) => std::ptr::eq(weak_ref.as_ptr(), self as *const Self),
             None => {
--- a/kvbm/kvbm-logical/src/registry/mod.rs
+++ b/kvbm/kvbm-logical/src/registry/mod.rs
@@ -5,7 +5,8 @@
 //!
 //! The [`BlockRegistry`] is the central coordination point for block deduplication in the
 //! KVBM system. It maps sequence hashes to registration handles using a
-//! [`dynamo_tokens::PositionalRadixTree`], enabling efficient prefix-based lookups.
+//! position-keyed two-level map ([`prt::PositionalRadixTree`]), enabling
+//! efficient prefix-based lookups.
 //!
 //! # Architecture
 //!
@@ -30,6 +31,7 @@
 
 mod attachments;
 mod handle;
+mod prt;
 mod registration;
 
 #[cfg(test)]
@@ -47,7 +49,7 @@ use std::sync::{Arc, Weak};
 
 use handle::BlockRegistrationHandleInner;
 
-pub(crate) type PositionalRadixTree<V> = dynamo_tokens::PositionalRadixTree<V, SequenceHash>;
+pub(crate) type PositionalRadixTree<V> = prt::PositionalRadixTree<V, SequenceHash>;
 
 /// Builder for [`BlockRegistry`].
 ///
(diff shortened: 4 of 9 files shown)


xiaguan:
Full-stack validation of the final branch (`a3b3ecac`), Qwen3-14B 2P+2D on 8×H200, cold restart per config:

- **Micro-probe** (8 decode streams direct to D + 4×16k cold restores through the router): ITL spikes **0** (was: every restore froze all streams).
- **RR heavy** (60 sessions × 5 turns, 8k+2k/turn, conc 12): 248 tok/s, TTFT p50/p99 1277/14701ms, TPOT p99 33.5, ITL p99 80.6 — at parity with main (249) on throughput, tail smoothness improved.
- **Prefix-affinity heavy**: 267/278 tok/s across two runs, TPOT p99 22.5, **ITL p99 22.7/22.8** — decode isolation matches a dedicated 1P+1D.
- Greedy byte-identical P/D acceptance and Qwen3-4B `hf_golden_gate` pass (numeric paths untouched by this PR: the registry change is CPU-structural, the pinned readback waits on the same stream point as the old sync).

Note: throughput/TTFT in this workload have ±5-10% run-to-run variance; TPOT/ITL p99 are stable across runs and are the metrics this fix targets.

PR #706, opened 2026-07-18
fix(qwen3): wait for async overlap prefill while idle

### LINKED ISSUE #699: qwen3: scheduler can park on blocking_recv while an async decode-overlap prefill is still in flight
### Description

In decode-overlap mode, `scheduler_loop`'s idle gate checks only `active`/`deferred`/`prefilling` before parking on `submit_rx.blocking_recv()`. It does not check `inflight_prefill_pending`. The in-flight async prefill's result is only consumed by the top-of-loop poll, which never runs while the loop is parked — so the prefill's requests hang until the next unrelated submission wakes the scheduler.

### Reachable path

A Unified step launches an async prefill and saves its requests in `inflight_prefill_pending`; the same step's decode effects can retire the last active request (EOS / `max_tokens`). On the next iteration `poll_async_prefill` returns `None` (prefill still running on the overlap stream), the queues are empty — the launched requests' chunks were already taken off `prefilling` — and the idle gate parks the loop. The stall is observable: `publish_load` counts `inflight_prefill_pending` into `num_running_reqs`, so metrics freeze at a nonzero running count while nothing progresses. 

### Acceptance criteria

- [ ] A scheduler test with a fake decode-overlap executor: the last active request retires in the same step that launches an async prefill; without the fix the prefill's requests see no progress until a later submission, with the fix they complete promptly.
- [ ] The idle path still blocks (on the channel or on the prefill event) — no busy spin.
- [ ] Dropping the final engine handle while an async prefill is in flight neither abandons the prefill result nor tears down worker/stream state prematurely.

Nyvo-io:
## Summary

- keep the scheduler from parking on new submissions while decode-overlap prefill work is still in flight
- wait on a blocking, timing-disabled CUDA event only when no CPU or decode work can make progress
- guard the full launched-but-not-transferred interval on both worker and executor handoffs
- defer GPU temporary destruction until prefill-stream quiescence is proven
- preserve fail-stop ownership when CUDA synchronization cannot prove stream quiescence

## Why

A unified decode-overlap step can retire the last active request while its prefill continues on the overlap stream. With no active, deferred, or prefilling work left, the scheduler previously blocked only on `submit_rx.blocking_recv()`, so the prefill result was not consumed until an unrelated request arrived.

The first fix still left a wider ownership gap after prefill launch: decode launch/synchronization, result construction, worker aggregation, and executor-side decode application could fail before durable async-prefill state owned the GPU work. Those returns could let scheduler cleanup recycle KV pages or drop temporary allocations while the prefill stream still referenced them.

## Implementation

The scheduler centralizes async-prefill result application and, when the overlap prefill is the only remaining work, calls `wait_async_prefill` before considering the engine idle.

`SplitConcurrent` carries the Green Context owner retained by `OverlapStreams`. GreenCtx mode records completion with `cuGreenCtxRecordEvent(gctx_prefill, event)`; shared-SM mode uses `cuEventRecord(event, prefill_stream)`. Events use `CU_EVENT_BLOCKING_SYNC | CU_EVENT_DISABLE_TIMING`.

Stream guards are armed before fallible prefill and decode forwards. The prefill guard stays armed until `LocalQwen3Lane::inflight_prefill` owns the launched work; an owned completion event then protects worker aggregation and executor-side decode application until `Qwen3Executor::async_prefill` takes ownership. Every pre-handoff error synchronizes before scheduler cleanup can recycle KV, and an unprovable synchronization failure aborts.

GPU-backed prefill temporaries, including errors raised inside the forward, move to a deferred-drop queue while a stream override is active. They are destroyed only after successful prefill-stream synchronization, preventing `cuMemFreeAsync` on the primary stream from racing kernels on the overlap stream.

Event create/record/cleanup errors are checked. A record failure preserves a simultaneous `cuEventDestroy_v2` cleanup failure in the returned context. After query or synchronize has already proven completion, a destroy failure is logged and treated as non-fatal: it may leak a driver resource, but cannot race KV or temporary-buffer reuse. Post-wait state/result mismatches retain the complete `anyhow` cause chain for the client.

## Validation

Final follow-up head `7698fdd` was validated on RTX 5090 / CUDA 12.8:

- Qwen3 Clippy with `-D warnings`
- complete Qwen3 library suite: 79 passed, 0 failed
- release `openinfer-server` build
- real `green-ctx` concurrent smoke: 7/7 requests completed, with 13 async-prefill unified steps and 13 completions
- real shared-SM `stream` smoke: 4/4 requests completed, with 6 async-prefill unified steps and 6 completions
- no invalid-handle, event record/query/synchronize, FATAL, worker-response, or Xid errors
- full Codex review of the complete PR plus follow-up changes: no actionable findings

The branch remains based on `origin/main` `afd7a64`. Final GitHub CI run `29889818043` passes all 15 checks on `7698fdd`, including formatting, DCO, attribution, locked Cargo metadata, CPU Clippy/tests, simulated frontend E2E, Qwen3 CUDA compile, and Qwen3 CUDA Clippy.

Closes #699


openinfer-qwen3/src/executor.rs
```diff
@@ -2971,6 +2978,22 @@ impl ModelExecutor for Qwen3Executor {
         }
         Some(result)
     }
+
+    fn wait_async_prefill(&mut self) -> Result<PrefillResult> {
+        let state = self
+            .async_prefill
+            .as_ref()
+            .ok_or_else(|| anyhow::anyhow!("no async prefill is in flight"))?;
+        let status = unsafe { cudarc::driver::sys::cuEventSynchronize(state.event) };
```
FeathBow: This still does not satisfy #699's no-busy-spin acceptance criterion. The event being synchronized here is created at L494-498 with only CU_EVENT_DISABLE_TIMING; CUDA documents that cuEventSynchronize busy-waits unless the event has CU_EVENT_BLOCKING_SYNC. The fake executor's Condvar wait masks that production behavior.

FeathBow (changes requested):

openinfer-qwen3/src/scheduler.rs
```diff
@@ -577,8 +598,33 @@ fn scheduler_loop<E>(
...
+                        &mut active,
+                        &mut prefilling,
+                        &mut inflight_prefill_pending,
+                        prefill_result,
+                    ),
+                    Err(error) => {
+                        warn!("decode-overlap: async prefill wait failed: {error:#}");
+                        let pending = inflight_prefill_pending.take().unwrap();
+                        let targets = pending.iter().map(pending_failure_target).collect();
+                        fail_touched_requests(
```
FeathBow: wait_async_prefill can reach this branch when cuEventSynchronize fails. fail_touched_requests then calls drop_request, returning the pending requests' KV pages even though the failed synchronization did not establish that the prefill stream has stopped using them.

FeathBow (changes requested):

openinfer-qwen3/src/executor.rs
Nyvo-io: Addressed in 655f373: the event is now created with CU_EVENT_BLOCKING_SYNC | CU_EVENT_DISABLE_TIMING, so the idle-only cuEventSynchronize blocks the host thread instead of busy-waiting. The normal top-of-loop cuEventQuery path remains non-blocking.

openinfer-qwen3/src/scheduler.rs
Nyvo-io: Addressed in 655f373: a wait error now sends request errors and fail-stops the executor without calling drop_request, so unquiesced prefill KV pages cannot re-enter the live allocator. Added decode_overlap_wait_failure_does_not_release_inflight_kv. Validation: all 12 GitHub checks pass; on RTX 5090 / CUDA 12.8 the two decode-overlap tests pass and the full Qwen3 lib suite passes 75/75.

openinfer-qwen3/src/executor.rs
```diff
@@ -493,7 +493,9 @@ fn execute_step_on_lane(
                 unsafe {
                     cudarc::driver::sys::cuEventCreate(
                         &raw mut event,
-                        cudarc::driver::sys::CUevent_flags_enum::CU_EVENT_DISABLE_TIMING as u32,
+                        cudarc::driver::sys::CUevent_flags_enum::CU_EVENT_BLOCKING_SYNC as u32
+                            | cudarc::driver::sys::CUevent_flags_enum::CU_EVENT_DISABLE_TIMING
+                                as u32,
                     );
                     cudarc::driver::sys::cuEventRecord(event, prefill_stream.0);
```
FeathBow: The blocking flag is fixed, but this event still cannot reliably represent GreenCtx prefill work. The worker is bound to the primary context, so cuEventCreate creates a primary-context event, while prefill_stream comes from cuGreenCtxStreamCreate. CUDA requires cuEventRecord’s event and stream to share a context; the call therefore fails in GreenCtx mode, but both create/record results are ignored. An unrecorded event represents empty work, so poll_async_prefill can treat it as immediately complete and fall through to the worker’s cuStreamSynchronize, bypassing the new blocking-event path. Could this use cuStreamGetGreenCtx + cuGreenCtxRecordEvent for GreenCtx, plain cuEventRecord for SharedSm, and check/clean up both failure paths? #695 already contains most of this dispatch; when combining them, the event also needs to retain CU_EVENT_BLOCKING_SYNC.

openinfer-qwen3/src/scheduler.rs
```diff
@@ -577,8 +598,38 @@ fn scheduler_loop<E>(
...
+                        // here could race the in-flight kernels.
+                        let message = error.to_string();
+                        for req in inflight_prefill_pending.as_ref().unwrap() {
+                            let _ = req.token_tx.send(TokenEvent::Error {
+                                message: message.clone(),
+                                prompt_tokens: req.prompt_tokens.len(),
+                                completion_tokens: 0,
+                            });
+                        }
+                        return;
```
FeathBow, line 666: This avoids returning the KV pages to the live pool, but return still drops Qwen3Executor. Its shutdown makes the worker exit without resolve_inflight_prefill, which drops InflightPrefillState and its GPU buffers even though the failed wait explicitly did not establish prefill-stream quiescence. Those CudaSlice frees are enqueued on their owning primary stream, with no ordering behind the overlap stream. Could this abort on an untrustworthy synchronization failure, or otherwise guarantee that the in-flight allocations are never freed? The fake test only proves that drop_request was not called; it cannot cover the CUDA teardown race.

FeathBow (changes requested):

Nyvo-io:
Addressed the latest review feedback in commit `8aba65d` and pushed it to the PR branch.

- Async prefill events are now created with `CU_EVENT_BLOCKING_SYNC | CU_EVENT_DISABLE_TIMING`; creation, Green Context lookup, recording, and cleanup errors are checked.
- The event helper calls `cuGreenCtxRecordEvent` for Green Context streams and `cuEventRecord` for ordinary shared-primary-context streams, so the event and stream use the same CUDA context.
- `cuEventQuery` treats only `CUDA_ERROR_NOT_READY` as pending. Any other query/synchronize failure fail-stops the real Qwen3 executor instead of allowing `Drop` to free buffers whose prefill stream may still be using them.
- If event setup fails after prefill launch, the prefill stream is synchronized before returning the error; a failed fallback synchronization aborts.

Validation on the RTX 5090 / SM120 AutoDL host:
- `cargo test --release -p openinfer-qwen3 --lib decode_overlap_` (2 passed, 0 failed; the test binary compiled the new Green Context APIs)

Please re-review the latest commit.

Nyvo-io:
Follow-up: the implementation commit was amended with the required DCO sign-off and force-with-lease pushed as `9ad98b7` (superseding `8aba65d`). The code and AutoDL validation are unchanged; the new GitHub CI run is in progress.

Nyvo-io:
DCO follow-up: the PR history has been rewritten so all four commits now carry valid `Signed-off-by` trailers matching their authors where applicable. The new head is `43fb4e5` (superseding `9ad98b7`); code content is unchanged. GitHub CI will rerun from this head.

openinfer-qwen3/src/green_ctx.rs
```diff
@@ -102,6 +102,43 @@ fn create_primary_stream() -> Result<CUstream> {
...
+                &raw mut event,
+                sys::CUevent_flags_enum::CU_EVENT_BLOCKING_SYNC as u32
+                    | sys::CUevent_flags_enum::CU_EVENT_DISABLE_TIMING as u32,
+            )
+        },
+        "cuEventCreate (async prefill)",
+    )?;
+
+    let mut green_ctx: sys::CUgreenCtx = ptr::null_mut();
+    let get_ctx = unsafe { sys::cuStreamGetGreenCtx(stream, &raw mut green_ctx) };
```
FeathBow: cuStreamGetGreenCtx still cannot reliably identify the GreenCtx streams created here. create_green obtains ctx_* with cuCtxFromGreenCtx but never makes either context current, while CUDA documents that cuGreenCtxStreamCreate ignores the caller's current context. In this sequence cuStreamGetGreenCtx therefore returns NULL, so this helper falls through to cuEventRecord for a GreenCtx stream. CUDA requires the event and stream to share a context, so the original CUDA_ERROR_INVALID_HANDLE can still occur and the cuGreenCtxRecordEvent branch is never reached. Could this use cuStreamGetCtx_v2(..., pGreenCtx) or retain/pass gctx_prefill explicitly? Please also run a real --decode-overlap green-ctx smoke test; the current fake scheduler tests do not exercise this driver path.

openinfer-qwen3/src/scheduler.rs
```diff
@@ -577,8 +598,38 @@ fn scheduler_loop<E>(
...
+                match executor.wait_async_prefill() {
+                    Ok(prefill_result) => apply_async_prefill_result(
+                        &mut executor,
+                        &mut active,
+                        &mut prefilling,
+                        &mut inflight_prefill_pending,
+                        prefill_result,
+                    ),
+                    Err(error) => {
+                        warn!("decode-overlap: async prefill wait failed: {error:#}");
```
FeathBow, line 652: This comment no longer matches HEAD: a failed cuEventSynchronize now aborts inside wait_async_prefill, so the only Errs reachable here are the two state-mismatch cases ("no async prefill is in flight" / "completed without a result"), and the latter has already established quiescence. Worth rewording so the next reader doesn't reason about safety from an unreachable scenario — same for the regression test's "wait failure" framing, which the real executor can no longer produce.

openinfer-qwen3/src/scheduler.rs
```diff
@@ -577,8 +598,38 @@ fn scheduler_loop<E>(
...
+                        // Event synchronization failed, so the prefill stream
+                        // may still be using these requests' KV pages. Fail the
+                        // requests but retain executor ownership until the
+                        // fail-stop teardown; returning them to the live pool
+                        // here could race the in-flight kernels.
```
FeathBow: error.to_string() drops the anyhow chain, the warn above already logs {:#}, so consider format!("{error:#}") here too if the client message should carry the cause.

FeathBow (changes requested):

openinfer-qwen3/src/green_ctx.rs
Nyvo-io: Addressed in 62e6ba8. SplitConcurrent now carries the retained gctx_prefill handle from OverlapStreams into the worker. GreenCtx records completion with cuGreenCtxRecordEvent(gctx_prefill, event), while SharedSm continues to use cuEventRecord(event, prefill_stream), so no runtime ownership inference is required. Validation: release build, Qwen3 Clippy with -D warnings, and both decode-overlap regression tests pass. On an RTX 5090, a real green-ctx concurrent smoke completed 7/7 requests with repeated async-prefill unified steps and no invalid-handle, event record/query/synchronize, FATAL, or Xid errors. A separate stream-mode smoke completed 4/4 requests and covered the plain cuEventRecord branch.

openinfer-qwen3/src/scheduler.rs
Nyvo-io: Addressed in 62e6ba8. The trait and scheduler comments now reflect the reachable semantics: CUDA query/synchronize failures fail-stop inside the real executor, and a returned Err is limited to post-wait state/result resolution after event completion established quiescence. The old fake wait-failure test was replaced with decode_overlap_missing_result_reports_chain_and_stops_scheduler, which simulates a completed event with a lost worker result.

openinfer-qwen3/src/scheduler.rs
Nyvo-io: Addressed in 62e6ba8. The client-facing error now uses format!("{error:#}"), and the regression test asserts that both the outer async-prefill context and the underlying fake worker-response cause are preserved. All 15 checks in CI run 29799900510 pass.

Nyvo-io:
All maintainer feedback is addressed on final head `6baaf3b`.

- The idle-only wait uses a blocking, timing-disabled CUDA event. GreenCtx ownership is retained and passed explicitly to `cuGreenCtxRecordEvent`; SharedSm uses `cuEventRecord`. Event setup/cleanup is checked, and untrustworthy CUDA query/synchronize failures remain fail-stop so in-flight KV and temporary buffers cannot be released unsafely.
- Scheduler comments and the regression test now match the reachable post-wait state/result-mismatch semantics, and the client error preserves the complete `anyhow` cause chain.
- The branch is rebased onto `origin/main@afd7a64` and condensed to one Nyvo-authored, Nyvo-committed, DCO-signed commit while retaining the upstream device-ordinal, batched-logprob, TP checkpoint-prefetch, and rustfmt changes.

Validation is complete:

- final GitHub CI run `29855508474`: 15/15 checks passed on `6baaf3b`
- RTX 5090 / CUDA 12.8: release build, Qwen3 Clippy with `-D warnings`, both deterministic scheduler regressions, and the full Qwen3 library suite passed
- real decode-overlap smoke: `green-ctx` 7/7 requests and `stream` 4/4 requests completed with repeated async-prefill unified steps and no invalid-handle, CUDA event, FATAL, Xid, or worker-response errors

Please re-review when convenient.

openinfer-qwen3/src/scheduler.rs
FeathBow, line 597: Thanks for making event-setup failure quiesce the prefill stream. There is still a wider ownership gap after the prefill launch, though: `execute_decode`, decode-stream synchronization, and `build_batch_decode_request_results` can return before the event and `lane.inflight_prefill` are installed. The scheduler then runs `fail_touched_requests -> drop_request`, which can return this Unified step's KV pages while the prefill stream still uses them. Could we guard the whole launched-but-not-transferred interval, synchronizing prefill on every early return and aborting if that synchronization fails? The later handoff in `Qwen3Executor::execute_unified` needs the same protection because `rkv.apply_decode(...)?` currently runs before `self.async_prefill` owns the event.

openinfer-qwen3/src/green_ctx.rs
```diff
@@ -105,7 +114,48 @@ fn create_primary_stream() -> Result<CUstream> {
...
+        "cuEventCreate (async prefill)",
+    )?;
+
+    let record = match green_ctx {
+        Some(green_ctx) => unsafe { sys::cuGreenCtxRecordEvent(green_ctx.0, event) },
+        None => unsafe { sys::cuEventRecord(event, stream) },
+    };
+    if record != sys::CUresult::CUDA_SUCCESS {
+        unsafe {
+            sys::cuEventDestroy_v2(event);
```
FeathBow: Small mismatch with the PR contract: both this destroy and the normal completion destroy in `poll_async_prefill` discard `cuEventDestroy_v2`'s result, while the PR says event cleanup failures are checked. Since this path already has the recording error, could we preserve the destroy failure as context as well, and handle/log the completed-event cleanup result in `poll_async_prefill`? If cleanup errors are intentionally non-fatal after completion is proven, the PR text should say that explicitly.

FeathBow (changes requested):

openinfer-qwen3/src/scheduler.rs
Nyvo-io: Addressed in `7698fdd`. The worker now arms a prefill-stream guard before entering the fallible forward and keeps it armed through decode launch/synchronization, decode-result construction, and completion-event creation. It disarms only after `lane.inflight_prefill` owns the launched work. The completion event is then an RAII owner across worker aggregation and `rkv.apply_decode`, and transfers only when `self.async_prefill` is installed. Any earlier return synchronizes before KV cleanup; an unprovable synchronization failure aborts. GPU temporaries created inside the forward are deferred until that synchronization, which also fixes the Rust unwind/drop-order gap found during the full PR review.

openinfer-qwen3/src/green_ctx.rs
Nyvo-io: Addressed in `7698fdd`. A record failure now returns both the record status and any simultaneous `cuEventDestroy_v2` cleanup failure. Normal query/synchronize completion routes through checked destruction as well. Once completion is already proven, destroy failure is logged and intentionally non-fatal because it can leak an event resource but cannot race KV or temporary-buffer reuse; the PR description now states that policy explicitly. The new CPU regression covers the combined record/destroy error and event ownership cleanup.

Nyvo-io:
All latest maintainer feedback is addressed on head `7698fdd` after a full PR review.

- The entire launched-but-not-transferred interval is guarded on the worker and executor handoffs, including fallible forward internals and Rust drop ordering.
- Event record/destroy double failures preserve both statuses; completed-event destroy failures are checked, logged, and explicitly documented as non-fatal only after quiescence is proven.
- Final RTX 5090 validation passed: format, Qwen3 Clippy `-D warnings`, 79/79 Qwen3 lib tests, release server build, GreenCtx 7/7 with 13 paired async-prefill steps, and shared-SM stream 4/4 with 6 paired steps; no event/FATAL/worker/Xid errors.
- Full Codex review of the complete PR plus follow-up changes reported no actionable findings.
- GitHub CI run `29889818043` passes all 15 checks.

Please re-review when convenient.

FeathBow:
Please rebase onto main before review. If you encounter conflicts, ensure tests are re-run as necessary.

Nyvo-io:
Rebased onto current `main` at `475a299` and pushed head `48e8f43` with `--force-with-lease`; GitHub now reports the PR as mergeable.

Addressed the remaining async-prefill admission gap:
- in-flight prefill requests now reserve their future KV footprint and decode slots until the async result is applied;
- added a deterministic regression covering admission during an in-flight prefill;
- retained the idle-wait and missing-result regressions from the prior review round.

Validation:
- `git diff --check`: passed
- E10 RTX 5090 D: `cargo fmt --all -- --check` passed
- targeted E10 test compilation was blocked by the machine's stale `vllm-text`/`vllm-tokenizer` stubs in `openinfer-vllm-support` (API symbols missing before the Qwen3 test binary linked); GitHub CI is running against the rebased head.
- all four PR commits carry valid DCO sign-offs.

Please re-review the rebased head.

Nyvo-io:
Follow-up on the rebased CI run: the two Qwen3 gates exposed a latest-main compatibility issue unrelated to the admission logic. `resolve_async_prefill_after_error` was still positioned inside the `ModelExecutor` trait impl after the trait method was removed upstream, and strict Clippy also caught two stale `unused_mut`s.

Fixed those minimally in `21890f9`, preserving the cleanup behavior, and pushed the DCO-signed commit. The new CI run is queued; the PR remains mergeable.

Nyvo-io:
Completed the requested follow-up on head `03ff3e0`.

- Rebased onto `origin/main@475a299`; GitHub reports the PR as `CLEAN`.
- Fixed the remaining Qwen3 Clippy `unused_mut` failure in the regression test.
- All 15 GitHub checks pass, including Qwen3 CUDA compile and Clippy.
- Final local review: `git diff --check` passed, all commits have DCO sign-offs, and the worktree is clean.

Please re-review the latest head.

openinfer-qwen3/src/scheduler/tests.rs
```diff
@@ -839,6 +959,213 @@ fn wait_until(timeout: Duration, mut predicate: impl FnMut() -> bool) -> bool {
...
+    handle
+        .submit(overlap_request)
+        .expect("submit overlap request");
+    decode_pause.release.wait();
+
+    assert!(wait_until(Duration::from_secs(1), || {
+        async_prefill_gate.wait_calls.load(Ordering::SeqCst) > 0
+    }));
+
+    let (deferred_request, mut deferred_rx) = request(16, 1);
```
FeathBow, line 1170: By the time this request is submitted, `wait_calls > 0` means the scheduler thread is already blocked inside `FakeExecutor::wait_async_prefill()` → `AsyncPrefillGate::wait()`. It therefore cannot drain `submit_rx` or call `admit_deferred_requests_with_inflight()` while the prefill remains in flight.

After `release()`, the scheduler first completes and applies that prefill, clearing the in-flight state, and only then processes this request. As a result, this test still passes if the in-flight admission-accounting change is reverted.

The stated slot boundary is not constructed either: `FakeExecutor::max_decode_batch_size()` returns 64, so three requests do not exhaust the decode slots.

Please keep an active decode runnable while the prefill gate is closed, submit the third request during those scheduler iterations, and constrain the decode-slot or KV budget so that the pre-fix implementation demonstrably over-admits while this implementation defers it. The regression should fail when the accounting patch is removed.

openinfer-qwen3/src/scheduler.rs
```diff
@@ -1128,10 +1183,22 @@ fn admitted_future_blocks<E: ModelExecutor>(
...
+fn admitted_future_blocks_with_inflight<E: ModelExecutor>(
+    executor: &E,
+    active: &[ActiveRequestState],
+    prefilling: &[PendingRequest],
+    inflight_prefilling: &[PendingRequest],
 ) -> usize {
     let block_size = executor.block_size();
     active_future_blocks(active, block_size)
         + prefilling_future_blocks(prefilling, block_size, |id| executor.prefetched_blocks(id))
+        + prefilling_future_blocks(inflight_prefilling, block_size, |id| {
```
FeathBow: This applies the queued-prefill accounting formula to requests whose current prefill chunk has already been scheduled. `Qwen3Executor::execute_unified()` calls `schedule_prefill_chunk()` before launching the worker, and `RequestKv::schedule_prefill()` allocates that chunk's KV blocks from the pool. Consequently, `executor.available_blocks()` already excludes those blocks when admission reaches this calculation.

`prefilling_future_blocks()` subtracts only `blocks_needed(req.prefill_pos)`, so for an in-flight request it treats the already-allocated current chunk as future demand and reserves it a second time. This makes admission unnecessarily conservative.

The effect can extend beyond delayed admission when KV offload is enabled: the inflated `reserve_floor` can cause `begin_kv_prefetch()` to decline the request, while `prefetch_offered` has already been set. That request is then not retried after the in-flight prefill completes, converting a possible KV restore into recomputation.

Please account in-flight requests from their actual post-schedule KV ownership/current scheduled state and reserve only the still-unallocated lifetime tail, rather than passing them through the queued-prefill formula unchanged.

FeathBow (changes requested):

Nyvo-io:
Addressed the latest review feedback in `9f80ebf` and completed the final review.

- In-flight async-prefill admission now reserves only the unallocated KV tail after `prefill_pos + step_chunk`; the already scheduled chunk is not charged twice.
- In-flight prefills continue to reserve their eventual decode slots.
- The overlap regression now submits the third request while an active decode remains runnable and the async prefill is still in flight, with a decode capacity of 2.
- Verified that the new regressions pass, and that reverting the accounting changes makes the slot, KV-tail, and overlap regressions fail.

Validation:
- E10 RTX 5090 D: targeted scheduler tests passed.
- E10: `cargo fmt --all -- --check` and strict Qwen3 Clippy passed.
- GitHub CI: all 15 checks passed.
- PR head is mergeable and clean.

openinfer-qwen3/src/scheduler.rs
```diff
@@ -1152,6 +1219,25 @@ fn prefilling_future_blocks(
...
+    prefilling: &[PendingRequest],
+    block_size: usize,
+    prefetch_credit: impl Fn(RequestId) -> usize,
+) -> usize {
+    prefilling
+        .iter()
+        .map(|req| {
+            let scheduled_prompt_tokens = req
+                .prefill_pos
+                .saturating_add(req.step_chunk)
```
FeathBow, line 1232: prefill_pos + step_chunk is exact only after the first chunk. For a new request, the scheduler’s saved clone still has prefill_pos == 0, while schedule_prefill_chunk() can first advance RequestKv::kv_position() through match_and_add_prefix() and then schedule up to step_chunk additional tokens.

For example, with a 2048-token prompt, a 512-token prefix hit, and a 1024-token chunk, the executor holds or schedules KV through token 1536, while this formula credits only 1024. The remaining 512 tokens are therefore still reserved twice.

Both new accounting tests use a no-prefix fake, so they remain green in this case. Please derive the reservation from the executor- authoritative post-schedule position and add a prefix-hit regression.

FeathBow (changes requested):

PR #708, opened 2026-07-18
perf(kvbm): plain HashMap inner level for the registry position map

xiaguan:
## What

Follow-up root-cause fix under #704: block registration cost was dominated not by the registry work itself but by a hidden allocation in `dynamo_tokens::PositionalRadixTree` — every first registration at a new block position allocates a full sharded `DashMap` as the inner level. Shard count scales with core count (4×cores rounded to a power of two): 256 shards on a 48-core box, **1024 on a 192-core H200 node**.

Instrumented breakdown of a 25µs/block registration (48-core box, per-stage timers inside `register_block`):

| stage | avg |
|---|---|
| `prt.prefix()` + inner `entry().or_default()` | **20.9µs** |
| handle `Arc` creation | 0.23µs |
| store slot transition (mutex + state) | 0.51µs |
| TinyLFU touch | 0.03µs |

The inner sharding buys nothing: every access goes through the outer entry's `RefMut`, which already holds the outer shard's write lock exclusively — and that guard doubles as the per-position critical section the handle-drop race protection relies on (locking semantics are bit-identical to the old version; the old `position()` bypass path had zero callers in this crate). Replaced the alias with a local two-level map whose inner level is a plain `HashMap` (allocation-free `Default`).

## Measured

1024-block restore commit (`commit_loaded_blocks`), 48-core box: **36.4ms → 1.67ms (22×)**, 1.63µs/block. On 192-core nodes the before-side is ~3× worse (1024-shard allocs measured ~70µs/block in situ during the #704 investigation), so the win is larger there.

This shrinks the constant that made #704 catastrophic; the chunked `Committing` pacing from #705 stays as the bounded-work-per-tick invariant for arbitrarily large restores.

- `kvbm-logical` full suite: 493 passed; clippy clean.
- Toxic-review pass: locking-semantics equivalence, handle-drop race, re-entrancy under the guard, and `len()` self-deadlock surface all verified; one dead trait bound removed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2455,6 +2455,7 @@ dependencies = [
  "bincode",
  "bytes",
  "criterion 0.5.1",
+ "dashmap",
  "derive_builder",
  "dynamo-tokens",
  "futures",
--- a/kvbm/kvbm-logical/Cargo.toml
+++ b/kvbm/kvbm-logical/Cargo.toml
@@ -10,6 +10,7 @@ license.workspace = true
 repository.workspace = true
 
 [dependencies]
+dashmap = { workspace = true }
 dynamo-tokens = { workspace = true }
 
 anyhow = { workspace = true }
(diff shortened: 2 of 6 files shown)


xiaguan:
Folded into #705 — one clean PR now carries the registry root-cause fix + the pinned readback; the chunked-commit pacing machinery is dropped entirely (1000+-block restores commit inline in ~2ms after the root-cause fix).

PR #709, opened 2026-07-18
feat(qwen35): add opt-in FlashInfer SM120 GDN prefill

Nyvo-io:
## Summary

- add an opt-in FlashInfer CuTe DSL SM120 AOT path for Qwen3.5-4B GDN prefill
- keep the existing Triton chunkwise pipeline as the default and fallback
- convert OpenInfer [H,K,V] recurrent state to FlashInfer [H,V,K] and back
- pass independent initial/final state buffers and convert prepared per-token log-gates to FlashInfer per-token alpha factors
- make build.rs handle vendored FlashInfer, CUDA 13 pip headers, CCCL/libcudacxx, runtime libraries, and pthread linkage

## Validation

On RTX 5090 D (SM120, driver 595.71.05):

```text
Finished release profile [optimized] target(s)

flashinfer_gdn_sm120 recurrence smoke test:
max_output_error=0.00645859
max_state_error=0.0088997
```

The smoke test uses non-zero q/k/v, per-token alpha/beta, and initial state, then compares the generated C ABI launch against a CPU token-by-token GDN recurrence. Direct nightly rustfmt and git diff --check also pass.

The C ABI smoke test supplies alpha directly. After Codex review identified a double-prefix issue in the Rust adapter, commit 9ec952f changed the adapter from exp(cumsum(g)) to per-token exp(g), matching FlashInfer's internal prefix/cumprod contract. A full Rust-path GPU rerun is still pending.

## Scope

The generated artifact is intentionally specialized to Qwen3.5-4B (32 value heads, key/value head dimension 128). Other shapes and builds without the opt-in AOT artifact retain the Triton path.

### DIFF
--- a/openinfer-kernels/Cargo.toml
+++ b/openinfer-kernels/Cargo.toml
@@ -21,6 +21,9 @@ tvm-ffi-triton-cubin = ["dep:tvm-ffi", "qwen35-4b"]
 # Qwen3.5 Triton AOT kernels (GDR chunkwise prefill) — the only feature that
 # needs Python + Triton at build time.
 qwen35-4b = []
+# Opt-in SM120 FlashInfer CuTe-DSL GDN prefill artifact. The build also
+# requires OPENINFER_FLASHINFER_GDN_AOT=1 and a CUDA 13+ CuTe toolchain.
+flashinfer-gdn-prefill = ["qwen35-4b"]
 deepseek-v2-lite = []
 # Shared MoE/MLA third-party substrate: DeepEP, DeepGEMM, and FlashMLA.
 moe = []
--- a/openinfer-kernels/build.rs
+++ b/openinfer-kernels/build.rs
@@ -1418,8 +1418,7 @@ fn compile_flashinfer_gdn_aot(
 
     let root = crate_root();
     let python = std::env::var_os("OPENINFER_FLASHINFER_PYTHON")
-        .map(PathBuf::from)
-        .unwrap_or_else(|| PathBuf::from("python3"));
+        .map_or_else(|| PathBuf::from("python3"), PathBuf::from);
     let generator = root.join("tools/flashinfer/gen_qwen35_gdn_aot.py");
     let artifact_dir = out_dir.join("flashinfer_gdn_sm120");
     // The FlashInfer Python sources are vendored as a submodule.  Keep the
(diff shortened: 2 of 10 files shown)


xiaguan:
Please add a measured FlashInfer-vs-Triton A/B on the same hardware and workloads, including TTFT and TPOT changes across representative prompt lengths and concurrency levels. We should not maintain two GDN prefill implementations long-term: select the winner based on the measurements and remove the other path before merging.

Nyvo-io:
Thanks for the review. I put a same-hardware correctness gate in front of the requested performance A/B, and the FlashInfer path does not clear it, so it is not a valid candidate to benchmark or retain.

RTX 5090 findings:

1. The first full Rust-path comparison used the same 2048-token synthetic prompt, 128 greedy output tokens, and three measured iterations. Triton was deterministic at token hash `857a65fe73b639b9`; FlashInfer was deterministic at a different hash, `cfb3b42cea601921`, with a different first token.
2. An operator-level 32-value-head regression exposed an AOT generator bug: q correctly used the upstream `[T,D,H]` TMA view, but k/v/output had incorrectly been compiled with the same view. FlashInfer requires `[D,T,H]` for k/v/output. I also found that the AOT symbol incorrectly promised `total_tokens % 64 == 0`, although real prompt lengths are not generally divisible by 64.
3. After fixing those two generator issues locally, the operator comparison improved from `max_output_diff=0.01709, max_state_diff=0.07116` to `max_output_diff=0.00528, max_state_diff=0.01883`, and passed the operator tolerance.
4. The model-level HF golden logits gate still failed decisively after that fix. Mean head-logprob delta improved from `2.1728` to `0.6768`, but the repository tolerance is `0.06`; p99 was `2.3126` versus the `0.20` tolerance. Many OpenInfer argmax tokens were absent from the HF top-64.

For comparison, the same 32-head operator test on the existing Triton path produced `max_output_diff=0.000122, max_state_diff=0.000282` and passed.

Because correctness fails before performance measurement, TTFT/TPOT numbers for this FlashInfer path would be misleading. Under the requested one-implementation policy, Triton is the winner by the mandatory correctness gate and there is no second implementation to retain. I am closing this PR rather than keeping an opt-in path that CI cannot exercise and that fails the model's golden gate on the target GPU.

I have retained the diagnostic patches, raw logs, and the 32-head operator regression locally for follow-up study. The CUDA-context-local module review point is valid for any future AOT integration, but it is moot for this withdrawn implementation.

PR #710, opened 2026-07-18
fix(qwen3): use configured device ordinal for decode overlap resources

### LINKED ISSUE #700: qwen3: green-context decode overlap creates resources on device 0 regardless of --device-ordinal
### Description

`enable_decode_overlap` hardcodes the stream device: `let device_ordinal = 0; // single-GPU path` before `OverlapStreams::create(device_ordinal, overlap)`; `create` turns it into the `CUdevice` for both overlap modes. In green-ctx mode that device drives `cuDeviceGetDevResource`/`cuGreenCtxCreate`, so with `--device-ordinal N != 0` the green contexts and their streams are created on device 0 while the model, KV cache, and compute stream live on device N — a real cross-device defect. In SharedSm (stream) mode the streams bind to the current primary context, so the wrong ordinal only mis-queries the SM count used for partition sizing — a lesser, distinct effect. Both are masked when the target card is remapped to logical 0 via `CUDA_VISIBLE_DEVICES`, which is why default setups never see it.

### Acceptance criteria

- [ ] Overlap resources are created on the configured device; the hardcoded 0 is gone, and SharedSm's SM-count query uses the configured device as well.
- [ ] A serve with green-ctx overlap and `--device-ordinal 1` (no `CUDA_VISIBLE_DEVICES` remap) on a multi-GPU box works on the configured device; today it targets device 0.

LanluZ:
The device ordinal passed to OverlapStreams::create was hardcoded to 0. This created green-context streams and SM-partition resources on device 0 even when the model, KV cache, and compute stream lived on a different device (--device-ordinal N).

Persist device_ordinal on Qwen3Executor from the model's device context at construction time, and use it in enable_decode_overlap instead of the hardcoded 0. Both the single-GPU and TP construction paths are covered.

Fixes #700.

## Description

`enable_decode_overlap` hard-coded `device_ordinal = 0` because `Qwen3Executor` did not persist the configured device ordinal after construction.

This change adds a `device_ordinal` field to `Qwen3Executor`, populated from the model's device context at construction time (both the single-GPU and TP paths), and uses it in `enable_decode_overlap` instead of the hardcoded `0`.

## Type of Change

- [ ] Documentation update

## Checklist

### DIFF
--- a/openinfer-qwen3/src/executor.rs
+++ b/openinfer-qwen3/src/executor.rs
@@ -959,6 +959,10 @@ pub struct Qwen3Executor {
     /// engine was built with events on — single-GPU, no LoRA). `None` on the
     /// plain path, where the whole feed costs nothing.
     kv_events: Option<ExecutorKvEvents>,
+    /// CUDA device ordinal this executor was built on. Used by
+    /// [`enable_decode_overlap`] to create overlap streams on the correct
+    /// device (the model, KV cache, and compute stream all live here).
+    device_ordinal: usize,
 }
 
 /// Executor-side state for the opt-in KV block-event feed.
@@ -1104,6 +1108,7 @@ impl Qwen3Executor {
             )?;
             (kv_mgr, None)
         };
+        let device_ordinal = model.device_ctx().device_ordinal;
         let metadata = Qwen3ExecutorMetadata {
             block_size: budget.block_size,
             stop_token_ids: model.config().stop_token_ids.clone(),
@@ -1186,6 +1191,7 @@ impl Qwen3Executor {
             speculative: None,
             dflash_ready_requests: HashSet::new(),
             kv_events,
+            device_ordinal,
         })
     }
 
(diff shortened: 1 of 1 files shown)


FeathBow (approved):
LGTM. Tested this on a multi-GPU GH200 node (aarch64, sm_90), inside an allocation exposing two GPUs as CUDA ordinals 0 and 1. Ran Qwen3-4B with --device-ordinal 1 --decode-overlap green-ctx at concurrency 8, which exercises async prefill during decode. The control binary’s Qwen3 sources were byte-identical to main.
 
 Previous:
  - At readiness, the server allocated 576 MiB on device 0 while the model resided on device 1. 
  - A single request still succeeded because the overlap streams are not exercised until a prefill arrives during decode. 
  - At concurrency 8, the first decode-overlap: unified step with async prefill failed with CUDA_ERROR_INVALID_HANDLE. All requests ended with finish_reason="error", with 154 error lines in the server log.

With the fix and otherwise identical arguments, device 0 remained at its 7 MiB baseline, device 1 reached 99% utilization, async prefills completed normally, and the server log contained no errors. A --device-ordinal 0 regression run also completed normally.

PR #721, opened 2026-07-18
perf(sampling): batch logprob extraction into one device reduction

### LINKED ISSUE #719: perf(sampling): GPU-reduce logprobs instead of scanning the full vocabulary on the engine thread
## Description

Qwen3 currently handles every decode row that requests logprobs by extracting a full logits row, copying it to the host, and calling the shared CPU implementation `token_logprob_from_row`. The Qwen path performs that work synchronously for each requested row in the shared step loop (decode result construction, host readback and reduction). The CPU routine makes full-vocabulary max and f64-exp-sum passes, followed by an optional top-k pass, so its cost grows with the number of logprobs rows in the batch and stalls unrelated rows before the next decode step can start.

Measured on Qwen3-4B, one sm_89 GPU, commit `f5d2bf2ad`, `--no-prefix-cache`, greedy 128-input/256-output requests, three repetitions with medians:

| concurrency | logprobs off | `logprobs=1` | `logprobs=5` |
|---|---|---|---|
| 1 | 9.82 ms TPOT / 102 tok/s | 10.69 ms / 94 tok/s | 10.71 ms / 93 tok/s |
| 8 | 10.72 ms / 730 tok/s | 17.40 ms / 456 tok/s | 17.44 ms / 455 tok/s |
| 32 | 13.08 ms / 2328 tok/s | 41.34 ms / 788 tok/s | 39.87 ms / 788 tok/s |

The added step time is approximately 0.85 ms per requested row at all three concurrency levels. In a mixed concurrency-8 batch, one logprobs request raises non-logprobs rows from 10.70 ms to 11.43 ms TPOT (+6.8%) and lowers total throughput from 733 to 686 tok/s (-6.4%). This makes logprobs a batch-wide multi-tenant latency cost rather than a cost isolated to the requesting row.

Nsight Systems counted one approximately 297 KiB D2H per requested token. CUDA copy, synchronization, allocation, and related API work total approximately 30 µs per extraction. Because the capture used `--sample=none`, it does not directly sample individual CPU functions; the approximately 0.82 ms residual is attributed to the host conversion and full-vocabulary scans by the end-to-end slope, process CPU-time delta, and the source path above. The near-identical `logprobs=1` and `logprobs=5` results show that the fixed logsumexp/full-row path dominates the incremental top-k work.

Prompt logprobs have the same scaling problem: a 1024-token prompt performs one full-row extraction per scored position and adds approximately 0.8 ms per prompt token. The device-side implementation should therefore be reusable for both decode rows and batches of prompt positions.

Suggested direction: compute logsumexp, the selected-token logprob, and the requested top-k while logits remain on the GPU, then return a compact O(k+1) result rather than an O(vocabulary) row. This is an output contract, not a requirement that the implementation be a single kernel; a numerically sound multi-stage reduction is acceptable.

Acceptance criteria:

- Decode logprobs no longer perform one full-vocabulary D2H and CPU scan per requested row.
- Prompt logprobs reuse the device reduction and do not extract every full logits row to the host position by position.
- For finite k, host readback is O(number of requested rows × (k+1)), including the selected-token result, rather than O(number of requested rows × vocabulary); full-vocabulary `-1` behavior remains governed by the frontend contract issue below.
- Selected-token logprobs, requested top-k entries, deterministic tie ordering, and error behavior match the current host reference within an explicit numerical tolerance.
- A before/after run of the concurrency and mixed-batch matrix reports TPOT/throughput, D2H count and bytes, and synchronization count. Kimi's larger whole-arena readback should be measured when it adopts the shared implementation; it is not a prerequisite for the Qwen fix.

FeathBow:
## Description

Closes #719.

Every decode row and prompt position that requested logprobs used to extract a full bf16 logits row (a ~297 KiB device copy plus a synchronous D2H) and scan the whole vocabulary on the engine thread, once per row per step. The scan runs inside the shared step loop, so the cost multiplied with the number of logprobs rows in a batch and stalled unrelated requests.

This change computes the log-softmax on the GPU instead. A new shared kernel (`logprob_topk_batch_bf16_cuda`) handles a batch of requested arena rows in one launch: per row, a fixed-tree max reduction, an exp sum accumulated in f64 for the logsumexp, the picked token's logprob, and the requested top-k selected in strictly decreasing (raw logit value, ascending token id) lexicographic order, with the logsumexp subtracted only in the returned values. `openinfer_sample::token_logprobs_batch` wraps it as the policy-level entry: one launch and a compact O(rows x (k+1)) readback per batch call — three small metadata uploads and three small readbacks replace one full-vocab copy and scan per row. Decode makes one call per step; prefill makes up to two, because first-token logits and all-position prompt logits live in different arenas. The Qwen3 decode, first-token, and prompt-logprob paths all route through the batched call; the per-row `extract_vec` + host-scan helpers are removed. The host reference `token_logprob_from_row` stays as the single-row fallback and as the conformance oracle, and now also selects its top-k by raw logits, so the ordering contract is uniform across the models that use it.

Safety and errors: the kernel-facing wrapper trusts device-resident indices, so it is `unsafe` with an explicit contract covering the index bounds and the stream-ordering of buffer traffic relative to the launch; before launching it checks buffer capacities, the arena backing size, and `i32` ranges. The safe policy entry validates every row index, picked token, and k on the host before upload, uses checked arithmetic, and rejects an active stream override, since its uploads, allocations, and readbacks are ordered on the primary stream while the kernel launch honors the override. A failed extraction now propagates as an error on the prompt-logprobs path too, instead of silently returning null entries for the whole batch.

Numerics and ordering: both paths select top-k by raw logits, breaking exact-logit ties by ascending token id, and subtract logsumexp only from returned values, so id ordering is independent of accumulator differences. On the sm_89 GPU A/B suite, ids matched exactly and values were within 5e-5 of the host reference. Per-row outputs were bitwise invariant to batch composition on the tested build.

Measured on Qwen3-4B, one sm_89 GPU, `--no-prefix-cache`, greedy 128-input/256-output requests, three repetitions with medians, same harness as the issue:

| concurrency | arm | before TPOT / tok/s | after TPOT / tok/s | after cost vs logprobs-off |
|---|---|---|---|---|
| 1 | logprobs=1 | 10.69 ms / 94 | 10.12 ms / 99 | +3% |
| 8 | logprobs=1 | 17.40 ms / 456 | 11.06 ms / 707 | +3% |
| 8 | logprobs=5 | 17.44 ms / 455 | 11.45 ms / 683 | +6% |
| 32 | logprobs=1 | 41.34 ms / 788 | 13.56 ms / 2243 | +2% |
| 32 | logprobs=5 | 39.87 ms / 788 | 13.83 ms / 2188 | +4% |

The logprobs-off arms are unchanged within run noise. In the mixed concurrency-8 batch, one logprobs request now raises non-logprobs rows by +2.6% instead of +6.8%, and an all-logprobs batch costs -3.2% throughput instead of -37%. Prompt logprobs on a 1024-token prompt drop from ~10x request latency (841-911 ms vs 84 ms) to ~1.14x (107 ms), about 0.013 ms per scored position. The prompt-logprobs arms were measured through the same temporary bridge instrumentation as the issue campaign (forcing engine-side echo when `prompt_logprobs` is present).

Nsight Systems over the same capture windows: the decode logprobs window goes from 6,949 D2H transfers / 1.87 GB and 6,951 stream synchronizations to 3,127 transfers / 0.90 MB and 1,581 synchronizations (three sub-KB readbacks per step instead of one 0.304 MB row per token), and the per-token device-to-device row copies disappear. The prompt-logprobs window goes from 12,296 D2H transfers / 3.73 GB to 68 transfers / 0.11 MB.

Validation: a GPU A/B suite (`openinfer-sample/tests/gpu_logprobs.rs`) checks the device path against the host reference — exact top-k id sequences on heavily tied rows across k values, a collapse counterexample where the logsumexp-shifted values tie but the raw logits differ, value tolerance, per-row k mixing, out-of-order arena indexing, k clamping at the vocab, rejection of out-of-range rows and picked tokens, of a declared row count beyond the arena backing, and of an active stream override, and bitwise equality of a row's outputs regardless of batch composition. `hf_golden_gate` (which compares served logprobs against HF goldens), `sampling_behavior`, and the sample-crate unit tests pass on the same GPU.

Residual cost: per decode batch call the path still performs three metadata uploads, six small device allocations, three compact readbacks, and one synchronization. At the measured residual (at most +3% TPOT for `logprobs=1` at every concurrency) this is left as is; a persistent scratch or a packed single readback is available follow-up if a later profile justifies it.

Scope: Qwen3.5 and Kimi keep the host path for now; the shared entry is model-agnostic and their adoption (including measuring Kimi's whole-arena readback) is follow-up work.

### DIFF
--- a/openinfer-kernels/csrc/shared/logprob_topk.cu
+++ b/openinfer-kernels/csrc/shared/logprob_topk.cu
@@ -1,0 +1,123 @@
+#include "common.cuh"
+
+#define LOGPROB_BLOCK 256
+
+// Fixed block size and reduction order keep outputs bitwise reproducible and
+// independent of batch composition.
+
+__device__ __forceinline__ bool rank_better(float lhs_val, int lhs_idx,
+                                            float rhs_val, int rhs_idx) {
+  return lhs_val > rhs_val || (lhs_val == rhs_val && lhs_idx < rhs_idx);
+}
+
+__global__ void logprob_topk_batch_kernel(
+    const __nv_bfloat16* __restrict__ x,
+    const int* __restrict__ row_indices,
+    const int* __restrict__ picked,
+    const int* __restrict__ top_k,
+    float* __restrict__ out_picked_lp,
+    float* __restrict__ out_topk_vals,
+    int* __restrict__ out_topk_ids,
+    int rows,
+    int n,
+    int k_max) {
+  extern __shared__ char shared_mem[];
+  float* shared_vals = reinterpret_cast<float*>(shared_mem);
+  int* shared_idxs =
+      reinterpret_cast<int*>(shared_mem + blockDim.x * sizeof(float));
+  // Aliases the arrays above: sizeof(double) == sizeof(float) + sizeof(int).
+  double* shared_sums = reinterpret_cast<double*>(shared_mem);
+
+  int slot = blockIdx.x;
+  if (slot >= rows) return;
+  const __nv_bfloat16* row_x = x + static_cast<size_t>(row_indices[slot]) * n;
+  int tid = threadIdx.x;
+
+  float local_max = -INFINITY;
+  for (int i = tid; i < n; i += blockDim.x) {
+    local_max = fmaxf(local_max, __bfloat162float(row_x[i]));
+  }
+  shared_vals[tid] = local_max;
+  __syncthreads();
+  for (int s = blockDim.x / 2; s > 0; s >>= 1) {
+    if (tid < s) {
+      shared_vals[tid] = fmaxf(shared_vals[tid], shared_vals[tid + s]);
+    }
+    __syncthreads();
+  }
+  float row_max = shared_vals[0];
+  __syncthreads();
+
+  double local_sum = 0.0;
+  for (int i = tid; i < n; i += blockDim.x) {
+    local_sum +=
+        static_cast<double>(expf(__bfloat162float(row_x[i]) - row_max));
+  }
+  shared_sums[tid] = local_sum;
+  __syncthreads();
+  for (int s = blockDim.x / 2; s > 0; s >>= 1) {
+    if (tid < s) {
+      shared_sums[tid] += shared_sums[tid + s];
+    }
+    __syncthreads();
+  }
+  float lse = row_max + static_cast<float>(log(shared_sums[0]));
+  __syncthreads();
+
+  if (tid == 0) {
+    out_picked_lp[slot] = __bfloat162float(row_x[picked[slot]]) - lse;
+  }
+
+  // Selection runs in strictly decreasing (value, -id) lexicographic order,
+  // so "after the previous winner" replaces a selected-set membership test.
+  int k = top_k[slot];
+  float prev_val = INFINITY;
+  int prev_id = -1;
+  for (int j = 0; j < k; ++j) {
+    float local_best = -INFINITY;
+    int local_id = n;
+    for (int i = tid; i < n; i += blockDim.x) {
+      float v = __bfloat162float(row_x[i]);
+      bool eligible = v < prev_val || (v == prev_val && i > prev_id);
+      if (eligible && rank_better(v, i, local_best, local_id)) {
+        local_best = v;
+        local_id = i;
+      }
+    }
+    shared_vals[tid] = local_best;
+    shared_idxs[tid] = local_id;
+    __syncthreads();
+    for (int s = blockDim.x / 2; s > 0; s >>= 1) {
+      if (tid < s) {
+        float rhs_val = shared_vals[tid + s];
+        int rhs_idx = shared_idxs[tid + s];
+        if (rank_better(rhs_val, rhs_idx, shared_vals[tid], shared_idxs[tid])) {
+          shared_vals[tid] = rhs_val;
+          shared_idxs[tid] = rhs_idx;
+        }
+      }
+      __syncthreads();
+    }
+    prev_val = shared_vals[0];
+    prev_id = shared_idxs[0];
+    if (tid == 0) {
+      out_topk_vals[static_cast<size_t>(slot) * k_max + j] = prev_val - lse;
+      out_topk_ids[static_cast<size_t>(slot) * k_max + j] = prev_id;
+    }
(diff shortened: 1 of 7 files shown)


xiaguan (approved):
Validated locally and approved.

I reviewed the complete path from the CUDA reduction through the unsafe Rust boundary and the Qwen3 prefill, decode, unified, and split-concurrent call sites. I found no correctness or API-boundary blocker.

Measured A/B against the merge base on 1x RTX 5070 Ti with Qwen3-4B, streaming completions, 128 input tokens, 128 output tokens, prefix cache disabled, one warmup and three measured iterations (median):

| Concurrency | Logprobs | Base TPOT | PR TPOT | Base throughput | PR throughput |
|---:|---:|---:|---:|---:|---:|
| 1 | 1 | 11.929 ms | 11.000 ms | 83.49 tok/s | 90.49 tok/s |
| 8 | 1 | 22.315 ms | 11.921 ms | 346.44 tok/s | 635.89 tok/s |
| 8 | 5 | 21.771 ms | 12.247 ms | 355.16 tok/s | 615.24 tok/s |
| 32 | 1 | 54.124 ms | 13.922 ms | 565.94 tok/s | 1978.83 tok/s |
| 32 | 20 | 54.345 ms | 15.917 ms | 562.31 tok/s | 1732.90 tok/s |

The no-logprobs control was effectively neutral (C1: 10.650 vs 10.647 ms TPOT; C32: 12.851 vs 12.835 ms), so the improvement is isolated to the intended path.

Local validation:
- GPU logprob parity/conformance: 7/7 passed on sm_120, including full model vocab, ties, out-of-order rows, k clamping, bounds, stream rejection, and batch invariance.
- Qwen3 HF golden gate: passed across eager, batched, CUDA graph padding, and prefix-cache replay paths.
- `openinfer-sample` library tests: 3/3 passed.
- `openinfer-qwen3` library tests: 75/75 passed.
- Release clippy for the affected crates: passed with warnings denied.

The allocation-per-call scratch and O(V*K) top-k scan are non-blocking here: the measured serving path is strongly positive, and the public facade caps K at 20.

PR #722, opened 2026-07-19
fix(logprobs): preserve None/Some(0)/Some(k) contract end-to-end (#720)

### LINKED ISSUE #720: frontend: preserve logprobs request semantics and return prompt logprobs instead of HTTP 500
## Description

The pinned vLLM engine contract represents output and prompt logprobs as two independent optional counts: `logprobs: Option<i32>` and `prompt_logprobs: Option<i32>`. `None` disables the corresponding result, `Some(0)` still requests the selected/scored token logprob without additional top entries, and `-1` denotes the full vocabulary.

OpenInfer loses that contract at the bridge boundary. `requested_logprobs` maps both `None` and `Some(0)` to `usize = 0`, and also maps `Some(-1)` to zero after the failed unsigned conversion. `GenerateRequest` has only one non-optional count for both output and prompt behavior. The frontend then hard-codes `echo: false`, drops every `PromptTokens` event, and always publishes `new_prompt_logprobs_tensors: None`, even though the engine event already carries prompt token IDs and logprobs.

This produces several externally visible failures:

1. A non-streaming completion with `echo=true` and `logprobs` asks vLLM for prompt logprobs as part of echo (conversion contract). OpenInfer completes generation without supplying them, and response assembly returns HTTP 500. This was reproduced on the benchmark head.
2. An explicit multi-token `prompt_logprobs` request reaches the same late HTTP 500 because the request is not propagated and the returned prompt tensors are always absent.
3. An explicit `logprobs=0` remains `Some(0)` in the frontend response contract, but OpenInfer turns it into the internal disabled value, so response assembly expects a chosen-token logprob that the engine never emits. This is different from omitting the field.
4. The southbound `-1` full-vocabulary sentinel is silently treated as disabled. If full-vocabulary logprobs are intentionally unsupported, this must be an early validation error rather than silent semantic degradation.

The one-line instrumentation change used for the performance campaign (`echo: sampling_params.prompt_logprobs.is_some()`) is not a complete fix: output and prompt counts may differ, optionality still disappears, and `PromptTokens` still never reaches the wire response.

Acceptance criteria:

- The bridge preserves distinct `None`, `Some(0)`, positive-k, and `-1` behavior; unsupported values fail before GPU work begins with an actionable client error.
- Output `logprobs` and `prompt_logprobs` can carry different counts without overloading one `usize` or one batch-level echo flag.
- Supported non-streaming `echo=true + logprobs`, explicit `prompt_logprobs`, and explicit `logprobs=0` requests return the requested values instead of a late HTTP 500.
- `PromptTokens` IDs and logprobs are encoded into the existing prompt-logprobs wire field with the pinned vLLM response semantics, including the leading prompt position that has no predecessor logprob.
- Focused frontend contract coverage distinguishes omitted, zero, positive, and unsupported full-vocabulary requests; an HTTP regression case covers the previously failing multi-token prompt response.
- This correctness change may retain the existing CPU calculation initially; the GPU reduction belongs to the separate performance issue above.

n-WN:
Closes #720

## Summary

The engine contract collapsed `logprobs: 0` and "logprobs disabled" into one value, dropped `prompt_logprobs` at the wire boundary, and the bridge silently discarded `TokenEvent::PromptTokens`. As a result, `echo=true` + `logprobs` and explicit `prompt_logprobs` requests hit HTTP 500 in response assembly, and `logprobs=-1` was silently mapped to 0 instead of failing fast.

## Changes

**Engine contract** (`openinfer-engine`)
- `GenerateRequest.logprobs: usize → Option<usize>` — `None` = disabled, `Some(0)` = chosen-token logprob only, `Some(k)` = chosen + k alternatives
- `GenerateRequest.echo: bool → prompt_logprobs: Option<usize>`

**vLLM frontend**
- `wire.rs`: `requested_logprobs` preserves `None`/`Some(0)`/`Some(k)`; `prompt_logprobs` is now read; `-1` (full-vocab sentinel) rejected early via `unsupported_sampling` for both fields, before any GPU work
- `bridge.rs`: submit maps `prompt_logprobs=Some(k)` to the engine prompt-logprobs request; `TokenEvent::PromptTokens` is encoded into `new_prompt_logprobs_tensors` with vLLM semantics (scored positions only, leading `None` position omitted; single-token prompt sends no payload)

**Qwen3** (the only crate that computes prompt logprobs)
- scheduler/plan/resolve gate on `prompt_logprobs.is_some()`; all echo invariants kept: whole-prompt single forward, no chunking, no prefix-cache match, dedicated Prefill routing, all-position logits

**Honor-or-reject** for crates without prompt-logprob support (#236 precedent)
- Kimi-K2: reject at admission (lm_head runs on the last position only); row options take `Option<usize>`
- Qwen3.5: reject at admission instead of emitting all-`None` `PromptTokens`
- GLM5.2: admission message now names completion/prompt logprobs
- DeepSeek-V2-Lite: reject prompt logprobs (was a silent all-`None` echo stub)

**Tests**
- wire: `None`/`0`/`k`/`-1` matrix for both fields
- bridge: `PromptTokens` → wire payload (multi-token and single-token cases)
- sim `frontend_e2e`: HTTP regression for `echo`+`logprobs` (was HTTP 500), explicit `prompt_logprobs`, `logprobs=0`, `-1` early rejection, single-token prompt
- model crates: scheduler/plan/executor tests migrated; batch-invariance gates updated for `Option` step-item logprobs

CPU logprob computation is unchanged; the GPU-side reduction is #719.

## Verification

| Gate | Result |
|------|--------|
| `cargo check` + `cargo clippy --all-targets` (9 crates) | ✅ clean |
| `openinfer-vllm-frontend` unit tests | ✅ 27 passed |
| `openinfer-sim --test frontend_e2e` (HTTP regression, CI gate) | ✅ 19 passed |
| model-crate `--lib` tests (qwen3/qwen35/kimi/glm52/deepseek) | ✅ 150 passed |
| qwen3 `dflash_speculative_gate` (H100) | ✅ 4 passed |
| qwen3 `hf_golden_gate` — logprobs vs HF golden, bf16 tolerance (H100) | ✅ 1 passed |

qwen35 / kimi-k2 golden gates self-skip on this box (model weights not present); their migrated test code compiles under `clippy --all-targets`.


openinfer-vllm-frontend/src/bridge.rs
```diff
@@ -518,8 +519,31 @@ fn reduce_request(
...
+                // The wire payload carries one position per *scored* prompt
+                // token (every prompt token except the leading one, which has
+                // no predecessor logits); the frontend prepends the leading
+                // `None` itself. A single-token prompt scores nothing, so we
+                // publish no payload and let the frontend's one-token branch
+                // answer instead.
+                if ids.len() > 1 {
+                    let positions: Vec<PositionLogprobs> = ids
+                        .iter()
+                        .zip(logprobs.iter())
```
FeathBow, line 532: `zip` silently truncates when `ids.len() != logprobs.len()`, while the following `filter_map` drops any non-leading position whose logprob is `None`. Both cases produce fewer wire positions than prompt positions, violating the positional contract and potentially shifting subsequent values.

We could validate equal lengths and permit `None` only for the protocol-defined leading position. Any other mismatch or missing value should terminate the request with an explicit error instead of degrading the payload.

openinfer-qwen3/src/scheduler/resolve.rs
```diff
@@ -128,7 +128,7 @@ fn resolve_prefill_outputs(
             continue;
         }
 
-        if req.echo {
+        if req.prompt_logprobs.is_some() {
```
FeathBow, line 131: Once prompt logprobs are requested, this path must not synthesize a successful all-None result. `extract_prompt_logprobs` converts both device extraction and D2H failures to `None` via `.ok()`, and the fallback below turns that missing result into `vec![None; prompt_len]`; the bridge then drops those positions.

IMO We could make extraction and result propagation fallible and emit `TokenEvent::Error` instead of converting real computation failures into missing logprobs.

FeathBow (commented):
Both upstream fixes required by this integration have landed:

- vllm-project/vllm#49111
- vllm-project/vllm#49113

However, all four vLLM Rust crates in this PR and current main remain pinned to 8e61b646, 462 commits behind f25953c. The OpenInfer build therefore still contains both the single-token prompt-logprobs failure and the zero-column payload panic. Consequently, the E2E coverage added by #722 cannot establish that the actual vLLM integration is correct.

Let's keep #722 blocked and revisit it after vLLM publishes the next version and OpenInfer refreshes all four crate revisions plus the lockfile. The Rust frontend and integration gates must then be rerun against that updated pin.

FeathBow:
main has advanced quite a bit, so this might require a redesign.

FeathBow:
## Description

Closes #720

Original contribution by @n-WN; the original commit is preserved unchanged.

Completion and prompt logprob requests lose information when zero is treated as disabled or prompt-score events are dropped. Requests such as `logprobs=0`, `echo=true` with logprobs, and explicit `prompt_logprobs` can consequently fail during response assembly instead of returning aligned scores.

This change preserves independent disabled / zero / top-k counts through the current request contracts and both frontend bridges. The stepped bridge encodes prompt scores positionally and aborts malformed payloads per request. Legacy model lines reject prompt scoring at admission. Qwen3 retains its GPU reductions and derives all-position scoring from requests, using whole, uncached, unchunked prefill. Model paths that cannot honor a requested score mode reject it explicitly, including zero-count requests. The shared engine-core boundary rejects every negative count before converting to usize, including requests that bypass HTTP validation. Qwen3.5 snapshots preserve each requested row together with its top-k count through sampling.

### Verification

Validation used Single GPU (sm_89, x86_64), with Qwen3-4B and Qwen3.5-4B. Backend unit runs had a 60-second hard timeout.

| Check | Result |
| --- | --- |
| Strict all-target Clippy | Frontend/sim/Qwen3 passed |
| Frontend/sim unit tests | 77 passed, including direct engine-core negative-count rejection before submission |
| Simulated HTTP protocol tests | 18 passed after removing duplicate echo coverage, including zero-count streaming |
| Qwen3 unit tests | 100 passed |
| Real Qwen3-4B gate | Three independent runs passed: cached-prefix bypass and six teacher-forced comparisons per run; maximum difference 0.061958 nat, with the bound tightened from 0.20 to 0.10 nat |
| Real Qwen3 HTTP contract | The nine retained cases passed under default and fixed numeric policies; earlier receipts also include the removed echo case |
| Qwen3.5 release all-target check and build | Passed again after pairing snapshots with top-k |
| Qwen3.5 unit tests | 98 passed; 7 existing tests ignored |
| Real Qwen3.5-4B HTTP contract | 4 cases passed after snapshot cleanup: completion counts omitted / 0 / 2 and zero-count streaming; non-streaming text and score payloads matched the earlier receipts exactly |
| Formatting and diff checks | Passed |

Same-machine A/B against main `4043d50a`, using the same model and `--batch-invariant --no-prefix-cache`: six retained non-streaming cases changed from HTTP 500 to 200; zero-count streaming changed from an error event to six scored tokens. Existing unscored/top-k control outputs and scores matched exactly.

Retained default-policy A/A records matched exactly. Default-policy A/B preserved generated text but differed by up to 0.0998 nat in chosen-token scores, so exact score equivalence is established only for the fixed policy. No performance improvement is claimed.

Final model validation is scoped to Qwen3 and Qwen3.5. The final DSV2 cleanup only combines the existing refusal predicates and updates the trace fixture text; it was statically reviewed without another model run. Earlier all-target release checks also passed for Gemma 4, Kimi-K2, DeepSeek-V2-Lite and Kimi-K3 with an SM100 target. Existing K3 kernels that require SM100f compile as unsupported stubs under CUDA 12.9, so this is compilation evidence only. GLM compilation remains blocked by the CUDA toolchain. Multi-GPU TP/EP and HF golden fixture gates were not run during this migration. The simulated HTTP tests establish protocol behavior; the real-model checks provide separate serving and score evidence.

Issue #727, opened 2026-07-19
qwen35: add adaptive decode-priority scheduling for mixed prefill/decode load


CAICAIIs:
## Summary

Qwen3.5 currently uses a fixed scheduling rule when active decode requests and pending prefill work coexist. This can protect throughput in some cases, but it can also create active-decode stalls under mixed workloads where long cold prefills arrive while existing requests are decoding.

This issue tracks an adaptive scheduler policy for Qwen3.5 that can reduce mixed prefill/decode ITL stalls without making the default serving path pay a broad TTFT or throughput regression.

## Motivation

The current fixed policy is too coarse for a default latency policy:

- Always prioritizing prefill can stall active decode behind long prompt work.
- Always overlapping or shrinking prefill can improve ITL tail, but may regress TTFT and/or throughput.
- Static request fields such as `max_tokens` are not enough to decide whether a request should trigger decode protection.

A default policy should react to runtime state, not just a single CLI mode or fixed threshold.

## Evidence

On a single RTX 5090 Qwen3.5 serving setup, opt-in stream overlap showed that active-decode stalls can be reduced, but it also exposed why the behavior should not simply become the default:

- Mixed-load ITL p99 improved substantially when overlap was enabled.
- Standard HTTP serving cells remained sensitive to TTFT and throughput tradeoffs.
- A `max_tokens >= 1024` gate preserved short-output cells better, but still regressed median TTFT on a long-output concurrency cell.
- Active-only smaller prefill chunks improved some mixed-load tail metrics, but regressed c16 / QPS serving throughput and TTFT.

These results suggest the bottleneck is schedulable, but the current policies are not yet a safe default.

## Proposed Direction

Add an adaptive Qwen3.5 scheduler policy that decides per scheduler tick how to balance decode and prefill work. Candidate signals include:

- active decode batch size
- pending prefill token count
- pending request queue age
- active request generated tokens / remaining budget
- recent ITL stall history
- prefill chunk size and estimated stall cost
- starvation/fairness limits for pending prefill

The policy should be able to choose among existing execution options, such as:

- normal prefill / decode / unified execution
- smaller prefill chunks while decode is active
- decode-priority ticks
- opt-in stream overlap as an execution mechanism when the runtime state justifies it

## Acceptance Criteria

A default `auto` policy should meet all of the following before becoming the default:

- No material regression on standard HTTP serving cells such as 1024/256 c1 and c16.
[...]

PR #730, opened 2026-07-20
feat(qwen35): add adaptive decode-priority scheduler policy

CAICAIIs:
## Summary

Addresses #727.

This PR adds a Qwen3.5 scheduler policy switch for prefill/decode balancing:

- adds `Qwen35SchedulerPolicy::{Auto, Off}` with `auto` as the default single-GPU Qwen3.5 policy
- keeps an explicit `--qwen35-scheduler-policy=off` mode to preserve the fixed chunked-prefill behavior
- lets Qwen3.5 `--max-batch` accept non-bucket capacities such as `5`; allocation still rounds to the next graph bucket internally, while admission uses the requested slot count
- records Qwen3.5 mixed-load `max_batch`, starvation warnings, and generated-token length/hash sanity fields in `bench_serving mixed`
- documents the policy, evidence, and claim boundary in `docs/models/qwen35/adaptive-scheduler-policy.md`

## Policy shape

The `auto` policy keeps the existing fixed chunk path for standard long-output serving cells and changes only narrow runtime states:

- active request near completion: run a decode-priority tick before continuing FIFO prefill
- low-pressure short-output prefill: allow one whole-prefill step to avoid repeatedly creating medium stalls
- backlog/deferred pressure or very long prefill: keep fixed chunking
- tensor-parallel Qwen3.5 remains fixed-policy because TP Phase 1 does not support unified mixed prefill+decode

## Validation

Environment used for GPU validation:

- 1x NVIDIA GeForce RTX 5090
- NVIDIA driver `595.71.05`
- CUDA toolkit `12.8`
- Rust `1.99.0-nightly`
- OpenInfer source: upstream/main `8dd3953` plus this patch
- Qwen3.5 model: public `Qwen/Qwen3.5-4B` snapshot, `config.json` sha256 `ddc63e1c717afa86c865bb5e01313d89d72bb53b97ad4a8a03ba8510c0621670`
- Qwen3.5 feature: `qwen35-4b`

Checks run:

```bash
cargo fmt --all -- --check
git diff --check
cargo test -p openinfer-qwen35-4b --features qwen35-4b adaptive_prefill_budget -- --nocapture
cargo test -p openinfer-server --features qwen35-4b qwen35 -- --nocapture
cargo build --release -p openinfer-server --features qwen35-4b
cargo build --release --bin bench_serving --features qwen35-4b
OPENINFER_TEST_MODEL_PATH=<model> \
```

Result: all checks above passed. The single-GPU Qwen3.5 scheduler e2e passed; TP2 stayed ignored on the one-GPU validation host.

## Benchmark notes

Standard request A/B, `1024/256`, warmup 1, iters 3, `--max-batch 16`:

| Policy | Cell | TTFT p50 ms | steady TPOT p50/p99 ms | request tok/s | output length | hash0 |
| --- | --- | ---: | ---: | ---: | --- | --- |
| `auto` | c1 | 49.816 | 6.942 / 7.022 | 140.68 | 256-256 | `0827a7035c7b7a89` |
| `off` | c1 | 50.176 | 6.949 / 7.150 | 140.37 | 256-256 | `0827a7035c7b7a89` |
| `auto` | c16 | 508.187 | 9.763 / 58.321 | 76.73 | 256-256 | `0827a7035c7b7a89` |
| `off` | c16 | 508.343 | 9.770 / 58.255 | 76.76 | 256-256 | `0827a7035c7b7a89` |

Mixed-load A/B, `--max-batch 5`, `bg_concurrency=4`, background `512/2048`, injection `4096/1`, `qps=0.5`, 5 cold injections:

| Policy | all ITL p50/p99/max ms | steady p99/max ms | stall p50/p99/max ms | stall gaps | warnings |
| --- | ---: | ---: | ---: | ---: | --- |
| `auto` | 7.386 / 7.845 / 196.554 | 7.754 / 58.553 | 7.941 / 196.552 / 196.554 | 60 / 4944 | none |
| `off` | 7.389 / 57.211 / 65.232 | 7.794 / 58.536 | 57.141 / 65.229 / 65.232 | 120 / 4904 | none |

Mixed output sanity:

| Policy | background output length | background hash0 | injection output length | injection hash0 |
| `auto` | 1236-1238 | `dea24e27083abe47` | 1-1 | `ec2064181e172bb6` |
| `off` | 1226-1228 | `40933b449d599567` | 1-1 | `ec2064181e172bb6` |

The retained mixed cell shows the intended low-pressure short-output p99 improvement without changing the representative `1024/256` standard cells. The `auto` mixed max remains a whole-prefill-sized event by design; this PR does not claim universal tail/max improvement.

## Claim boundary

- This is scheduler-policy and benchmark-instrumentation work for Qwen3.5.
- It does not claim vLLM parity.
- It does not make stream overlap the default.
- It does not change prefix-cache behavior.
- It does not claim production readiness.
- The retained evidence covers representative standard cells and one mixed-load A/B. A wider QPS pressure sweep can run separately before stronger default-readiness wording.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -57,6 +57,7 @@ Organized by domain (model line / subsystem / playbook / lesson) instead of by l
 | `models/qwen35/tp-design.md` | Qwen3.5 TP design: Phase 1 is eager dense TP on Qwen3's controller/worker runtime; validate TP2 first, fail closed for indivisible degrees and TP+CUDA Graph, shard dense full-attention/MLP, and leave sharded linear/GDR state to follow-up. |
 | `models/qwen35/tp-implementation.md` | Qwen3.5 TP Phase 1 implementation record: eager dense TP2 worker/scheduler path, short/long HF logits gates, scheduler e2e, and real OpenAI-compatible HTTP smoke pass; remaining TP work is kept as follow-up, not a Phase 1 claim. |
 | `models/qwen35/mixed-load-itl-470.md` | Issue #470: full cold `--max-batch 8/bg=4` matrix on RTX 4090 (24/24 valid) + starvation negative control. Qwen3.5 is not immune; chunking bounds max/per-step stall but raises p99 at low QPS (~14→~80–92ms) and pulls p99/max back from the prefill wall to the chunk wall at high load; `qps·prefill_s≳1` is a throughput wall (chunking can't fix it, and ON's +15% TTFT can trip it earlier). The old "p99 immunity" was a slot-starvation artifact. |
+| `models/qwen35/adaptive-scheduler-policy.md` | Issue #727 adaptive scheduler policy record: default `off`, opt-in `auto`, hard `--max-prefill-tokens` cap, TP `auto` rejection, and pre-review whole-prefill benchmark tradeoff retained as non-default evidence. |
 
 ## models / glm52
 
--- a/docs/models/qwen35/adaptive-scheduler-policy.md
+++ b/docs/models/qwen35/adaptive-scheduler-policy.md
@@ -1,0 +1,149 @@
+# Qwen3.5 Adaptive Scheduler Policy
+
+> **TL;DR:** Issue #727 now lands Qwen3.5 scheduler policy plumbing with
+> conservative defaults: `off` remains the default, `auto` is explicit opt-in,
+> `--max-prefill-tokens` remains a hard per-step cap, and TP rejects `auto`
+> instead of silently downgrading to `off`.
+>
+> **Last touched:** 2026-07
+
+## Preparation
+
+- **Read**:
+  - `docs/index.md` - Qwen3.5 roadmap, scheduler, benchmark, and evidence docs are the relevant route.
+  - `docs/models/qwen35/roadmap.md` - #469 is the current HTTP boundary; #470/#727 must keep mixed-load evidence separate from serving parity.
+  - `docs/subsystems/scheduler/scheduler.md` - the scheduler is single-threaded GPU ownership with chunked prefill and unified prefill+decode.
+  - GitHub #727 - asks for adaptive decode-priority policy work with explicit off mode and standard-cell regression protection.
+  - GitHub PR #730 review - requires `off` as the default, `--max-prefill-tokens` as a hard cap, and explicit TP rejection for `auto`.
+- **Relevant history**:
+  - `docs/benchmarks/qwen35-4b-serving-vllm-rtx5090-2026-07.md` - standard HTTP cells and QPS pressure are the current serving regression boundary.
+  - Prior stream-overlap exploration stayed out of scope; #727 reuses the existing unified/decode/chunk controls before adding any new execution stream.
+- **Plan**:
+  1. Add a small Qwen3.5 scheduler policy enum with default `off` and explicit opt-in `auto`.
+  2. Move the adaptive decision into pure scheduler-plan helpers with unit tests for fixed, hard-cap, final-chunk, and decode-finish protection cases.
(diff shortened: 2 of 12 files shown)


xiaguan (changes requested):
Request changes. The scheduling seam and `Off` compatibility path are useful, but two behavior blockers make `Auto` unsafe as the default.

1. `Auto` violates the documented `--max-prefill-tokens` cap. `openinfer-server/src/config.rs:179-183` defines this value as the cap on prompt tokens forwarded in one scheduler step. However, `choose_prefill_budget` can return `front.remaining_tokens` at `openinfer-qwen35-4b/src/scheduler/plan.rs:96-104`, up to eight times `base_budget`. For example, with any active decode, no second prefill/deferred request, `--max-prefill-tokens=128`, and a 1024-token prompt with `max_tokens=1`, the scheduler forwards all 1024 prompt tokens in one tick. Please preserve the configured value as a hard cap, or rename/redefine the option through an explicit API change; an adaptive policy must not silently exceed it.

2. The whole-prefill decision ignores active decode pressure. `choose_prefill_budget` receives the entire active slice but never uses `active.len()`. Thus one active decoder and 63 active decoders make the same whole-prefill decision, and `scheduler.rs:953-987` makes every active decoder wait for that unified step. The only mixed A/B covers decode width 4, yet it already moves max ITL from 65.232 ms (`off`) to 196.554 ms (`auto`) while improving p99. That is a tradeoff, not enough evidence for a globally safer default. Issue #727 also requires the QPS 8/12/16, long-output, and wider active-decode validation matrix, which is still missing.

Please keep `Off` as the enum, legacy API-entrypoint, and CLI default until that matrix establishes an explicit SLA objective. `Auto` can remain opt-in meanwhile. Before flipping the default, its decision should account for active decode width (or use a measured stall budget) rather than treating “no second prefill request” as low pressure.

xiaguan:
One additional configuration blocker from the full path audit: TP accepts `--qwen35-scheduler-policy auto` but silently runs `Off`. The server forwards the policy (`openinfer-server/src/main.rs:303-318`), while the multi-device branch returns before consuming it (`openinfer-qwen35-4b/src/lib.rs:187-202`) and the TP scheduler hard-codes `Off` (`scheduler.rs:227-236`). The documentation acknowledges TP is fixed-policy, but the CLI/API reports no rejection or downgrade. Please fail fast for TP + `Auto`, or make the effective TP policy explicitly `Off` at configuration resolution.

CAICAIIs:
> One additional configuration blocker from the full path audit: TP accepts `--qwen35-scheduler-policy auto` but silently runs `Off`. The server forwards the policy (`openinfer-server/src/main.rs:303-318`), while the multi-device branch returns before consuming it (`openinfer-qwen35-4b/src/lib.rs:187-202`) and the TP scheduler hard-codes `Off` (`scheduler.rs:227-236`). The documentation acknowledges TP is fixed-policy, but the CLI/API reports no rejection or downgrade. Please fail fast for TP + `Auto`, or make the effective TP policy explicitly `Off` at configuration resolution.

Thanks, updated the PR to keep `Off` as the default, make `Auto` opt-in, preserve `--max-prefill-tokens` as a hard cap, and reject Qwen3.5 TP + `auto` instead of silently downgrading. Docs are updated accordingly, and CI is green on the refreshed head.

xiaguan (approved):
LGTM. The refreshed head addresses all requested changes:

- `Off` remains the enum, legacy API, scheduler, and CLI default.
- `Auto` is explicit opt-in and never exceeds `--max-prefill-tokens`; the whole-prefill expansion was removed.
- Qwen3.5 TP + `Auto` now fails fast at both CLI validation and library entry points.
- The relevant default, cap, and TP validation tests are present, and CI is green.

The wider QPS, long-output, and active-width matrix remains a future requirement before any default-readiness change, as documented; it does not block this opt-in plumbing.

PR #731, opened 2026-07-20
perf(core): parallelize weight prefetch for single-rank Qwen loads

FeathBow:
## Description

On client-cold loads from the measured Lustre filesystem, sequential tensor copies fault safetensors mmap pages through one demand-driven stream. WeightPrefetch runs up to eight background readers over 16 MiB chunks to populate the page cache ahead of that copy loop. The mmap copy path remains authoritative. Dropping the guard cancels and joins every worker, while setup and worker failures are reported once with the first cause retained.

Prefetch is enabled only at the single-rank Qwen3 and Qwen3.5 whole-model call sites. TP and drafter loaders remain unchanged to avoid duplicate full-checkpoint reads.

Measured on a GH200 node (aarch64, sm_90) with model weights on a Lustre parallel filesystem, 3 cold + 2 warm repetitions per arm; cold = client page cache evicted via `posix_fadvise(DONTNEED)` before every repetition. The paired campaign ran an earlier build with the same read loop; the final RAII/call-site form received separate sanity validation below.

| model | arm | cold: weights / engine ready | warm: weights / engine ready |
|---|---|---|---|
| Qwen3.5-27B (55.6 GB) | off | 5.3 s / 10.4 s | 1.26 s / 6.29 s |
| | on | 4.0 s / 9.0 s | 1.36 s / 6.31 s |
| Qwen3.5-9B (19.3 GB) | off | 1.63 s / 5.93 s | 0.49 s / 4.74 s |
| | on | 1.41 s / 5.67 s | 0.50 s / 4.78 s |

Scope of these numbers: `fadvise` evicts the client cache but not the storage servers', and this filesystem was serving ~9 GB/s from server cache during the run, so they cover the client-cold/server-warm regime only. Genuinely cold storage is not measured and no claim is made about that regime.

## Validation

- `hf_golden_gate` passes for qwen3 and qwen35 on the campaign build, covering both model paths after the loader change.
- Final-form sanity, Qwen3.5-27B cold/warm: prefetch engages (11 shards, 55.6 GB), and every run reaches server ready with zero prefetch threads alive and zero failure warnings.
- Final-form parity, Qwen3-4B warm startup against a main-loader binary: engine ready 4.49–4.50 s (main loader) vs 4.50–4.52 s (prefetch), weight phase 0.38 s vs 0.39 s; the prefetch log line appears only in the patched arm.
- Workspace build and `cargo clippy` clean on the final form.

### DIFF
--- a/openinfer-qwen35-4b/src/weights.rs
+++ b/openinfer-qwen35-4b/src/weights.rs
@@ -9,8 +9,9 @@ use std::time::Instant;
 use super::config::{Config35, LayerType, TensorParallelConfig};
 use openinfer_core::tensor::{DeviceContext, DeviceMatrix, DeviceVec, HiddenStates};
 use openinfer_core::weight_loader::{
-    deserialize_shards, load_shard_info_fixed, load_tensor_1d, load_tensor_1d_f32, load_tensor_2d,
-    load_tensor_2d_col_shard, load_tensor_2d_row_shard, mmap_shards, precompute_rope,
+    WeightPrefetch, deserialize_shards, load_shard_info_fixed, load_tensor_1d, load_tensor_1d_f32,
+    load_tensor_2d, load_tensor_2d_col_shard, load_tensor_2d_row_shard, mmap_shards,
+    precompute_rope,
 };
 
 /// Full attention layer weights (8 layers in Qwen3.5-4B).
@@ -221,6 +222,8 @@ impl Qwen35Model {
 
         let (shard_paths, weight_map) = load_shard_info_fixed(model_path)?;
         debug!("Loading {} safetensor shard(s)", shard_paths.len());
+        let prefetch =
+            (tensor_parallel.world_size == 1).then(|| WeightPrefetch::spawn(&shard_paths));
         let mmaps = mmap_shards(&shard_paths)?;
         let shards = deserialize_shards(&mmaps)?;
 
@@ -456,6 +459,7 @@ impl Qwen35Model {
         )?;
 
         ctx.sync()?;
+        drop(prefetch);
         info!(
             "GPU model loaded in {:.0}ms",
             t_gpu.elapsed().as_secs_f64() * 1e3
--- a/openinfer-core/src/weight_loader.rs
+++ b/openinfer-core/src/weight_loader.rs
@@ -3,11 +3,14 @@
 use anyhow::Result;
 use cudarc::driver::CudaSlice;
 use half::bf16;
-use log::info;
+use log::{info, warn};
 use memmap2::Mmap;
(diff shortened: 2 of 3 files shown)


openinfer-qwen35-4b/src/weights.rs
```diff
@@ -456,6 +459,7 @@ impl Qwen35Model {
         )?;
 
         ctx.sync()?;
+        drop(prefetch);
```
xiaguan, line 462: 这个好像不用手动 drop 吧？这个语义应该是 prefetch.wait()? 等待他结束?

openinfer-qwen35-4b/src/weights.rs
```diff
@@ -221,6 +222,8 @@ impl Qwen35Model {
 
         let (shard_paths, weight_map) = load_shard_info_fixed(model_path)?;
         debug!("Loading {} safetensor shard(s)", shard_paths.len());
+        let prefetch =
+            (tensor_parallel.world_size == 1).then(|| WeightPrefetch::spawn(&shard_paths));
```
xiaguan, line 226: 这里 tp == 1 是为什么？

xiaguan:
话说普通 fs 这个会有效果吗？

openinfer-qwen35-4b/src/weights.rs
FeathBow: After ctx.sync() the remaining prefetch is useless, and waiting would delay ready whenever it falls behind. The explicit drop stops the workers before the KV pool / graph setup that follows.

FeathBow:
> 话说普通 fs 这个会有效果吗？

Should work on any FS (it just populates the page cache), but the gain scales with how far single-stream demand faulting sits below multi-stream throughput — large on Lustre, likely small on local NVMe with good readahead.

openinfer-qwen35-4b/src/weights.rs
xiaguan: 我的想法是，把它提出来作为一个 pub 的接口去调用？

FeathBow: Makes sense. For this PR, though, a named method would only wrap the existing cancel-and-join Drop path. I’d prefer to keep the RAII API here, but we could discuss whether a distinct public operation would be useful.

openinfer-qwen35-4b/src/weights.rs
FeathBow: TP prefetch is not implemented in this PR. Since the loader runs once per rank while the current prefetch scans the full shard set, enabling it here would rescan the checkpoint world_size times. A shared or rank-0 prefetch coordinator could be considered as follow-up work.

PR #734, opened 2026-07-21
perf(kimi-k2): compact local top-1 readback into one pinned transfer

### LINKED ISSUE #716: perf(kimi-k2): compact local top-1 readback into one pinned transfer
## Description
Kimi's local top-1 path synchronizes the compute stream and then copies the full-capacity ID and value buffers through two pageable D2Hs, although only `active_rows` are consumed. This adds a full-stream barrier and two host transfers to the decode tail.

Pack each active row's ID and value into one compact device packet, keep a persistent pinned landing buffer in the existing decode scratch, copy only the active prefix, and wait on that transfer's completion event.

Acceptance criteria:

- One active-prefix D2H per decode step instead of two full-capacity D2Hs.
- No full-stream synchronization solely for top-1 readback.
- Token IDs, values, and cross-rank selection semantics remain unchanged.
- A before/after trace reports D2H and synchronization counts plus TPOT; the issue does not assume a speedup before measurement.

LanluZ:
## Description

Fixes #716

Kimi's batched local top-1 readback synchronized the whole compute stream and then copied the full-capacity ID and value buffers through two pageable D2Hs, although only `active_rows` are consumed. This adds a full-stream barrier and two host transfers to the decode tail.

This change:

- Adds a dedicated `pack_top1_packets_cuda` kernel that interleaves each active row's `{i32 id, bf16 value}` into one compact 8-byte device packet (shared argmax kernels are untouched).
- Keeps a persistent device packet buffer (`CudaSlice<u8>`) + pinned host landing buffer (`PinnedHostSlice<u8>`) in `SamplingScratch`.
- Issues **one active-prefix D2H** (`active_rows * 8` bytes) into the pinned buffer instead of two full-capacity pageable copies.
- Waits on the transfer's completion event (cudarc `PinnedHostSlice` event) instead of `cudaStreamSynchronize` — no full-stream barrier solely for top-1 readback.

Token IDs, values, and cross-rank selection semantics are unchanged; the row range checks are preserved verbatim.

I was unable to complete end-to-end testing locally. Requesting the reviewer / maintainer to verify.

## Type of Change

- [ ] Documentation update

## Checklist

### DIFF
--- a/openinfer-kernels/csrc/shared/argmax.cu
+++ b/openinfer-kernels/csrc/shared/argmax.cu
@@ -429,4 +429,27 @@ void markov_step_argmax_cuda(const __nv_bfloat16* base,
                      partial_indices, out_tokens, sampled_tokens, block_size,
                      step, rows, tiles_per_row);
 }
+
+struct Top1Packet {
+  int id;
+  __nv_bfloat16 value;
+};
+
+__global__ void pack_top1_packets_kernel(const int* ids,
+                                         const __nv_bfloat16* values,
+                                         Top1Packet* packets, int rows) {
+  int row = blockIdx.x * blockDim.x + threadIdx.x;
+  if (row < rows) {
+    packets[row].id = ids[row];
+    packets[row].value = values[row];
+  }
+}
+
+void pack_top1_packets_cuda(const int* ids, const __nv_bfloat16* values,
+                            void* packets, int rows, cudaStream_t stream) {
+  int threads = 128;
+  int blocks = (rows + threads - 1) / threads;
+  pack_top1_packets_kernel<<<blocks, threads, 0, stream>>>(
+      ids, values, reinterpret_cast<Top1Packet*>(packets), rows);
+}
 }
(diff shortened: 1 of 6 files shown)


xiaguan:
Thanks for working on this. The implementation direction matches #716, and a little runtime evidence would help us validate both correctness and the intended improvement. The current description notes that end-to-end testing was not completed; the generic CI checks also do not exercise the Kimi-K2 runtime path.

Could you please add results from the same hardware/software/configuration for:

1. Kimi-K2 E2E correctness on the affected greedy local-top1 path, including the multi-rank/cross-rank selection path if that is the production configuration.
2. A before/after trace showing D2H call count, D2H bytes, and synchronization/event behavior per decode step, as requested in #716.
3. A before/after TPOT benchmark with the exact command/configuration and representative active-row counts. Please include neutral or negative cases as well.

These results will let us confirm that the runtime behavior is preserved and evaluate the optimization in its intended context. Once they are attached, we can continue with the detailed review.

Issue #738, opened 2026-07-22
Qwen3: first content token deferred to first decode step adds ~one decode step to TTFT


xiaguan:
## Summary

On Qwen3, the first **content** token is not emitted during the prefill step — it is deferred to the following decode step. This adds roughly one decode step (~TPOT) to time-to-first-token for every request.

## Measurement

Single request, Qwen3-4B, local single GPU, prompt ~500 tokens (prefix-cache warm). Nanosecond timestamps were inserted at each hop of the token egress path (scheduler emit → bridge demux receive → bridge send → ZMQ write → `vllm_server` SSE layer) and correlated per request.

Client-observed TTFT ≈ **24.0 ms**, decomposed:

| segment | measured |
|---|---|
| queue (scheduler admission) | 0.02 ms |
| prefill (GPU forward) | 11.7 ms |
| openinfer egress (scheduler → bridge → ZMQ → vllm_server SSE layer) | 0.4 ms |
| **wait for first content token (one extra decode step)** | **~10.4 ms** |

The `vllm_server` SSE layer yields each `ChatEvent` in <0.05 ms, so serialization/transport is not the cost. The decisive evidence is the `ChatEvent` arrival sequence for a single request:

```
Start
LogprobsDelta   +0.098 ms
BlockStart      +10.382 ms   <-- one decode step
BlockDelta      +0.056 ms    <-- first content token
```

and two `TokenEvent::Token` arrivals at the bridge for the same request, **10.7 ms apart** (= one decode step).

## Root cause

`resolve_prefill_outputs` routes the prefill result's `first_token` through `PendingEffect::Promote` (promote request to the active/decode set). The token is only emitted as a content token on the **next** decode step, not during prefill. So prefill produces only metadata events (`Start`, `LogprobsDelta`); the first content token waits a full decode step.

Relevant code: `openinfer-qwen3/src/scheduler/resolve.rs` (`resolve_prefill_outputs`, `PendingEffect::Promote`) and `openinfer-qwen3/src/scheduler/effects.rs` (`apply_effects` Promote arm).

## Impact

Single-request TTFT could drop by ~one decode step (~10 ms here, near-halving TTFT in this configuration) if the prefill-produced `first_token` were emitted as a content token during the prefill step.

## Open question before fixing

Verify whether the deferral is a correctness requirement — e.g. stop-token handling or sampling bookkeeping that expects the request to already be in the active set when its first token is emitted. The fix should preserve those invariants.

## Environment

[...]

PR #739, opened 2026-07-22
feat(tracing): fastrace OTLP request tracing with scheduler phase spans

xiaguan:
## What

Adds host-side request tracing to the engine via [fastrace](https://github.com/fast/fastrace) → OpenTelemetry OTLP. When enabled, each request produces a span tree — `request → {queue, prefill, decode}` — exported to any OTLP trace backend (Jaeger, Grafana Tempo, VictoriaTraces). Tracing is **off by default**; setting `OPENINFER_TRACE_OTLP_ENDPOINT` turns it on.

This measures the host-side orchestration timeline that CUDA Graph is built to hide and that nsys/ncu don't attribute per-request: queue wait vs prefill vs decode, per request. It complements — does not replace — GPU profiling.

## Why

Metrics tell you TTFT p99 rose; they can't tell you *why* a specific request was slow. With per-request spans you can filter in TraceQL — e.g. `{ name="prefill" && duration > 100ms }` — and see whether a slow request was stuck in the scheduler queue or in prefill compute. Those have opposite fixes (add capacity vs optimize the kernel).

## Design

- **Contract**: `GenerateRequest` gains `trace_parent: Option<fastrace::collector::SpanContext>` (backward-compatible; `None` = tracing off). `SpanContext` is `Copy`, so it rides through the scheduler's `Clone` request state without holding a live, non-`Clone` `Span`.
- **Frontend** (`openinfer-vllm-frontend`): opens the request root span at submit, passes its context via `trace_parent`, and attributes the pre-scheduler frontend segment (`frontend_ms`) — the one segment only the frontend can see.
- **Scheduler** (`openinfer-qwen3` only): a `PhaseTracker` owns a `HashMap<RequestId, Span>` side-table and opens/closes `queue`/`prefill`/`decode` spans at the **real** state transitions on the scheduler thread — queue at admission, prefill bracketing the forward, decode at first token, closed on every termination path. This is deliberate: spans timed at the scheduler, not inferred from event arrival at the downstream demux, so `prefill` reflects real compute rather than inter-event gap.
- **Reporter** (`openinfer-core::tracing`): installs a `fastrace-opentelemetry` OTLP exporter gated on the env var; no reporter installed when unset, so `Span::root` is a cheap no-op and the scheduler skips span work. `flush()` at process exit.
- **Other model crates**: only receive `trace_parent: None` in `GenerateRequest` literals — they degrade gracefully to a request-level root span with no internal phases. Scheduler instrumentation is scoped to Qwen3 for now.
- **`deploy/tracing/`**: one-command Grafana + Tempo backend (OTLP in, TraceQL out). Backend is swappable — OTLP is the fixed interface.

## Scope notes

- fastrace was already a dependency but effectively dead (one operator-internal `#[trace]` annotation with no reporter wired). That annotation is removed; tracing is now a real, closed loop.
- The version pin `fastrace-opentelemetry = 0.17` tracks the `opentelemetry 0.31` already in the tree (0.18+ jumps to 0.32 and would pull a second, incompatible opentelemetry).

## Testing

- `cargo test --release -p openinfer-engine -p openinfer-core -p openinfer-qwen3 -p openinfer-vllm-frontend --lib` — all green.
- End-to-end: ran Qwen3-4B with an OTLP endpoint, drove concurrent multi-turn load with vllm-bench, verified queue/prefill/decode spans appear in Grafana/Tempo with real per-phase durations and are filterable by `duration` in TraceQL.

## Related

While validating this, per-hop nanosecond measurement surfaced a real TTFT issue (first content token deferred to the first decode step) — filed separately as #738.

### DIFF
--- a/deploy/tracing/docker-compose.yml
+++ b/deploy/tracing/docker-compose.yml
@@ -8,8 +8,9 @@
 #   OPENINFER_TRACE_OTLP_ENDPOINT=http://127.0.0.1:4317 \
 #     cargo run --release -- --model-path /path/to/Qwen3-4B
 #   # open http://localhost:3000 → Explore → Tempo → TraceQL, e.g.
-#   #   { name="ttft" && duration > 500ms }
-#   # to find every request whose time-to-first-token blew past 500ms.
+#   #   { name="prefill" && duration > 500ms }
+#   # to find requests whose prompt phase (first chunk on GPU → first token)
+#   # blew past 500ms. Emitted spans: request → {queue, prefill, decode}.
 #
 # Switching backend later (VictoriaTraces, etc.) is an endpoint/image change,
 # not a code change — everything downstream of OTLP is swappable.
@@ -22,10 +23,11 @@ services:
     command: ["-config.file=/etc/tempo/tempo.yaml"]
     volumes:
       - ./tempo.yaml:/etc/tempo/tempo.yaml:ro
+    # Host ports bind to loopback only: this stack is a local inspection tool,
+    # and Grafana below allows anonymous access — don't expose it on the LAN.
     ports:
-      - "4317:4317"    # OTLP gRPC (openinfer exports here)
-      - "4318:4318"    # OTLP HTTP
-      - "3200:3200"    # Tempo query API (Grafana talks to this internally)
+      - "127.0.0.1:4317:4317"  # OTLP gRPC (openinfer exports here)
+      - "127.0.0.1:3200:3200"  # Tempo query API (Grafana talks to this internally)
 
   grafana:
     image: grafana/grafana:11.3.0
@@ -33,13 +35,14 @@ services:
     restart: unless-stopped
     environment:
       GF_AUTH_ANONYMOUS_ENABLED: "true"
-      GF_AUTH_ANONYMOUS_ORG_ROLE: "Admin"
+      # Viewer, not Admin: Explore/dashboards need no write access.
+      GF_AUTH_ANONYMOUS_ORG_ROLE: "Viewer"
       GF_AUTH_DISABLE_LOGIN_FORM: "true"
       # Enable the TraceQL editor and search UX.
       GF_FEATURE_TOGGLES_ENABLE: "traceqlEditor traceqlSearch"
     volumes:
       - ./grafana-datasource.yaml:/etc/grafana/provisioning/datasources/tempo.yaml:ro
     ports:
-      - "3000:3000"    # Grafana UI
+      - "127.0.0.1:3000:3000"  # Grafana UI
     depends_on:
       - tempo
--- a/deploy/tracing/docker-compose.yml
+++ b/deploy/tracing/docker-compose.yml
@@ -1,0 +1,45 @@
+# Trace backend: Grafana + Tempo.
+#
+# Tempo stores spans and answers TraceQL; Grafana is the UI. openinfer exports
+# over OTLP/gRPC (unchanged Rust side) — point OPENINFER_TRACE_OTLP_ENDPOINT at
+# Tempo's 4317.
+#
+#   docker compose -f deploy/tracing/docker-compose.yml up -d
+#   OPENINFER_TRACE_OTLP_ENDPOINT=http://127.0.0.1:4317 \
+#     cargo run --release -- --model-path /path/to/Qwen3-4B
+#   # open http://localhost:3000 → Explore → Tempo → TraceQL, e.g.
+#   #   { name="ttft" && duration > 500ms }
+#   # to find every request whose time-to-first-token blew past 500ms.
+#
+# Switching backend later (VictoriaTraces, etc.) is an endpoint/image change,
+# not a code change — everything downstream of OTLP is swappable.
+
+services:
+  tempo:
+    image: grafana/tempo:2.6.1
+    container_name: openinfer-tempo
+    restart: unless-stopped
+    command: ["-config.file=/etc/tempo/tempo.yaml"]
+    volumes:
+      - ./tempo.yaml:/etc/tempo/tempo.yaml:ro
+    ports:
+      - "4317:4317"    # OTLP gRPC (openinfer exports here)
+      - "4318:4318"    # OTLP HTTP
+      - "3200:3200"    # Tempo query API (Grafana talks to this internally)
+
+  grafana:
+    image: grafana/grafana:11.3.0
+    container_name: openinfer-grafana
+    restart: unless-stopped
+    environment:
+      GF_AUTH_ANONYMOUS_ENABLED: "true"
+      GF_AUTH_ANONYMOUS_ORG_ROLE: "Admin"
+      GF_AUTH_DISABLE_LOGIN_FORM: "true"
+      # Enable the TraceQL editor and search UX.
+      GF_FEATURE_TOGGLES_ENABLE: "traceqlEditor traceqlSearch"
+    volumes:
+      - ./grafana-datasource.yaml:/etc/grafana/provisioning/datasources/tempo.yaml:ro
+    ports:
+      - "3000:3000"    # Grafana UI
+    depends_on:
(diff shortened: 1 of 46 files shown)


openinfer-qwen3/src/scheduler/phase_trace.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    /// The request's prompt work first reached the GPU: close `queue`, open
+    /// `prefill`. Idempotent across chunked-prefill steps.
+    pub(super) fn enter_prefill(&mut self, id: RequestId) {
+        if let Some(t) = self.tracked.get_mut(&id) {
+            if t.phase != Phase::Queue {
+                return;
+            }
+            t.phase = Phase::Prefill;
+            t.span = Span::root("prefill", t.parent);
```
FeathBow, line 90: Could we clarify whether this span is intended to represent total prompt-phase wall time or actual prefill execution time? After the first chunk, ContinuePrefill requeues the request, while subsequent enter_prefill calls leave the existing span open because the tracked phase is already Prefill. As a result, the duration also includes inter-chunk scheduler iterations, decode work, and scheduling delay for other requests until the final chunk produces the first token. This makes a query such as prefill > 100ms ambiguous between slow forward work and time waiting between chunks. One option would be to rename and document this as prompt-phase wall time; another would be to emit per-step prefill spans around each execution, optionally under an outer prompt-phase span. Either approach would make the TraceQL result easier to interpret correctly.

deploy/tracing/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
...
+      - "4317:4317"    # OTLP gRPC (openinfer exports here)
+      - "4318:4318"    # OTLP HTTP
+      - "3200:3200"    # Tempo query API (Grafana talks to this internally)
+
+  grafana:
+    image: grafana/grafana:11.3.0
+    container_name: openinfer-grafana
+    restart: unless-stopped
+    environment:
+      GF_AUTH_ANONYMOUS_ENABLED: "true"
```
FeathBow, line 37: Since this Compose stack is documented as a local inspection tool, could we make that locality explicit in the port mappings? Compose mappings without a host IP bind on 0.0.0.0, and this configuration also enables anonymous access with the Admin role while disabling the login form. On a shared or remotely reachable GPU host, that can unintentionally expose the Grafana administration and datasource surfaces. Binding the host-facing ports to loopback—for example 127.0.0.1:3000:3000, 127.0.0.1:4317:4317, and 127.0.0.1:3200:3200—would align the configuration with its stated local-only scope. It may also be safer to use anonymous Viewer unless Admin permissions are needed, and the unused OTLP/HTTP port 4318 could remain unpublished.

openinfer-server/src/main.rs
```diff
@@ -133,6 +136,10 @@ async fn main() -> anyhow::Result<()> {
     }
     .context("vLLM frontend server failed")?;
```
FeathBow: Could we flush after both successful and failed server runs? The ? here returns before the call below whenever the HTTP server or local engine bridge fails, so the last buffered spans are not exported on an especially useful diagnostic path. Preserving the result, calling openinfer_core::tracing::flush(), and then propagating the error—for example, let result = ...await.context(...); flush(); result?;—would retain the current success behavior while covering Err as well.

deploy/tracing/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
+# Trace backend: Grafana + Tempo.
+#
+# Tempo stores spans and answers TraceQL; Grafana is the UI. openinfer exports
+# over OTLP/gRPC (unchanged Rust side) — point OPENINFER_TRACE_OTLP_ENDPOINT at
+# Tempo's 4317.
+#
+#   docker compose -f deploy/tracing/docker-compose.yml up -d
+#   OPENINFER_TRACE_OTLP_ENDPOINT=http://127.0.0.1:4317 \
+#     cargo run --release -- --model-path /path/to/Qwen3-4B
+#   # open http://localhost:3000 → Explore → Tempo → TraceQL, e.g.
```
FeathBow, line 10: Could we update this example to use one of the span names emitted by the implementation? The current tree contains request, queue, prefill, and decode, but no ttft span, so this query will not return the requests described in the comment. Once the prefill semantics are settled, an example using the current scoped intrinsics could be { span:name = "prefill" && span:duration > 500ms }, or the equivalent query for another emitted phase.

openinfer-qwen3/src/scheduler/phase_trace.rs
```diff
@@ -0,0 +1,95 @@
...
+    span: Span,
+}
+
+/// Side-table of in-flight request phase spans. Owned by the scheduler loop.
+#[derive(Default)]
+pub(super) struct PhaseTracker {
+    tracked: HashMap<RequestId, Tracked>,
+}
+
+impl PhaseTracker {
```
FeathBow, line 62: Could we add a small behavioral test that inspects the emitted span records? The current test changes keep existing call sites compiling with trace_parent: None, but they do not verify parentage, span names, transition order, repeated transitions, or cleanup. A TestReporter-backed state-machine test could cover queue → prefill → decode → finish, rejection, and repeated chunk transitions without requiring a GPU; a fake-scheduler case could then pin down the intended chunked-prefill timing. This would give the new tracing contract direct regression coverage rather than relying only on the manual happy-path E2E.

FeathBow (commented):
Thanks for pinging me on this, I love this idea! I’ve left a few detailedcomments below :)

openinfer-qwen3/src/scheduler/phase_trace.rs
xiaguan: Good catch — went with the rename-and-document option: `prefill` is now explicitly documented as prompt-phase wall time (first chunk on GPU → first token), which deliberately includes the inter-chunk scheduler iterations under chunked prefill (5256c506). Rationale: `queue + prefill` reads as TTFT, and what a slow request waited on between its own chunks is visible in the trace waterfall as other requests' decode/prefill spans overlapping the gap. Per-step `prefill_step` child spans are the fallback if that cross-request view ever proves insufficient — deferred until there's a concrete need. The contract test added in the same commit pins the single-span-across-chunks behavior.

deploy/tracing/docker-compose.yml
xiaguan: Done in 71c7753d: host ports now bind to loopback (`127.0.0.1:3000:3000`, `127.0.0.1:4317:4317`, `127.0.0.1:3200:3200`), anonymous access is downgraded to Viewer, and the unused OTLP/HTTP 4318 mapping is dropped.

openinfer-server/src/main.rs
xiaguan: Done in 71c7753d: the serve result is preserved, `openinfer_core::tracing::flush()` runs, then the error propagates — the failed-server path keeps its last buffered spans.

deploy/tracing/docker-compose.yml
xiaguan: Done in 71c7753d: the example now queries `{ name="prefill" && duration > 500ms }` and the comment lists the emitted spans (`request → {queue, prefill, decode}`). With the wall-time semantics settled (see the prefill-span thread), `queue + prefill` is the TTFT decomposition.

openinfer-qwen3/src/scheduler/phase_trace.rs
xiaguan: Added a TestReporter-backed contract test in 5256c506 covering queue → prefill → decode → finish, parentage under the request root, repeated chunked-prefill transitions (idempotent), rejection cleanup, and the tracing-off no-op — no GPU or scheduler harness needed. Deferring the fake-scheduler timing case: with the wall-time semantics now documented, the idempotency assertion pins the intended chunked-prefill behavior at a fraction of the harness cost.

PR #751, opened 2026-07-24
feat(glm52): define TP4 prefill-only contract

xiaguan:
## Summary

- add `--glm52-prefill-only` for the TP4 topology and require prefix caching, one active request, and `max_tokens=1`
- enforce `max_tokens=1` at the OpenAI HTTP boundary (`400 invalid_request_error`) and repeat it in scheduler intake for non-HTTP callers
- add a page-aligned `--glm52-prefill-chunk-size` contract with a 16,384-token default
- reserve the future large-M prefill scratch envelope before VRAM-derived `max_model_len` selection
- keep every compatibility prefill step eager: prefill-only startup neither captures nor replays whole-step decode graphs
- add explicit TP slice-bank timing so the tensor-replicated weight-load second pass is visible

## Scope boundary

This is the contract/capacity PR, not the native large-M executor. Prompt rows temporarily ride the existing small-row whole-step implementation, executed eagerly. The prompt-tail output is returned as the first predicted token and the request finishes immediately; that token is never fed back into the model.

The native TP4 prefill executor remains eager and lands separately. This PR makes no prefill-throughput claim.

## Request contract

Prefill-only mode:

- requires `--moe-topo=tp4 --tp-size=4`
- keeps the existing content-hashed 64-token prefix cache enabled
- admits one active request per TP group
- requires `max_tokens=1`
- rejects DSpark, KV offload/external P/D, `--no-prefix-cache`, and decode graph export

## Capacity policy

The chunk size is a future compute tile, not a prompt-length limit. It defaults to 16K and must be page aligned. Auto `max_model_len` continues to use the fleet-minimum post-weight VRAM probe and exact arena ledger, with an additional reservation:

```text
256 MiB fixed + 72 KiB * prefill_chunk_size
```

The default reservation is 1,476,395,008 bytes per rank. Explicit `--max-model-len` values are checked against the same budget, and the chunk must fit inside the selected cap.

## Validation

Rebased onto `main` after the pinned-staging weight loader merged.

- `cargo check --release -p openinfer-server --features glm52`
- GLM5.2 library suite: 76 passed, 16 ignored
- HTTP contract test: invalid `max_tokens` returns 400 before engine submission
- `cargo fmt --all --check`
- `git diff --check`
- real 4×GB300 TP4 launch with the GLM-5.2-FP8 checkpoint
  - auto `max_model_len=204352`, 16K chunk reservation 1.4 GiB/rank
  - startup contained zero whole-step graph capture markers
  - `max_tokens=1`: HTTP 200, one completion token, `finish_reason=length`
  - `max_tokens=2`: HTTP 400 in 0.43 ms; no scheduler rejection marker

## Weight-load observation

The earlier apparent gap between a ~12 s base loader profile and ~208 s rank-load wall time came from an unprofiled tensor-replicated second pass. Every TP rank independently re-reads and host-gathers all 75 MoE layers (257 experts × 6 tensors) into its intermediate slice banks.

On the warm-cache validation launch:

- base resident plan: 1.64–2.13 s/rank
- TP slice-bank second pass: 44.09–44.62 s/rank (0.59 s/layer)
- whole rank weight phase: 46.80 s
- engine load: 53.00 s

The first colder run's 208 s phase and the warm run's 46.8 s phase are therefore the same structural second pass under different page-cache residency, not a pinned-staging regression. Consolidating that pass is follow-up work.


xiaguan:
@codex review

xiaguan:
@codex review

PR #759, opened 2026-07-24
perf(glm52): accelerate native TP4 prefill

xiaguan:
## Summary

Accelerate the native GLM-5.2 FP8 TP4 prefill-only path and keep the execution contract prefill-only.

- increases the outer prefill chunk tile to 16K while keeping attention/indexer work in bounded subtiles;
- replaces per-expert routed work with FlashInfer's TRT-LLM fused-MoE runner and checked-in SM100 cubins;
- gathers compact paged index keys and runs the upstream DeepGEMM SM100 unpaged MQA kernel;
- accumulates attention tiles before the output projection/reduction;
- uses NCCL BF16 all-reduce for prefill and removes the unused custom P2P fallback;
- preserves paged prefix-cache admission/reuse and documents automatic model-length capacity;
- records TP4 versus TP4+EP4 backend and concurrency sweeps.

The TRT-LLM interfaces and cubins come from `flashinfer-cubin 0.6.13`, matching the vendored FlashInfer revision. Only the four selected SM100 kernels and their four metadata records are checked in; the full 2,108-kernel artifact is not vendored. Provenance and hashes are recorded under `openinfer-kernels/cubin/glm52/trtllm_moe/`.

## Performance

Single 4×GB300 host, TP4, GLM-5.2-FP8, FP8 E4M3 KV, concurrency one, one greedy output token, 16K chunk budget, warmed server:

| Input tokens | OpenInfer median TTFT | vLLM median TTFT | OpenInfer / vLLM |
| ---: | ---: | ---: | ---: |
| 1K | 170 ms | 306 ms | 0.56× |
| 4K | 459 ms | 304 ms | 1.51× |
| 16K | 1.745 s | 0.898 s | 1.94× |
| 64K | 7.186 s | 3.773 s | 1.90× |
| 128K | 10.312 s | 8.071 s | 1.28× |

The detailed methodology, vLLM TP4/EP4 all-to-all sweep, concurrency sweep, and evidence boundaries are in `docs/models/glm52/tp4-prefill-only.md`.

## Validation

- `cargo check -p openinfer-server --features glm52`
- `cargo test -p openinfer-glm52 --lib` — 78 passed, 17 ignored
- `git diff --check`
- live 4-GPU startup at `--max-model-len 1000000`
- real HTTP one-token prefill-only response
- five fixed greedy prompts matched vLLM TP4's first token
- concurrent requests and warm prefix-cache reuse


xiaguan:
@codex review

xiaguan:
Trimmed the generated TRT-LLM metadata from 2,108 kernels to the four cubins used by this path; the PR is now ~10.5k additions and the SM103 server build was rerun from the native source. @codex review

PR #761, opened 2026-07-25
fix(qwen35): size the TP sampling scratch from the decodable vocab

### LINKED ISSUE #760: qwen35: TP serving fails because the sampling scratch is sized from the padded vocab
## Description

Every tensor-parallel prefill and decode fails on Qwen3.5-4B with:

```
select_batch: logits vocab 248077 != scratch vocab 248320
```

248077 is the tokenizer's decodable vocabulary; 248320 is the checkpoint's padded `vocab_size`.

#676 restricted token selection to the decodable vocabulary: `Config35::selection_vocab` is narrowed at load time when the tokenizer stops short of the checkpoint width, and every logits arena is sized from it — prefill, the decode buffers, and the unified forward path, along with the `SampleScratch` those paths build. The tensor-parallel executor was not part of that change; it still builds its `SampleScratch` from `vocab_size`. That line predates #676 (it arrived with the TP phase 1 work in #627), and `tp_executor.rs` does not appear in #676's changed files.

`select_batch` requires the scratch width to equal the logits width. The check is deliberate — a mismatch would otherwise be a silent kernel out-of-bounds rather than a clean error — so it fails closed and takes every TP request with it.

Scope of the breakage: any checkpoint whose tokenizer defines fewer ids than the checkpoint pads to, on the tensor-parallel path only. Single-GPU serving is unaffected because it sizes both sides from `selection_vocab`. Qwen3 and Kimi have no equivalent narrowing and are unaffected.

Why CI is green: the four tests covering this path — the two `hf_golden_gate` TP2 logits gates, `e2e_scheduler` TP2, and `serving_tp2` — are all `#[ignore]`, since they need two CUDA devices, NCCL, and weights.

The fix is to size the TP scratch from `selection_vocab` like every other call site. Verification is a red/green differential on two GPUs: the current tree reproduces the error, the fixed tree passes all four tests.

FeathBow:
## Description

Fixes #760 

#676 bounded token selection to the tokenizer's decodable vocabulary and sized every logits arena from `selection_vocab`, but it did not touch `tp_executor.rs`, where the tensor-parallel worker still built its `SampleScratch` from the checkpoint's padded `vocab_size` — a line that predates it, from #627. `select_batch` requires the scratch width to equal the logits width, because a mismatch would otherwise be a silent kernel out-of-bounds rather than a clean error, so on any checkpoint whose tokenizer stops short of the padded width every tensor-parallel prefill and decode failed with `select_batch: logits vocab 248077 != scratch vocab 248320`. Qwen3.5-4B is such a checkpoint, so TP serving was broken for it.

The four tests that cover this path — the two `hf_golden_gate` TP2 logits gates, `e2e_scheduler` TP2, and `serving_tp2` — are all `#[ignore]` because they need two CUDA devices, NCCL, and weights, which is why CI stayed green.

After the change the only remaining uses of `vocab_size` in this crate are the narrowing comparison itself and the checkpoint tensor-shape validation, both of which want the padded width. Qwen3 and Kimi have no equivalent narrowing and are unaffected.

## Type of Change

### DIFF
--- a/openinfer-qwen35/src/tp_executor.rs
+++ b/openinfer-qwen35/src/tp_executor.rs
@@ -822,7 +822,7 @@ impl TpWorkerPrepared {
         let decode_buffers = model.create_batch_decode_buffers_with_capacity(max_batch)?;
         let sample_scratch = openinfer_sample::SampleScratch::new(
             model.device_ctx(),
-            model.config().vocab_size,
+            model.config().selection_vocab,
             max_batch,
         )?;
         Ok((


FeathBow:
Verified on two GH200s. The arms are the same three test binaries built from clean `887e774` and from `887e774` plus this commit, run alternately per test so a mid-run change in machine state cannot look like the fix.

| test | main | this commit |
|---|---|---|
| `hf_golden_gate` TP2 | 0 passed, 2 failed | 2 passed |
| `e2e_scheduler` TP2 | 0 passed, 1 failed | 1 passed |
| `serving_tp2` | 0 passed, 1 failed | 1 passed |

Every failure on the main arm carries `select_batch: logits vocab 248077 != scratch vocab 248320`, and the string is absent from all three passing runs.

Issue #758, opened 2026-07-24
[Model] Gemma 4 roadmap


FeathBow:
Crate: `openinfer-gemma4` · Cargo feature: `gemma4`

Ship text-only first. Bring up the shared dense text stack on 12B (H200 correctness first, then a measured RTX 5090 envelope), use it to land heterogeneous local/global attention and hybrid KV management, then add the 26B-A4B MoE on one H200. Validate 31B BF16 on one H200 after the common stack is green; it must not block the strategic MoE line. Do not commit to "26B-A4B QAT INT4 on one 5090" or "31B BF16 TP2 on consumer GPUs": Google does not publish a compressed-tensors W4A16 checkpoint for the 26B MoE because of quality loss, and the 31B BF16 memory envelope is not credible on 2×32 GB without a measured, context-bounded budget. Treat quantization and Gemma's MTP drafter as separate design tracks.

The largest new infrastructure is not just SWA plumbing. It is heterogeneous attention and KV storage: local layers use head_dim 256, 8 KV heads, full RoPE and a 1024-token window; global layers use head_dim 512, size-dependent global KV heads, proportional partial RoPE (128/512 dimensions), and K/V projection sharing. Current OpenInfer FlashInfer wrappers instantiate only HD128/HD256 and compile `sliding_window=false`, so HD512 backend selection and a real sliding-window variant are M0 go/no-go spikes.

## Positioning

The near-term product boundary remains small-to-mid models, single node, coding/agent backend, and fast startup.

- **Product anchor:** 12B is the consumer-GPU anchor and first correctness carrier. Its 5090 claim is conditional on a retained memory/serving envelope, not on weight bytes alone.
- **Strategic payload:** 26B-A4B is the first Gemma MoE and the main reason to harden a single-GPU, no-EP BF16 MoE path.
- **Dense flagship:** 31B BF16 is an H200 target. It is valuable but must not delay the 26B MoE after the shared text stack is green.
- **Shared boundary:** Gemma may provide evidence for shared token routing/grouping primitives, but extraction into `openinfer_kernels::moe` happens only after the Gemma implementation can be compared with the existing GLM5.2 and Kimi consumers. Router math, dtype and weight layouts remain model-local until proven otherwise.

All Gemma 4 lines start at Bring-up. Promotion to Maturing requires passing the objective gates below; "roadmap target" does not itself imply Maturing support.

## Targets

[...]

PR #775, opened 2026-07-27
fix(server): fail closed on unrecognized model configs

### LINKED ISSUE #774: server: model detection silently routes unknown configs to the qwen engines
## Description

`detect_model_type` reads `config.json` and ends in two unconditional fallbacks: any config carrying a `text_config` object is routed to the Qwen3.5 engine, and everything else is routed to Qwen3. Only the DeepSeek-V2-Lite, GLM5.2 and Kimi-K2 branches validate what they matched; the two Qwen tails validate nothing.

The failure mode is silent misrouting. An unsupported model directory starts the wrong engine and fails later during model-specific config or weight loading with an unrelated field, shape, or tensor error, not one about the model being unsupported. e.g. every Gemma 4 checkpoint ships a `text_config`, so pointing today's server at one silently starts the Qwen3.5 engine instead of saying "this model is not supported".

A smaller bug in the same tail: the Qwen3 fallback's feature-gate message suggests `--features qwen3-4b`, a feature name that no longer exists (the feature is `qwen3`).

## Proposed fix

- give `openinfer-qwen3` and `openinfer-qwen35` the same `probe_config_json` treatment as glm52/kimi/deepseek — identity assertions only, no size pinning, since both lines serve multiple sizes with one crate
- route Qwen3 by `model_type "qwen3"` and Qwen3.5 by `model_type "qwen3_5"` (or `text_config.model_type "qwen3_5_text"`); the identity strings are uniform across every published dense size these crates serve — the MoE variants declare their own `model_type` and are out of scope — (Qwen3 dense 0.6B–32B: `Qwen3ForCausalLM`, no `text_config`; Qwen3.5 0.8B–27B: `Qwen3_5ForConditionalGeneration` + `text_config.model_type "qwen3_5_text"`)
- when nothing matches, error instead of defaulting: name the seen `model_type` and `architectures`, and list the model families compiled into this build

A config with `model_type` missing entirely should also land in that error — today it silently routes to Qwen3.

FeathBow:
## Description

Fixes #774

`detect_model_type` ended in two unconditional tails: any config carrying a `text_config` object was routed to the Qwen3.5 engine and everything else to Qwen3, with no validation. An unsupported model directory started the wrong engine and failed later during model-specific config or weight loading with an unrelated field, shape, or tensor error instead of an "unsupported model" one. Every Gemma 4 checkpoint ships a `text_config`, so the #758 bring-up would have silently started the Qwen3.5 engine.

- `openinfer-qwen3` and `openinfer-qwen35` gain `probe_config_json`, mirroring the existing glm52/kimi/deepseek probes. Identity assertions only, no size pinning — both crates serve multiple sizes. The identity strings were checked against every published dense size these crates serve (the MoE variants such as Qwen3-30B-A3B declare their own `model_type` and are not supported by these engines): Qwen3 0.6B/1.7B/4B/8B/14B/32B all carry `model_type "qwen3"` + `Qwen3ForCausalLM` and no `text_config`; Qwen3.5 0.8B/2B/4B/9B/27B all carry outer `model_type "qwen3_5"` + `Qwen3_5ForConditionalGeneration` + `text_config.model_type "qwen3_5_text"`.
- The server routes Qwen3.5 by outer `model_type` or `text_config.model_type`, Qwen3 by `model_type`, both through their probes; the deepseek/glm52/kimi branches are untouched.
- The fallthrough is now an error naming the seen `model_type` and `architectures` verbatim — the raw JSON value when the key is present (wrong-typed values included), `missing` only when it is absent; a config without `model_type` used to silently route to Qwen3 — plus the model families compiled into this build.
- The Qwen3 feature-gate message suggested `--features qwen3-4b`, a feature name that no longer exists; it now says `--features qwen3`.

Behavior change: a config that previously fell through to an engine by default now errors. Any dense Qwen3 / Qwen3.5 checkpoint supported by these engines keeps routing exactly as before; only configs that no engine actually supports stop being accepted.

## Test Env

Eight unit tests in `server_engine.rs` — routing for both Qwen identities, Gemma 4 12B and 26B identity configs rejected (these flip into routing tests when a gemma4 family lands), an unknown family rejected, a config with no `model_type` rejected, wrong-typed identity fields rendered verbatim rather than as missing, and a probe failure on wrong `architectures`.

Verification on Single GPU (sm_89, x86_64), on a base that includes #771: `cargo fmt --check`; `cargo clippy --all-targets -- -D warnings` for openinfer-server (default features and `--features qwen35`), openinfer-qwen3, and openinfer-qwen35; `cargo test --release -p openinfer-server --lib` under default features (7 tests) and `--features qwen35` (8 tests, so the qwen35 routing test really ran); `cargo check -p openinfer-server --no-default-features` for the no-model-features build.

## Type of Change

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3385,6 +3385,7 @@ dependencies = [
  "rand 0.10.1",
  "serde",
  "serde_json",
+ "tempfile",
  "tikv-jemallocator",
  "tokio",
  "tokio-util",
--- a/openinfer-qwen3/src/config.rs
+++ b/openinfer-qwen3/src/config.rs
@@ -2,7 +2,10 @@ use std::fs;
 
 use anyhow::Context;
 use anyhow::Result;
+use anyhow::bail;
+use anyhow::ensure;
 use serde::Deserialize;
+use serde_json::Value;
 
 pub(crate) const PREFILL_ATTENTION_CTA_TILE_Q: i32 = 64;
 
@@ -503,4 +506,22 @@ impl TensorParallelConfig {
     pub(crate) fn is_sharded(self) -> bool {
         self.world_size > 1
     }
+}
+
+/// Identity check that `json` is a Qwen3 config; size and shape validation belong to the config loader.
+pub fn probe_config_json(json: &Value) -> Result<()> {
+    let model_type = json.get("model_type").and_then(Value::as_str).unwrap_or("");
+    if model_type != "qwen3" {
+        bail!("not a Qwen3 config: model_type={model_type}");
+    }
+    let architectures: Vec<&str> = json
+        .get("architectures")
+        .and_then(Value::as_array)
+        .map(|arr| arr.iter().filter_map(Value::as_str).collect())
+        .unwrap_or_default();
+    ensure!(
+        architectures.contains(&"Qwen3ForCausalLM"),
+        "Qwen3 architectures must contain Qwen3ForCausalLM"
+    );
+    Ok(())
 }
--- a/openinfer-qwen3/src/lib.rs
+++ b/openinfer-qwen3/src/lib.rs
@@ -24,6 +24,7 @@ use std::path::Path;
 use std::path::PathBuf;
 
 use anyhow::Result;
+pub use config::probe_config_json;
 use log::info;
 use log::warn;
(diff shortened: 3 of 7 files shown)


FeathBow:
@codex review please.

PR #784, opened 2026-07-29
feat(kernels): instantiate FlashInfer attention at head_dim 512

### LINKED ISSUE #783: gemma4: instantiate flashinfer attention at head_dim 512
## Description
Gemma 4 global attention layers run head_dim 512 (bf16, one to four KV heads depending on size). The vendored FlashInfer v0.6.14 supports hd512 end to end — the 99KB-smem K/V time-sharing layout for SM86/89/120/121, the VO-split paged path, and the `FA2DetermineCtaTileQ` hd512 branch — but nothing in this repo instantiates it: the compiled head dims are 64/128 (qwen3 and drafters), 192/128 (kimi MLA) and 256 (qwen35). Without an hd512 instantiation no Gemma 4 global layer can execute a single attention op.

## Proposed Plan

- three instantiations in a standalone TU: single prefill, paged batch prefill, paged batch decode
- FFI declarations and safe wrappers
- correctness gates against a naive f64 host reference on a 99KB-smem GPU (sm_89), covering the GQA shapes the three published sizes actually use — 16 query heads over 1 KV head (12B), and the 2/4-KV-head configurations of the larger sizes — plus paged layouts with page-boundary crossings

Two constraints known up front that the implementation has to encode:

- the decode dispatcher compiles GQA groups {1,2,3,4,8} only, so group 16 (12B global layers) has to decode through the paged-prefill path; upstream flashinfer#3563 tracks widening the decode dispatch
- the paged-prefill cta_tile_q override helper accepts 16|64|128 while hd512 only ever selects 16|32, so the hd512 paged-prefill entry point cannot take an override and always uses the kernel's own derivation

Throughput is out of scope here. Correctness on the 99KB-smem SKU settles the capacity question — the configuration fits the smallest shared-memory budget the kernels target — but it does not certify other architectures: sm_90/sm_120 need their own correctness and performance runs when those platforms enter the picture.

FeathBow:
## Description

Fixes #783 

Gemma 4 global attention layers run head_dim 512, a shape no supported model line has needed: the vendored FlashInfer v0.6.14 carries hd512 support end to end (the 99KB-smem K/V time-sharing layout, the VO-split paged path, the `FA2DetermineCtaTileQ` hd512 branch), but this repo instantiated nothing at 512. This adds the three instantiations and the plumbing to reach them.

- New TU `csrc/shared/paged_attention_hd512.cu` with `single_prefill_cuda_hd512`, `batch_prefill_paged_cuda_hd512`, `paged_attention_decode_cuda_hd512`, mirroring the hd256 trio. A separate TU so the long FlashInfer template build parallelizes with `paged_attention.cu`; build.rs adds it to the FlashInfer-include set and the nvcc scheduling priorities.
- FFI declarations and safe wrappers (`batch_prefill_paged_hd512_into`, `paged_attention_batch_decode_hd512_into`, `paged_attention_batch_decode_via_prefill_hd512_into`, `single_prefill_hd512_into`). The wrappers take batch size and total tokens from the plan rather than redundant parameters. `PrefillPagedPlan` now records the geometry it was built with (head dim, query/KV head counts) and the hd512 wrappers validate it, so a plan built for another head configuration is rejected instead of misdispatching (a preallocated plan starts with geometry unset and records it on first update, so no constructor signatures outside openinfer-kernels change). Direct-decode metadata device arrays travel in an `Hd512DecodeMetadata` aggregate (private fields, plain constructor); the wrapper derives the batch size from `q.seq_len` — the single shape owner — and unconditionally validates every slice length against it before touching pointers; what cannot be checked host-side — that page indices stay within the KV buffer — is documented as caller contract.
- The paged-prefill wrapper takes no cta_tile_q override: hd512 only ever selects 16 or 32 via `FA2DetermineCtaTileQ`, and the shared override helper accepts 16|64|128.
- The decode dispatcher compiles GQA groups {1,2,3,4,8} only, so group 16 (Gemma 4 12B global layers: 16 query heads over 1 KV head) decodes through the paged-prefill path; both decode routes are wrapped and tested. Upstream flashinfer#3563 tracks widening the decode dispatch.
- ~`tests/hd512_attention_smoke/` (main/reference/fixture modules): host-reference tests comparing each kernel entry point against naive f64 attention at the global-attention shapes the three published sizes actually use (16/1, 16/2, 32/4), plus tail-aligned short-query prefill and both decode routes. The paged tests compute the reference over the post-run device pages, so fill/scatter conventions cannot desynchronize reference and kernel; the single-prefill test compares against the same bf16-quantized host K/V it uploads~.

## Test Env

Verification on Single GPU (sm_89, x86_64): `cargo build --release -p openinfer-kernels`, `cargo fmt --check`, `cargo clippy --release -p openinfer-kernels --all-targets -- -D warnings`, ~and the smoke suite 7/7 passing with mean-normalized deltas 1.2e-3 to 1.8e-3~. The suite requires the worktree's own flashinfer submodule checkout (v0.6.14); it skips cleanly on machines without a CUDA device.

Throughput numbers are deliberately out of scope. These results establish correctness on sm_89 and settle the 99KB shared-memory capacity question; they do not certify other architectures — sm_90/sm_120 need their own correctness and performance runs.

## Type of Change

### DIFF
--- a/openinfer-kernels/csrc/shared/paged_attention_hd512.cu
+++ b/openinfer-kernels/csrc/shared/paged_attention_hd512.cu
@@ -1,0 +1,291 @@
+// FlashInfer attention instantiations at HEAD_DIM=512 (Gemma 4 global
+// attention layers). Separate TU so the long FlashInfer template builds
+// compile in parallel with paged_attention.cu.
+
+#include "ffi_guard.cuh"
+
+#include <cuda_runtime.h>
+#include <cuda_bf16.h>
+#include <cstdint>
+
+#include <flashinfer/attention/decode.cuh>
+#include <flashinfer/attention/prefill.cuh>
+#include <flashinfer/attention/default_decode_params.cuh>
+#include <flashinfer/attention/default_prefill_params.cuh>
+#include <flashinfer/attention/variants.cuh>
+#include <flashinfer/page.cuh>
+
+using namespace flashinfer;
+
+using DType  = __nv_bfloat16;
+using IdType = int32_t;
+using ParamsT = BatchDecodeParams<DType, DType, DType, IdType>;
+using Variant = DefaultAttention</*custom_mask=*/false,
+                                 /*sliding_window=*/false,
+                                 /*logits_soft_cap=*/false,
+                                 /*alibi=*/false>;
+
+using BatchPrefillParamsT = BatchPrefillPagedParams<DType, DType, DType, IdType>;
+using PrefillParamsT = SinglePrefillParams<DType, DType, DType>;
+
+// Helper: build paged_kv_t from our page-first layout.
+//
+// Our KvPool stores all layers interleaved in one buffer. For a given layer L:
+//   k_data = base + L * layer_stride
+//   v_data = base + L * layer_stride + kv_block_len
+//   stride_page = page_stride  (spans all layers — jumps to same layer in next page)
+//   NHD within-block: stride_n = num_kv_heads * head_dim, stride_h = head_dim
+static paged_kv_t<DType, IdType> make_paged_kv(
+    void*    kv_data,
+    int64_t  k_offset_elems,
+    int64_t  v_offset_elems,
+    int32_t* page_indices,
+    int32_t* page_indptr,
+    int32_t* last_page_len_d,
+    int32_t  num_kv_heads,
+    int32_t  head_dim,
+    int32_t  page_size,
+    int32_t  batch_size,
+    int64_t  stride_page)
+{
+    DType* k_data = reinterpret_cast<DType*>(kv_data) + k_offset_elems;
+    DType* v_data = reinterpret_cast<DType*>(kv_data) + v_offset_elems;
+
+    // kv_strides[0] = stride_page, [1] = stride for NHD-n, [2] = stride for NHD-h
+    int64_t kv_strides[3] = {
+        stride_page,
+        static_cast<int64_t>(num_kv_heads) * head_dim,
+        static_cast<int64_t>(head_dim),
+    };
+
+    return paged_kv_t<DType, IdType>(
+        num_kv_heads, page_size, head_dim, batch_size,
+        QKVLayout::kNHD,
+        k_data, v_data, kv_strides,
+        page_indices, page_indptr, last_page_len_d,
+        /*rope_pos_offset=*/nullptr);
+}
+
+extern "C" {
+
+int single_prefill_cuda_hd512(
+    void*    q,
+    void*    output,
+    void*    k_cache,
+    void*    v_cache,
+    int32_t  num_qo_heads,
+    int32_t  num_kv_heads,
+    int32_t  seq_len,          // number of Q tokens (qo_len)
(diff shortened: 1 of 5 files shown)


openinfer-kernels/tests/hd512_attention_smoke/main.rs
```diff
@@ -0,0 +1,403 @@
+mod fixture;
```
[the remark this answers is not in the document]
FeathBow: tests/hd512_attention_smoke/main.rs is an auto-discovered integration-test target.

openinfer-kernels/csrc/shared/paged_attention_hd512.cu
```diff
@@ -0,0 +1,291 @@
...
+    params.block_valid_mask  = nullptr;
+    params.kv_chunk_size_ptr = kv_chunk_size_ptr;
+    params.max_total_num_rows = seq_len;
+    params.total_num_rows    = total_num_rows;
+    params.padded_batch_size = padded_batch_size;
+    params.partition_kv      = false;
+
+    uint32_t group_size = num_qo_heads / num_kv_heads;
+    int64_t packed_qo_len = static_cast<int64_t>(seq_len) * group_size;
+    uint32_t cta_tile_q = FA2DetermineCtaTileQ(packed_qo_len, head_dim);
```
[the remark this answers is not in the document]
FeathBow: Makes sense.

FeathBow:
@codex review please.

openinfer-kernels/tests/hd512_attention_smoke/fixture.rs
```diff
@@ -0,0 +1,406 @@
+use cudarc::driver::CudaSlice;
```
xiaguan: remove this smoke test? We will introduce something new to test this kernel ?

FeathBow: I'd like to keep it until the new tests land. cas it's the only gate (but already passed). Happy to drop it once the replacement covers this.

xiaguan: or make it ignore? Run a test that will never fail is useless?

xiaguan: We shouldn't duplicate FlashInfer's kernel test matrix. We should trust FlashInfer's kernel correctness and test only the contracts owned by OpenInfer: instantiation, FFI wiring, metadata/layout handling, and dispatch behavior.

For each test, we should ask: what OpenInfer contract does it cover, when would it fail, and what confidence would we lose by removing it? If it only revalidates FlashInfer numerics, we should remove it.

FeathBow: makes sense

FeathBow: Added this fixture while debugging a numerical accuracy issue during bring-up and forgot to remove it. Should not remain as a permanent test here. Thanks for catching this.

PR #786, opened 2026-07-29
feat(qwen35): overlap Unified prefill with active decode

### LINKED ISSUE #715: qwen35: reduce active-decode stalls from serial Unified prefill
## Summary

PR #690 closed #689 and removed most of the Qwen3.5-4B HTTP c16 gap by batching the linear-attention decode kernels. This is a post-#690 follow-up for the smaller residual gap. It does not replace #689's attribution of the original gap.

At current `main`, `scheduler/plan.rs` selects `ExecutionPlan::Unified` whenever a scheduled prefill chunk and active decode coexist. `unified_forward.rs::unified_step` then runs the chunk's prefills serially before launching `batch_decode_graph` for the active requests. Those Unified ticks can delay the corresponding active decode step by the scheduled prefill work.

This issue tracks the scheduler/executor admission and Unified-step policy. It is separate from #691, which targets the GDN prefill kernel itself, and from #470, which owns the mixed-load ITL and adaptive chunk-budget work.

## Retained evidence

Environment and workload:

- GPU: 1x NVIDIA GeForce RTX 5090 32GB
- OpenInfer: `9fdd4201243ed4999d09e20730c7fd5186e98d4f` (#690 merge commit)
- Model: `Qwen/Qwen3.5-4B`
- Baseline: vLLM 0.25.1
- Client: `vllm bench serve` against the OpenAI completions endpoint
- Primary cell: random input 1024, fixed output 256, concurrency 16, greedy, ignore EOS
- Prefix cache: disabled for the primary comparison
- Repetitions: 3 per formal cell

Post-#690 HTTP results:

The fixed-concurrency cells use 256 output tokens. The QPS sweep follows the #469 1024/128 contract.

| Cell | completed OI/vLLM | failed OI/vLLM | avg out OI/vLLM | TPOT ms OI/vLLM | output tok/s OI/vLLM |
| --- | ---: | ---: | ---: | ---: | ---: |
| 1024/256 c1 | 12/12 | 0/0 | 256/256 | 7.01/6.22 | 138.9/119.9 |
| 1024/256 c16 | 48/48 | 0/0 | 256/256 | 12.20/9.66 | 1204.8/1424.2 |
| 1024/128 QPS 8 | 48/48 | 0/0 | 128/128 | 14.31/12.25 | 839.3/857.9 |
| 1024/128 QPS 12 | 72/72 | 0/0 | 128/128 | 23.40/15.73 | 1168.2/1284.6 |
| 1024/128 QPS 16 | 96/96 | 0/0 | 128/128 | 30.11/21.87 | 1333.7/1559.6 |

All formal cells completed with zero failed requests and no zero-token outputs. Short SHA-256 hashes over each run's generated outputs were retained:

<details>
<summary>Output sanity hashes</summary>

| Cell | OpenInfer run hashes | vLLM run hashes |
| --- | --- | --- |
| 1024/256 c1 | `c1859f4869e6cdce` x3 | `d0278b9bd97940b1` x3 |
| 1024/256 c16 | `c0754a6e906c9bf7`, `9fc630b6067c7610`, `1f390163ad87f744` | `6706a20e21c99034`, `3cbd498e519437f8`, `36185749eb210883` |
| 1024/128 QPS 8 | `983e2d3497716ff7`, `a7f90b65f74833a3` x2 | `eaf996c598d3f4e2`, `0be07fb9ac90dfd7`, `73c7d838f6b58c50` |
| 1024/128 QPS 12 | `fb017d797cf9a0e2`, `35496426a32d0f78`, `e1dc123486287432` | `20b6c5f2f32257c3`, `422b6b72b4727bc2`, `a8727b09f2f89ced` |
| 1024/128 QPS 16 | `cab81238502741ba`, `a24de22559e6ea15`, `54acbbd134a1fb45` | `a1759ecb94ed982e`, `1c76ce228f31cc82`, `978428266ba28e21` |

</details>

The c16 trace replay separately completed 48/48 requests, averaged 256 output tokens, and had no zero-token output.

One valid c16 serving trace shows:

| Trace field | Result |
| --- | ---: |
| Scheduler ticks | 737 Decode / 47 Unified / 1 Prefill |
| Tick ratio | 93.9% Decode / 6.0% Unified / 0.1% Prefill |
| Scheduler wall share | 70.8% Decode / 28.6% Unified / 0.6% Prefill |
| Unified wall time | 62.73 ms average |
| Decode batch size | p50 16 |
| Decode buckets | `{1:2, 2:2, 4:4, 8:8, 16:768}` |
| `batch_decode_graph` CPU enqueue | 0.134 ms average |
| Token/position H2D | 0.007 ms average |
| Paged-KV metadata sync | 0.045 ms average |
| Bridge output batch | 14.86 tokens average, p50 16 |
| Bridge msgpack | 0.013 ms average |
| ZMQ send | 0.012 ms average |

Unified ticks are only 6% of ticks but consume 28.6% of scheduler wall time. The 47 Unified ticks occur during admission/ramp-up and follow the source order of scheduled prefill first, then active decode. This supports the scheduler/executor path as the next optimization target. It does not prove that every arrival delays every active row. Bridge serialization, output transport, and decode metadata sync are too small to explain the remaining gap in this trace.

The trace used wall-clock timing around asynchronous graph launch. Its sampling/wait bucket includes GPU completion, so it does not isolate steady decode GPU time. A direct c16 diagnostic measured about 7.96 ms steady TPOT, but that is attribution evidence only and is not used as HTTP serving evidence.

## Design constraints

A fix may use split-concurrent prefill/decode execution, a scheduler policy that gives active decode a bounded turn, or another measured approach. Qwen3's decode-overlap path is useful prior art, but Qwen3.5 has separate recurrent/conv state and scratch ownership that must be checked explicitly.

The policy must account for both sides of the tradeoff:

- Delaying prefill can improve active-request TPOT while hurting new-request TTFT.
- Any chunk-budget change belongs with #470 and must account for its measured low-QPS ITL p99 tradeoff.
- Concurrent streams require independent scratch/state ownership and must preserve CUDA Graph correctness.
- TP eager behavior is outside the initial single-GPU scope and must keep its current explicit behavior.

## Acceptance criteria

- Add a scheduler/executor test for active decode plus a scheduled prefill chunk. The test must encode the chosen bound: decode runs before the chunk, concurrently with it, or after no more than a stated token/time budget. Merely asserting that Unified eventually returns a decode result is insufficient.
- Keep `e2e_scheduler` and `hf_golden_gate` green on a real Qwen3.5-4B GPU run.
- Run same-host before/after HTTP `1024/256` c1 and c16, 3 repetitions each, with prefix cache disabled.
- Re-run QPS 8/12/16 or explain any unsupported/timeout/OOM cells. Keep every failed cell in the report.
- For every cell, record exact serve and bench commands, engine commit/version, completed/failed/timeout counts, average output tokens, output sanity/hash, TTFT, TPOT, ITL, and output throughput.
- Retain a serving trace that shows scheduler plan mix, Unified wall time, active decode gaps during admission, real decode batch/bucket distribution, GPU step timing if available, and frontend output cost.
- Show a repeated c16 TPOT improvement beyond run-to-run noise without new failures or a Qwen3.5 accuracy regression. Report c1 TTFT/TPOT. If the change alters chunking or overlap, coordinate with #470 and report its mixed-load ITL regression gate.

## Non-goals

- No prefix-cache comparison in the primary result.
- No bridge batching or metadata-sync rewrite without new evidence that those costs became material.
- No kernel-first optimization unless CUDA-event or profiler evidence shows the serving decode GPU step itself has grown to about 15 ms or more.
- No vLLM parity, SOTA, production-ready, or broad hardware claim from this issue.

CAICAIIs:
## Summary

Closes #715.

Qwen3.5 previously executed Unified prefill chunks serially before active decode. This PR adds an opt-in Shared-SM overlap path so one prefill chunk can run on a dedicated CUDA stream while active decode continues on the model stream.

The default remains unchanged: `--decode-overlap off`.

## Changes

- Add `Qwen35DecodeOverlap::{Off, SharedSm}`.
- Add `--decode-overlap off|stream` to the server and serving benchmark CLI.
- Run at most one in-flight prefill on an owned CUDA stream.
- Poll completion events while decode is active and wait explicitly when the last decoder retires.
- Drain the prefill stream on error, cancellation, drop, and shutdown.
- Route prefill and decode through separate cuBLAS handles.
- Keep stream-specific routing limited to Qwen3.5 async prefill.
- Add mode-aware `ITL_STEP` actions:
  `overlap_launch`, `overlap_decode`, `overlap_complete`, and `overlap_wait`.
- Reject unsupported combinations before model loading:
  - tensor parallelism with overlap;
  - `scheduler-policy=auto` with `decode-overlap=stream`;
  - Qwen3.5 Green Context overlap.

## Results

Verified on 1x NVIDIA RTX 5090 32 GB, CUDA 12.8, `sm_120`, with Qwen3.5-4B.

| Workload | Off | Stream | Change |
| --- | ---: | ---: | ---: |
| 1024/256, c1 TPOT | 6.99 ms | 7.07 ms | +1.20% |
| 1024/256, c16 TPOT | 11.48 ms | 11.13 ms | -3.04% |
| 1024/128, QPS 8 TPOT | 12.12 ms | 11.31 ms | -6.68% |
| 1024/128, QPS 12 TPOT | 14.85 ms | 13.70 ms | -7.74% |
| 1024/128, QPS 16 TPOT | 16.88 ms | 15.52 ms | -8.05% |
| Mixed-load ITL p99 | 58.23 ms | 26.15 ms | -55.1% |

All HTTP runs completed with zero failures, timeouts, error strings, or zero-token outputs. The c1 tradeoff and higher loaded TTFT keep overlap opt-in.

Nsight Systems confirmed concurrent prefill/decode kernel execution across the two CUDA streams. The trace contained 28,082 cross-stream kernel-overlap pairs.

## Verification

- Release workspace library tests: `73 passed, 6 ignored`.
- Qwen3.5 default-off scheduler e2e: passed.
- Shared-SM lifecycle e2e: passed; observed two `overlap_wait` actions.
- Qwen3.5 HF golden gate: `2 passed, 2 ignored`.
- Qwen3 and Qwen3.5 release checks: passed.
- Stream-routing regression test: passed.
- `cargo fmt --all -- --check`: passed.
- `git diff --check`: passed.

The evidence covers single-GPU Qwen3.5-4B on RTX 5090. It does not claim tensor-parallel overlap, Green Context support, prefix-cache coverage, broad hardware coverage, vLLM parity, SOTA, or production readiness.

### DIFF
--- a/docs/models/qwen35/unified-prefill-overlap.md
+++ b/docs/models/qwen35/unified-prefill-overlap.md
@@ -90,6 +90,11 @@ Unsupported combinations fail before model loading:
   Prefill-associated actions now publish their own active-width distribution;
   a synthetic overlap launch/decode/complete/wait log reports `decode_n=4`
   while counting the 8192-token chunk once.
+- RLCR found that treating every `StreamOverrideGuard` as prefill also disabled
+  graph-safe N=1 GEMMs for existing Qwen3 decode/Green Context overrides. A
+  nested, RAII-restored prefill marker now narrows that routing rule to Qwen3.5
+  async prefill. Its unit test and affected Qwen3/Qwen3.5 builds passed, and the
+  Shared-SM lifecycle gate still observed two `overlap_wait` actions.
 
 ## Verification Contract
 
@@ -98,7 +103,7 @@ Unsupported combinations fail before model loading:
 | GPU | 1x NVIDIA GeForce RTX 5090, 32 GB |
 | Driver / CUDA toolkit | `595.71.05` / `12.8.93`, target `sm_120` |
 | Rust / Triton build env | nightly 2026-07-10 / Triton 3.7.1 |
-| Source | upstream/main `e3f91120` plus this patch; validation source matched the local tree by checksum |
+| Source | benchmark binaries: `c405b556` on upstream/main `e3f91120`; the RLCR follow-up only narrows generic-vs-prefill override classification and was separately GPU-lifecycle verified |
 | Model | `Qwen/Qwen3.5-4B`, revision `851bf6e806efd8d0a36b00ddf55e13ccb7b8cd0a` |
 | HTTP client | vLLM 0.25.1 `bench serve` only; no vLLM server was used |
 | Server binary sha256 | `5f19998277f4be7577846cf3e78bc3fef43151a5add0b0a56dd443a06d39b091` |
--- a/docs/models/qwen35/unified-prefill-overlap.md
+++ b/docs/models/qwen35/unified-prefill-overlap.md
@@ -2,11 +2,11 @@
 
 > **TL;DR:** Issue #715 adds opt-in, single-GPU `--decode-overlap stream` for
 > Qwen3.5. One prefill chunk may run on a second CUDA stream while active decode
-> continues. On one RTX 5090, three-run HTTP c16 mean TPOT improved 3.04% and
-> QPS 8/12/16 mean TPOT improved 6.68%/7.74%/8.05%; c1 regressed 1.20%, so the
-> default remains serial `off`. Stream mode is currently limited to
-> `--max-batch <= 32` until the bucket-64 decode GEMMs have an independent
-> non-prefill cuBLAS route.
+> continues. The default remains serial `off`. Stream mode is currently limited
+> to `--max-batch <= 32` until the bucket-64 decode GEMMs have an independent
+> per-stream cuBLAS route. The original HTTP table below predates that cap and
+> used the server's implicit Qwen3.5 max-batch default, so treat it as pre-cap
+> evidence until the HTTP cells are re-run with an explicit safe max batch.
 >
 > **Last touched:** 2026-08
 
@@ -98,6 +98,17 @@ Unsupported combinations fail before model loading:
   nested, RAII-restored prefill marker now narrows that routing rule to Qwen3.5
   async prefill. Its unit test and affected Qwen3/Qwen3.5 builds passed, and the
   Shared-SM lifecycle gate still observed two `overlap_wait` actions.
+- Follow-up review after the PegaInfer rename fixed the `bench_serving` Qwen3.5
+  crate path, made the `MAX_SHARED_SM_DECODE_BATCH <= GEMM_LT_MAX_N` relationship
+  a compile-time assertion, and added a back-reference at the decode GEMM tuning
+  filter. It also removed the dead prefill cuBLAS workspace allocation: cuBLAS
+  resets a handle's workspace when `cublasSetStream()` is called, so the safe
+  bucket cap is about avoiding undocumented concurrent use of one cuBLAS handle
+  across decode replay and async prefill until bucket-64 has a per-stream route.
+- The HTTP A/B table is now explicitly marked pre-cap evidence. The archived
+  server launch command was not retained, and the old source defaulted omitted
+  Qwen3.5 `--max-batch` to 64. Current `stream` mode rejects that default, so
+  cite the HTTP numbers only after a re-run with explicit `--max-batch <= 32`.
 
 ## Verification Contract
 
@@ -111,6 +122,7 @@ Unsupported combinations fail before model loading:
 | HTTP client | vLLM 0.25.1 `bench serve` only; no vLLM server was used |
 | Server binary sha256 | `5f19998277f4be7577846cf3e78bc3fef43151a5add0b0a56dd443a06d39b091` |
 | `bench_serving` sha256 | `cf5040aa4bbad60011a3fa3799f23d8aa2bae8ef399f35bd5a7872a7d98c1936` |
+| Current HTTP rerun requirement | Use `--decode-overlap stream --max-batch <= 32`; the server default is 64 for Qwen3.5 and is intentionally rejected in stream mode |
 
 Qwen3.5 does not expose product prefix reuse or accept
 `--no-prefix-cache`; the CLI rejects it as an unused option. The primary result
@@ -129,7 +141,7 @@ flag.
 - Release server and `bench_serving` builds passed. Startup rejection probes
   passed for TP, `Auto + SharedSm`, and Green Context.
 
-### HTTP A/B
+### HTTP A/B (pre-cap evidence; needs explicit max-batch rerun)
 
 The same release binary served both modes. Each round ran `off` then `stream`
 with a 15-second cooldown. Fixed-concurrency cells used request rate `inf`;
@@ -147,8 +159,27 @@ vllm bench serve \
   --save-result --save-detailed
 ```
 
-Median of three runs. `ok/fail` is the total across all three. TPOT is the
-median run's mean TPOT; the latency columns retain p50/p99 from the median run.
+The exact historical server launch command was not retained in the archived
+notes. Inspecting the benchmark source commit shows Qwen3.5 server startup
+filled an omitted `--max-batch` from Qwen3.5 `MAX_DECODE_BATCH`, so omitting
+`--max-batch` meant the HTTP table used the implicit 64-slot default.
+Current stream mode rejects that shape; use commands like these for a current
+HTTP rerun:
+
+```bash
+cargo build --release -p pegainfer-server --features qwen35
+
+target/release/pegainfer \
+  --model-path <model> --served-model-name qwen35-715 --port <port> \
+  --cuda-graph=true --qwen35-scheduler-policy off \
+  --max-batch 32 --max-prefill-tokens 1024 \
+  --decode-overlap <off|stream>
+```
+
+Median of three historical runs. `ok/fail` is the total across all three. TPOT is
+the median run's mean TPOT; the latency columns retain p50/p99 from the median
+run. These rows are retained to show the original direction and output sanity,
+not as the current post-cap HTTP result.
 
 | Cell | mode | ok/fail | observed avg in/out | TTFT p50/p99 ms | TPOT mean/p99 ms | ITL p50/p99 ms | output tok/s |
 | --- | --- | ---: | ---: | ---: | ---: | ---: | ---: |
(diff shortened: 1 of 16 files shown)


pegainfer-server/src/bin/bench_serving/cli.rs
```diff
@@ -89,6 +89,23 @@ pub(crate) enum CliQwen35SchedulerPolicy {
...
+pub(crate) enum CliDecodeOverlap {
+    #[default]
+    Off,
+    Stream,
+}
+
+impl CliDecodeOverlap {
+    #[cfg(feature = "qwen35")]
+    pub(crate) fn resolve(self) -> openinfer_qwen35::Qwen35DecodeOverlap {
+        match self {
```
FeathBow: should be `pegainfer`. As-is, `cargo build --features qwen35` fails with E0433.

btw consider unifying this enum with the same-named one in `pegainfer-server/src/config.rs` (different variants and fallibility).

docs/models/qwen35/unified-prefill-overlap.md
```diff
@@ -0,0 +1,269 @@
...
+  Shared-SM lifecycle gate still observed two `overlap_wait` actions.
+
+## Verification Contract
+
+| Field | Value |
+| --- | --- |
+| GPU | 1x NVIDIA GeForce RTX 5090, 32 GB |
+| Driver / CUDA toolkit | `595.71.05` / `12.8.93`, target `sm_120` |
+| Rust / Triton build env | nightly 2026-07-10 / Triton 3.7.1 |
+| Source | benchmark binaries: `c405b556` on upstream/main `e3f91120`; the RLCR follow-up only narrows generic-vs-prefill override classification and was separately GPU-lifecycle verified |
```
FeathBow, line 120: Could you add the exact server command (`--max-batch`, `--max-prefill-tokens`, CUDA Graph, features)? These binaries predate the `--max-batch <= 32` gate and the server default (64) no longer starts — if the runs used the default, the table needs a re-run at `<= 32`.

pegainfer-qwen35/src/lib.rs
```diff
@@ -34,6 +34,14 @@ pub use scheduler::DEFAULT_MAX_PREFILL_TOKENS;
 /// Maximum supported Qwen3.5 decode scheduler slots.
 pub const MAX_DECODE_BATCH: usize = batch_decode_graph::MAX_BATCH;
 
+/// Current safe Qwen3.5 Shared-SM overlap admission cap.
+///
+/// Decode buckets above this value fall back to the workspace-backed prefill
+/// cuBLAS handle today; running them while async prefill is in flight would make
+/// two streams share that handle/workspace. Keep overlap opt-in below that
+/// boundary until the larger decode buckets have an independent GEMM route.
+pub const MAX_SHARED_SM_DECODE_BATCH: usize = ops::GEMM_LT_MAX_N;
```
FeathBow, line 45: Two suggestions here

1. Add `const _: () = assert!(MAX_SHARED_SM_DECODE_BATCH <= ops::GEMM_LT_MAX_N);` and a back-reference at the tune filter (`weights.rs:691`) — the equality is load-bearing. (Non-bucket `max_batch` is fine: the loader rounds at `weights.rs:203`; verified at mb=5.)

2. The "workspace-backed" rationale in this comment needs an update: cuBLAS 12.9 says `cublasSetStream()` "unconditionally resets the cuBLAS library workspace back to the default workspace pool", and `gemm_cuda` calls it every invocation — so the 32 MiB bound in `cublas_init` (`linear.cu:234`) detaches after the first GEMM and is a dead allocation today (remove it, or re-bind per stream). The accurate rationale for the gate: both paths share one handle's default pool, replay-vs-concurrent-call isolation is undocumented, and the docs flag multi-stream single-handle as nondeterministic — per-stream workspaces/handles are the documented fix, and the clean path to bucket-64 later. The gate itself is right; my sm_89 probes found no corruption, which bounds nothing on other archs.

FeathBow (commented):
Ran this on sm_89, CUDA 12.9, Qwen3.5-4B.

| # | Check | Result |
|---|---|---|
| 1 | qwen35 lib tests / kernels lib tests | 74 + 15 passed |
| 2 | Qwen3.5 `hf_golden_gate` | 2 passed, deltas match your table |
| 3 | Qwen3 `hf_golden_gate` (cross-model probe of the shared kernels change) | 1 passed |
| 4 | Shared-SM lifecycle e2e | pass |
| 5 | Boundary parity: `bg = max_batch - 1` decoders so bucket 4/32 actually replays during an 8192-token prefill, overlap witnessed structurally, all 16 injection steps asserted bitwise (token id, selected logprob bits, top-5 ids and value bits) | pass at both boundaries |
| 6 | Gate-lifted probe: `max_batch=64`, 40 decoders overlapping successive prefills, oracles on both sides, off-vs-off / stream-vs-stream controls | no corruption; all divergence reproduces in off-vs-off |

Not covered: TP2, sm_120, HTTP layer, performance (cross-arch numbers aren't comparable).

CAICAIIs:
> Ran this on sm_89, CUDA 12.9, Qwen3.5-4B.
> 
> #	Check	Result
> 1	qwen35 lib tests / kernels lib tests	74 + 15 passed
> 2	Qwen3.5 `hf_golden_gate`	2 passed, deltas match your table
> 3	Qwen3 `hf_golden_gate` (cross-model probe of the shared kernels change)	1 passed
> 4	Shared-SM lifecycle e2e	pass
> 5	Boundary parity: `bg = max_batch - 1` decoders so bucket 4/32 actually replays during an 8192-token prefill, overlap witnessed structurally, all 16 injection steps asserted bitwise (token id, selected logprob bits, top-5 ids and value bits)	pass at both boundaries
> 6	Gate-lifted probe: `max_batch=64`, 40 decoders overlapping successive prefills, oracles on both sides, off-vs-off / stream-vs-stream controls	no corruption; all divergence reproduces in off-vs-off
> Not covered: TP2, sm_120, HTTP layer, performance (cross-arch numbers aren't comparable).

Thanks for the validation! I pushed a follow-up commit (`7b32330d`) addressing the review notes: the `bench_serving` Qwen3.5 crate rename, the Shared-SM max-batch boundary, the stale cuBLAS workspace rationale, and the pre-cap HTTP evidence wording.

I also validated the follow-up on RTX 5090 / sm_120 / CUDA 12.8:
- `bench_serving` qwen35 compile: pass
- Qwen3.5 `e2e_scheduler`: pass (TP2 ignored on single GPU)
- Qwen3.5 `hf_golden_gate`: pass (TP2 ignored on single GPU)

FeathBow (commented):
LGTM on the impl :)

Two non-blocking suggestions: retain decoder-side parity in the Shared-SM E2E, and rerun the HTTP A/B with explicit --max-batch 32.

CAICAIIs:
> LGTM on the impl :)
> 
> Two non-blocking suggestions: retain decoder-side parity in the Shared-SM E2E, and rerun the HTTP A/B with explicit --max-batch 32.

thanks！

FeathBow (approved):

FeathBow (commented):
The frontend has changed, so before merging, we need to rebase and then re-run checks.

Issue #799, opened 2026-07-31
GLM5.2: synchronous pegaflow restore/save on the engine thread — fleet-wide ITL p99 tail (EP), head-of-line prefill stall (TP4 P)


xiaguan:
## Summary

The pegaflow host-tier restore and save legs still run synchronously on the per-rank engine thread. That was harmless in the coordinator era (every rank joined at step boundaries anyway), but free-running (#795) turned any per-rank blocking into a whole-fleet stall: EP ranks step unconditionally into the fixed-cadence DeepEP chain (`scheduler/mod.rs:430`), so one rank pausing makes all peers wait inside the collective.

Measured on a 1P (TP4) + EP16 (4-node Slurm fleet) native-MTP P/D stack, multi-turn chat workload (48 conv x 5 turns, 1024+256/turn in, 128/turn out, temp 0, 64 GiB host pool), artifacts `bench-results/glm52-pd-multiturn-*-ep16.json`.

## Evidence

**D-side (confirmed by control run):** multi-turn c16 ITL p50/p90/p99 = 30.6/38.7/**98.0** ms; same fleet, same per-rank load, single-turn decode-heavy c16 (no mid-flight admissions) = 31.6/32.1/**40.3** ms. The p99 tail is admission-coupled, not decode-plane. Each turn's admission full-history restore (~130 MB by turn 5 without session affinity) blocks that rank's serve loop; 15 peers eat the pause in the DeepEP chain.

**P-side (structural, same code path):** TP4 prefill-only enables the identical admission restore (prefill-only forces the prefix cache on; `vllm_pd` is none), on 4 mirrored engines running the same serve loop, synced per layer by NCCL AR — one rank's restore delays the chunk for all 4 ranks and head-of-line-blocks every queued prefill's TTFT.

**Suspected same-family:** an episodic 2–6 s stall of one whole phase-locked wave (moves between turns run to run, clean D logs) on the EP4 shape; prime suspect is the P-side save/publish burst when N conversations release + re-arrive simultaneously.

## Code

- `openinfer-glm52/src/scheduler/offload.rs:923` `restore_host_prefix` — doc says 'Blocks on the load — admission is a step boundary'; the block is `handle.wait()` at `offload.rs:959`. The 'every rank is joined' safety argument is coordinator-era and no longer holds.
- Called from `scheduler/admission.rs:352-356`, which runs inside the engine serve loop via `self.admit()` (`scheduler/mod.rs:420-424`).
[...]

PR #801, opened 2026-07-31
fix(glm52): async admission restore (#799) + remove vLLM-compat P/D

xiaguan:
## Summary
- **#799 (D-side): admission restores no longer block the engine loop.** Both the native-MTP P/D restore and the plain host-tier restore previously called \`handle.wait()\` on the engine thread; under free-running DP one rank stalled in admission stalls every peer in the fixed-cadence DeepEP chain (measured: multi-turn c16 ITL p99 98 ms vs 40 ms single-turn control). Loads are now submitted and polled via \`LoadHandle::poll()\` at step boundaries while the request parks at its queue front.
  - Abandoned loads (disconnect, replaced front, deadline reject) keep their destination pages on a scrap list until the DMA settles — the copy still writes those pages after the handle drops.
  - Pages held by a parked front's own in-flight load are credited back to the admission physical budget, otherwise the front's own reservation double-counts against its need and wedges the queue.
  - Trade-off: each restore leg costs one extra step-boundary latency (~25–45 ms) on the restoring request's TTFT, in exchange for not stalling the whole fleet's decode cadence.
- **Remove the vLLM-compat P/D prefill path.** The supported deployment is the native-MTP handoff; the vLLM 0.24 prefill peer path (#657) is retired: VllmPdState admission (miss breaker, local-prefill fallback), vLLM hash keying + arena renaming, the RoPE deinterleave fixup kernel, \`Glm52VllmCompatOptions\`, and the glm52 wiring of the \`--kv-pd-vllm-*\` flags (qwen3 keeps its own vLLM-compat path and the shared flags). \`pd-m2-execution.md\` is deleted with it.

Remaining #799 items (P-side \`save_native_tail\` off the engine thread, save submit-depth audit, session-affinity routing) are follow-ups.

## Test plan
- [x] \`cargo test --release -p openinfer-glm52 --lib\` — 107 passed; new offline contract tests cover park→settle→admit, failed-load degradation, abandoned-load page holds, native pending-load scrap on clear/new-front (only pre-existing env-dependent \`moe_tp::scale_staging_geometry\` fails without GPU, same on main)
- [x] \`cargo test --release --workspace --lib\` — green except GPU-requiring kv_pool tests (no GPU in build container, same on main)
- [x] \`cargo build --release --features glm52\`
- [x] Live EP16 P/D multi-turn re-bench — ITL p99 98→31 ms (c16), 74→30 ms (c64), zero rejects (see comment)


xiaguan:
Live EP16 P/D multi-turn re-bench done (same 4-node fleet/nodes/router as the baseline, Slurm job 5268 with this branch's binary on D; P unchanged). Multi-turn 1024/256/128 × 5 turns, temp 0:

| metric | c16 old → new | c64 old → new |
| --- | --- | --- |
| ITL p99 | 98.0 → **30.7 ms** | 73.7 → **30.4 ms** |
| ITL p90 | 38.7 → 29.9 | 37.6 → 29.2 |
| TPOT p50 | 11.6 → 9.6 | 11.5 → 9.9 |
| TTFT p50 | 524 → 645 ms | 5.18 → 5.24 s |
| TTFT p99 | 1.28 → 1.73 s | 8.41 → 8.73 s |
| out tok/s | 948 → 1,018 | 1,158 → 1,156 |

240/240 and 640/640 turns, zero rejects. ITL p99 now within ~3 ms of p50 at both concurrencies — the admission-coupled tail is gone; the predicted low-load TTFT tax (~one step-boundary round trip per restore leg) shows at c16 and disappears into queueing at c64.

xiaguan:
Addressed both Codex findings in 2cb9174:

1. **Teardown vs outstanding loads** — real gap. The serve loop cannot exit with a *parked* front (it sits in `pending`, which must drain first), but an *abandoned* load (disconnect) can outlive the queue. Teardown now runs a deadline-bounded (5 s) drain barrier over pending + abandoned loads before the offload engines — and after them the workers' arena memory — drop, mirroring the existing `flush_saves` barrier for the D2H side. Covered by `teardown_drain_settles_outstanding_loads_before_engines_drop`.
2. **serving-status.md capability row** — updated the P/D row to the native-MTP handoff; it still described the removed vLLM-prefill path.

xiaguan:
Follow-up design filed as #802 — unifying the restore/save state machines into one pollable OffloadTask (covers the remaining #799 engine-side items and enables restore pipelining). This PR stands as-is; #802 is the refactor on top.

PR #809, opened 2026-07-31
docs(gemma4): specify the two-group KV cache contract

### LINKED ISSUE #807: gemma4: define the KV cache contract for the two attention groups
## Description

Gemma 4 caches at two head dims — 256 for its sliding layers, 512 for its full-attention layers — and the shared `KvLayout` applies one head dim uniformly to every layer. Heterogeneous buffers over one block pool is already an established pattern here, so that part is not the problem. What is new is that the two groups want different lifetimes: the sliding group needs to release pages the full-attention group is still reading, and a request holds one assignment list, so a block is held or released for both groups at once.

Two kernel prerequisites block correct serving rather than one. The sliding mask is required because sliding layers must not attend past 1024 positions. Paged attention instantiated at head_dim 512 is required because the full-attention group's pages are otherwise unreadable — the shared translation unit compiles nothing at that width. Both must exist before the model serves correctly, and the mask needs an explicit conversion — Gemma counts the current token inside its window while FlashInfer's `window_left` is an inclusive distance, so 1024 has to be passed as 1023 or the model attends one token too far. With both in place and no reclamation the model is correct and budget-limited, reaching tens of thousands of positions; the declared 262144 is what it cannot reach, at 84 GiB for one unreclaimed request at the smallest supported size. Reclamation is not purely a capacity exercise either: it splits an array that today carries two meanings, which is a correctness problem of its own.

Several decisions here constrain work that follows and are cheaper to settle now than to discover: whether K and V can share storage, how tensor parallelism shards a group with a single KV head, what CUDA graph capture has to duplicate, and how the prefill chunk size becomes a capacity parameter rather than a latency knob.

## Proposed Plan

- State the two group layouts with their per-token and per-request costs at all three sizes, derived from the published configs rather than from one size.
- Separate the two kernel prerequisites from the capacity work, and say what a correct engine can serve before any eviction exists.
- Enumerate what page eviction breaks, each read in the tree — including what it does *not* break, so the implementation does not chase phantom obstacles.
- Decide the pool topology with the failure mode that rules out the simpler option stated structurally rather than as a race.
- Settle `attention_k_eq_v`, the per-group admission and reservation arithmetic, the tensor-parallel sharding rule, what graph capture duplicates, the RoPE budget, and the explicit exclusions.

## Acceptance Criteria

- Every structural number is stated for all three sizes and is reproducible from the published configs.
- Every claim about existing engine behaviour cites the code it was read from; claims that do not survive checking are corrected rather than repeated, including the ones the document previously made.
- Correctness prerequisites and capacity work are costed separately, so a reader can tell which is which.
- Where the contract says an existing mechanism is missing, it is missing — not merely unfamiliar.
- The document says what it cannot settle and what would settle it.

FeathBow:
## Description

Closes #807 

## Description

`docs/models/gemma4/kv-cache.md` records the KV contract for Gemma 4. No code is included; no engine for this model line exists yet, and these decisions constrain the kernel, RoPE and scheduler work that follows.

Gemma 4 caches at two head dims with two lifetimes. Heterogeneous buffers over one pool is the established pattern — `BlockPool` owns no GPU memory and glm52 already runs one arena width per layer plus a narrower one on a layer subset — so the buffer side is a copy job, and one pool stays correct for as long as nothing is reclaimed. Reclaiming sliding pages is what forces the split: a request holds one assignment list, so releasing a sliding page releases the full-attention page at the same id, and keeping it reclaims nothing. There is no third state. Two pools are therefore the precondition for independent reclamation rather than a correctness requirement for short contexts.

Two attention backends are prerequisites for correct serving. A sliding-window instantiation — added alongside the existing variant rather than replacing it, so other model lines are unaffected — entered as `sliding_window - 1`, because Gemma counts the current token inside its 1024-token window while FlashInfer's `window_left` is an inclusive distance. And paged attention at head_dim 512, without which the full-attention group's pages are unreadable; the shared translation unit is built for other widths and rejects that one at runtime. A property of the second is load-bearing here: FlashInfer's decode dispatcher instantiates a fixed set of GQA groups, and at TP1 12B's full-attention group is 16 query heads over a single KV head, outside that set, so the two groups take structurally different decode paths — and because query heads shard while a lone KV head can only be replicated, that ratio is per rank rather than per config.

The mask also needs a conversion the contract now pins: Gemma counts the current token inside its 1024-token window, FlashInfer's `window_left` is an inclusive distance, so the value passed is `sliding_window - 1`. Passing 1024 straight through attends one token too far, and nothing here would catch it.

Most of what follows is capacity — unreclaimed, one 12B request at the declared 262144 positions wants 84 GiB across both groups; reclaimed it wants about 4.6 GiB — but reclamation itself is not. It breaks the write path rather than the view: `positions` is one array serving both the absolute RoPE position and a view-relative append slot, and a page row that does not start at 0 turns the second into an out-of-bounds write through an unprotected pointer. The view constructor itself is already position-relative and needs no change, which the document states explicitly so the implementation does not go looking for a shared-API problem that is not there.

The document also covers what the earlier draft omitted. Qwen3's "KV heads must divide the world size" tensor-parallel policy cannot be reused here: the full-attention group has one KV head at 12B and has to be replicated rather than sharded — which also means its GQA group is a per-rank property, group 16 at TP1 but back inside the compiled decode set at TP2 and above, so backend selection and graph capture must key off the resolved mapping rather than the config ratio. Sliding residency is a window plus a prefill burst, and since `max_prefill_tokens` caps the step's total across requests rather than each request's, that burst is bought once for the step; charging it per request overstates an eight-way pool by 1.8×. Admission does not allocate, so the new work is a group-specific reservation function rather than a rollback mechanism that already exists. CUDA graph capture is compatible with a shrinking page row, with the open question being whether the full-attention group is captured at all, since Qwen3 skips capture for exactly the GQA groups its decode kernel cannot instantiate.

## Type of Change

- [x] Documentation update


FeathBow:
Two follow-up commits on top of the original contract, both correcting things the first version got wrong rather than adding scope.

Sliding pool bound. The formula rounded the step's total prefill tokens to pages once, which undercounts, because every active row rounds its own frontier independently. With page size 16 and a 1024 window, eight rows taking [1003, 3×7] — exactly the 1024 budget — need [128, 66×7] = 590 pages where the two-term formula gives 584. Six pages, 30 MiB at 12B, and the failure mode is admission accepting a request the second pool then cannot schedule. There is now a third term, (concurrency - 1), with the text noting that summing each row's real offset and chunk is the exact form and this is the cheap upper bound. The eight-way example moves to 591 pages, 2.89 GiB; the shared-burst form is still 1.75× tighter than charging the burst per request.

Per-rank capacity. The contract previously deferred this to "must say whether bytes are charged once or once per rank". It now says. Each rank allocates its own pools, and the full-attention group has two rank mappings rather than one: while the global KV head count is at least the world size, each rank owns a contiguous run of them and the group's bytes divide; below that, each head is replicated to a contiguous span of ranks, leaving one head per rank and a per-token cost that stops shrinking. Every capacity figure in the document is now marked per rank, the existing tables as TP1, with 12B scaling 84 / 44 / 24 GiB per rank at TP1 / TP2 / TP4.

An earlier claim that replicated bytes never divide was wrong — that is only the second regime. It holds at 12B because the single global KV head is already at that floor at TP1, but 26B and 31B do divide until they bottom out (31B: 80 → 40 → 20 KiB per token across TP1/2/4).

Legal world sizes. Integer division on its own is not a contract, since a world size dividing neither head count drops heads without erroring: 12B at 6 covers 12 of 16 query heads and 6 of 8 sliding KV heads. The three conditions a legal world size satisfies are now stated (Q % P == 0, Kv % P == 0, G % P == 0 || P % G == 0) and added to the loader's validation list, called out as distinct from the GQA divisibility already there — the two are easy to conflate, which is how the world-size one went missing.

Reclamation prerequisites. The body already said the assignment list truncates from the end and the only partial drop is Idle-only LIFO, but the closing section reduced the blocking question to the cache-relative positions array alone. RequestKv::prefill_view and decode_view both build KvView's seq_len from the absolute kv_position, so a front-trimmed page row trips the exact-cover assertion on the next step even though the constructor itself is position-relative. The closing section now lists three prerequisites — the positions array, resident origin and length as first-class request state, and an ownership path for releasing a live request's oldest assigned blocks — and marks which are local to this model line and which may not be. The third has no existing analogue.

RoPE sizing. The section quoted the post-change buffer size as the current cost. On the present signature one parameter is both the layout stride and the frequency denominator, so a correct full-attention cache is 512 MiB, not 128; passing 128 today buys the smaller buffer at the price of wrong frequencies. Current cost and required layout are now separate rows: 768 MiB as things stand, 384 MiB once the parameters are split. Any KV budget example assuming 384 MiB is off by that difference.

xiaguan:
Thanks for documenting this. We’re currently redesigning the KV-cache module interface, and that work may change some of the assumptions and API boundaries described here. Could you please convert this PR to draft for now? We can resume review once the interface direction is settled.

Issue #812, opened 2026-08-01
glm52: EP32 production design point — 32 slots, verify-span buckets, configurable drafts, 200K context


xiaguan:
Target operating point (production): decode EP32/EP64, ~24-30 active decode requests per rank, avg context ~200K, MTP drafts 2 (down from 5), ≥500 output tok/s per GPU.

## Measured starting point (2026-08-01, EP4 saturation, `bench-results/glm52-ep4-saturation-*.json`)

| concurrency | node tok/s | per-GPU | TPOT p50 | ITL p50 |
| --- | --- | --- | --- | --- |
| c16 (4 slots/rank) | 558 | 140 | 21.5 ms | 40.1 ms |
| c32 (8 slots/rank, full) | 638 | 160 | 41.6 ms | 41.6 ms |
| c48 (full + queue) | 701 | 175 | 41.8 ms | 41.7 ms |

The smoking gun is TPOT/ITL: at half occupancy the verify span still fits the bucket (≈1.9 tokens/step); at full occupancy TPOT == ITL — **speculation collapses to span 1 because a verify span (1 anchor + N drafts) has no bucket rows left**. The planner degrades span ≈ bucket/slots, and the bucket ceiling (8) equals the slot count.

## Physics that bound the design

- **KV capacity**: MLA ≈ 52.7 KB/token/rank (78×656 B + 21×132 B, fp8). ~250 GiB usable after EP32 weights → ≈4.7 M resident tokens/rank → **≈23-24 concurrent 200K requests**, not 30. Admission's full-lifetime honor-or-reject already self-limits; 32 slots is a cap, not a promise.
- **Long context cost lives in the indexer, not attention**: sparse MLA attends topk-2048 regardless of length, but the DSA indexer scans all keys (132 B/token) — linear in context. Validated only to 16K. At the target step budget (500 tok/s = 30×2/step → **≤120 ms/step**), the 200K indexer term is the go/no-go number and must be measured first.

## Staged plan (each stage independently landable + testable)

[...]

PR #823, opened 2026-08-02
feat(glm52): decouple the KV pool from the slot count — budget-filled pool, 131K cap on 32 slots (#812)

xiaguan:
## Summary

Stacked on #817 (retargets to main when that merges). The KV pool stops being `slots × max_model_len`: the cap stays what it always was — the per-request servable limit, page-table width, admission bound — and the pool block count becomes a **launch-time budget decision**. Every leftover budget byte after the fixed arenas becomes KV capacity, at the exact marginal cost read from the same ledger the cap derivation uses (the ledger is linear in blocks, so the fill lands within one block of the target).

## Why

- A 200K-class cap used to force `32 × 200K` of KV provisioning — either impossible or absurdly wasteful — and the derived cap **shrank as slots grew**. Mixed-context deployments (30 × 100K + 2 × 200K) had no shape at all.
- On prefill-only ranks the pool IS the prefix cache (#811 retains released prefixes in pool headroom); slot-count sizing left most of P's HBM idle.
- Admission already had the right per-request guard: lifetime reservation (`prompt + max_tokens`, honor-or-reject, queue when full). This PR just stops pre-multiplying the worst case into the pool.

## Semantics

- **Defaults unchanged:** the cap derivation keeps its legacy nominal (`slots × cap`) as the search target and the fill floor where affordable — a default launch gets the same cap as before, plus a larger pool.
- **Explicit big caps unlock:** where the nominal doesn't fit, the floor drops to a one-request pool and the fill takes it from there. `--max-model-len 131072` on 32 slots — rejected outright before — now launches and serves.
- **The pool size is MEASURED, not estimated** (second commit): workers build everything except pool-scaled slabs (`BuildFixed`), report `cuMemGetInfo` free VRAM, launch min-reduces and sizes the pool from fact at the exact per-block slab marginal, then workers attach the slabs (`FinishKv`). Exactly-known post-measure charges (DeepGEMM comm buffers, DSpark lane) subtract precisely. The only remaining estimate is `GLM52_GRAPH_RESERVE_GIB` (default 3) for lazy whole-step graph instantiation + runtime workspaces — backstopped by the existing post-build headroom re-probe. An earlier ledger-based fill needed an 8 GiB fudge for exactly the accounting-drift reason the MTP/dspark ledgers were bitten by this week; the measured design removes that failure class. `GLM52_KV_POOL_BLOCKS` pins the count for A/B runs.
- One process-wide block count publishes via `glm52_set_pool_blocks` before any worker builds; the scheduler pool, every cache slab, the MTP page-id mirror, and `kv_total_blocks` all read it (unit tests fall back to legacy sizing).

## Validation (EP4, 32×2 profile; re-run on the measured two-phase build)

- Default 16K cap: measured fill lands **21,640 blocks = 1.38M tokens (2.6× nominal, +469 blocks over the ledger estimate — measurement beats accounting)**, VRAM 279.6/284.2 GiB, build clean, zero failures; c192 decode throughput neutral (2,977.6 tok/s, mean TPOT 32.2 ms — identical to pre-decoupling repeats); greedy smoke byte-identical. The pool log shows the whole equation: min rank free after fixed build, graph reserve, exact post-finish charges, bytes/block.
- **Pool-pressure arbitration**: 12 concurrent ~118K prompts totalling **1.416M tokens against a 1.35M-token pool** — the overflow queued at admission (lifetime reservation) and **12/12 completed, zero failures**. This is the decoupling's core claim: concurrency × context no longer pre-provisioned, arbitrated at admission instead.
- **131,072 cap × 32 slots**: launches (one-request floor; outright rejected pre-decoupling), measured pool 21,544 blocks, and a **57,316-token prompt serves correctly**.
- New solver-invariant test (`budget_fill_spends_the_leftover_on_pool_blocks`); `cargo test -p openinfer-glm52 --lib` green except the known env-only `moe_tp::scale_staging_geometry`.

## Rider

`OPENINFER_KV_TRANSFER_MODE=kernel|direct` (openinfer-kv-offload, default `direct` unchanged): the pegaflow transfer backend was hardcoded to Direct per the #704 decode-co-residency finding; the env knob makes it selectable per deployment. A/B on the prefill-only offload path showed identical curves for both backends — which exposed that host restore isn't wired on prefill-only ranks at all (evicted prefixes recompute; `host_restore` gates on `prefill_chunk_size.is_none()`). The knob stays for when P-side restore lands; the P characterization data is in #812.

## Not in scope

P-side prefix-hit-rate gains (the same fill turns idle prefill-rank HBM into prefix-cache retention) want a multi-turn bench once a P/D window is up — follow-up on #812.

### DIFF
--- a/docs/models/glm52/paged-kv-prefix-cache.md
+++ b/docs/models/glm52/paged-kv-prefix-cache.md
@@ -25,20 +25,34 @@ was static:
 
 ### One pool, two arenas, shared ids
 
-Each rank's `BlockPool(page=64, blocks = 8*ceil((cap+1)/64) + 1)` hands out
-block ids; the rank's per-layer MLA packed caches (656 B/token) and index-K
-caches (full-indexer layers) are both indexed by those ids
-(`glm52_pool_blocks` is the single sizing formula shared by the pool,
-`Glm52RankModel::build`, and the `glm52_arena_bytes` VRAM ledger). The
-trailing `+1` is the reserved **padding page**: padding rows and graph
-pre-capture write there, nobody reads it meaningfully (Kimi's benign-garbage
-argument). The per-slot `cap+1` (not `cap`) keeps 8 concurrent max-shape
-requests (`prompt + max_tokens = cap + 1`) admissible: each really draws
-`ceil((cap+1)/64)` pages including its dangling-token page — a toxic-review
-finding; the naive `8*ceil(cap/64)` pool silently degraded max-shape
-concurrency to 7.
+Each rank's `BlockPool(page=64)` hands out block ids; the rank's per-layer
+MLA packed caches (656 B/token) and index-K caches (full-indexer layers) are
+both indexed by those ids, and the pool, `Glm52RankModel::finish_kv`, and the
+scheduler all share the ONE launch-decided block count.
 
-Total KV VRAM is the per-slot design's capacity + 9 pages (~30 MB); the win
+**Pool sizing is a measured launch decision, not a formula (#823).** The old
+`slots*ceil((cap+1)/64) + 1` provisioning multiplied the servable cap into
+the pool (a 200K cap forced slots×200K of KV — unaffordable — and the derived
+cap shrank as slots grew). Now the workers build everything except the
+pool-scaled slabs, report measured free VRAM, and launch min-reduces and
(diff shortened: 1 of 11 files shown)


xiaguan:
@codex review

xiaguan:
@codex review

PR #848, opened 2026-08-05
fix(qwen3): cap PerToken CUDA Graph buckets

### LINKED ISSUE #780: qwen3: PerToken decode-graph memory grows linearly with batch bucket and is not budgeted
## Summary

Under the PerToken batch-invariance policy, each bucket-N decode CUDA graph bakes **~36 layers × 6 projections × N `cublasGemmEx` nodes**, because the PerToken GEMM oracle (`gemm_per_token_cuda`, `openinfer-kernels/csrc/shared/linear.cu:772`) loops one GEMM per row. Graph-exec memory therefore grows linearly with the bucket size; across all 36 decode buckets this sums to **~1.8 GB** for Qwen3-4B (measured on RTX 5070 Ti 16GB via `cuMemGetInfo` probes during a full bucket sweep: drain starts around bucket 56, ~-134 MB by bucket 224, OOM near bucket 232 with util=0.90).

The KV-cache memory budget (`util`, default 0.90) never reserves this per-policy graph memory, so PerToken + CUDA graph can OOM — deterministically at startup once decode graphs are pre-captured, or mid-serving when capture is lazy.

## What is unaffected

- **Tuned** (production default) and **Pin** policies: their graphs have O(layers) nodes, measured flat at ~2 MB/bucket across the sweep.
- Eager (non-graph) decode.

## Proposed directions

1. **Budget it**: measure/reserve the PerToken graph-exec footprint during memory profiling so the KV pool shrinks to fit (same pattern as the DFlash drafter reservation).
2. **Cap it**: restrict the set of pre-captured buckets under PerToken (large buckets are the expensive ones) and fall back to eager decode above the cap.

Option 1 keeps the batch-invariance envelope intact; option 2 trades coverage of the envelope at high batch.

## Context

- Surfaced by `tests/batch_invariance_decode_gemm_graph.rs` failing with `cuGraphInstantiateWithFlags: CUDA_ERROR_OUT_OF_MEMORY` on a 16GB card at util 0.90; the test now builds with util 0.75 as a workaround (its numerics assertions are pool-size independent).
- Related: #435 (batch-invariant tracking).

## Status: help wanted

We worked around this on our side and are **not planning to implement the fix in-house right now** — handing it to the community. The mechanism and both candidate fixes are scoped above; a single consumer GPU (16GB is enough to reproduce the OOM at util 0.90) is all that's needed to verify. Happy to review a PR.

RicardoMin:
## Description

Fixes #780

PerToken routes projection GEMMs through a row-wise numerical path. CUDA Graph
capture records one GEMM node per token row, so graph executable memory grows
with larger decode buckets.

This change limits PerToken CUDA Graph capture to bucket 32. Batches above the
cap keep the PerToken numerical path but fall back to eager decode. Tensor
parallel startup uses the same rule when performing graph pre-capture and
finalization.

## Type of Change

- [ ] Documentation update

## Validation

- `cargo fmt --all --check`
- `git diff --check`
- `cargo check --release -p pegainfer-qwen3 --lib`
- Single-GPU `batch_invariance_decode_gemm_graph`
- TP2 `tp_pertoken_graph_cap`
- TP2 `tp_concurrent_decode`

Observed:

- Single GPU: batch 32 replays the captured graph; batch 33 executes eager GEMMs twice.
- TP2: `bs=32 served 0->0`; `bs=33 served 506->1012`.

## Checklist

- [x] My code follows the style guidelines of this project.
- [ ] I have run the complete local test suite and all tests pass.


RicardoMin:
The CI workflow for this fork PR is still awaiting maintainer approval.
Could a maintainer approve the run when convenient?

xiaguan:
sure

pegainfer-qwen3/src/batch_decode.rs
```diff
@@ -94,13 +95,32 @@ impl Qwen3Model {
...
+        let requested_bucket = use_cuda_graph.then(|| bucket_for(bs));
+
+        if graph_use == DecodeGraphUse::Serve {
+            if let Some(bucket) = requested_bucket {
+                if !per_token_graph_bucket_allowed(bufs.policy_at_construction, bucket) {
+                    log::debug!(
+                        "PerToken decode bucket {bucket} exceeds graph cap{}; using eager decode",
+                        super::batch_decode_buffers::PERTOKEN_GRAPH_MAX_BUCKET
+                    );
+                    use_cuda_graph = false;
```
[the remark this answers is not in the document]
RicardoMin: Thanks for catching this. The single-GPU memory profile now captures the
largest allowed PerToken graph before running the maximum-row eager probe,
so its executable memory is included in the KV budget. TP profiling remains
eager to avoid uncoordinated NCCL graph capture.

RicardoMin:
@codex review

pegainfer-qwen3/src/batch_decode.rs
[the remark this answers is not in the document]
RicardoMin: Updated docs/models/qwen3/decode-attention.md to document the PerToken
CUDA Graph cap. Buckets up to 32 use graph replay, while larger buckets
retain the PerToken arithmetic but execute eagerly. The document also
distinguishes the SplitKv attention path from the CUDA Graph execution mode.

RicardoMin:
Could a maintainer please approve the pending CI workflow so the checks can run? Thank you!

pegainfer-qwen3/src/unified_forward.rs
```diff
@@ -73,6 +74,28 @@ impl Qwen3Model {
             } else {
                 crate::batch_decode::DecodeGraphUse::Serve
             };
+            // PerToken buckets above 32 intentionally run eager, so the usual
+            // max-row probe would miss the largest graph that serving can still
+            // instantiate. Capture that allowed bucket first on the single-GPU
+            // path so its executable memory is included in the KV budget. TP
+            // keeps this profile eager because ranks are not coordinated here.
+            let policy = numeric_policy();
```
FeathBow: Bucket 32 is not a memory upper bound: BatchDecodeBuffers retains independent graph state for every allowed bucket, so buckets 1, 2, 4,  8, 16, 20, and 24 can still be captured lazily after KV sizing and remain live alongside bucket 32. #780 is specifically a cumulative-residency OOM; please profile or reserve the complete retained set and attach the original 16 GB base-fails/head-passes receipt.

pegainfer-qwen3/src/executor.rs
```diff
@@ -1544,15 +1545,24 @@ impl Qwen3Executor {
                 Ok(())
             };
             run_phase(PrecapturePhase::Warmup)?;
-            for bucket_idx in 0..BATCH_BUCKETS.len() {
+            let graph_policy = pegainfer_kernels::ops::numeric_policy();
```
FeathBow: The public Qwen3 launch path maps `--batch-invariant` to Pin and rejects tensor parallelism for that mode; #780 is also explicitly scoped as verifiable on one consumer GPU. This TP branch is therefore exercised only by callers that directly set the diagnostic PerToken policy and construct the low-level executor, as the new test does.

If that combination is not a supported contract, the TP filtering and 199-line TP test are unnecessary scope for this fix. If it is intended to be supported, please state that contract and retain the following per-rank memory verification as part of the PR evidence.

pegainfer-qwen3/src/batch_decode_buffers.rs
```diff
@@ -26,6 +26,31 @@ pub(crate) const BATCH_BUCKETS: &[usize] = &[
     1, 2, 4, 8, 16, 20, 24, 32, 40, 48, 56, 64, 72, 80, 88, 96, 104, 112, 120, 128, 136, 144, 152,
     160, 168, 176, 184, 192, 200, 208, 216, 224, 232, 240, 248, 256,
 ];
+/// Largest batch bucket for which PerToken keeps a CUDA Graph
+pub(crate) const PERTOKEN_GRAPH_MAX_BUCKET: usize = 32;
```
FeathBow: Bucket 32 is a plausible conservative boundary because #780 first observed visible drain around bucket 56, but the current counters only prove graph-versus-eager routing. They do not show that the complete retained allowed set [1, 2, 4, 8, 16, 20, 24, 32] closes the original 16 GB, utilization-0.90 OOM.

Please attach the exact-head command, GPU/model/CUDA configuration, and a base-fails/head-passes receipt from the same full bucket sweep. A post-sweep cuMemGetInfo reading would make the remaining headroom auditable. Please also report the decode-step or TPOT cost at the 32/33 boundary.

pegainfer-qwen3/src/executor.rs
```diff
@@ -1544,15 +1545,24 @@ impl Qwen3Executor {
...
+            let captured_buckets = BATCH_BUCKETS
+                .iter()
+                .filter(|&&bucket| per_token_graph_bucket_allowed(graph_policy, bucket))
+                .count();
+
+            for (bucket_idx, &bucket) in BATCH_BUCKETS.iter().enumerate() {
```
FeathBow: TP profiling is eager, and the KV pool has already been sized and allocated before this loop captures the eight allowed PerToken graphs on every rank. The new TP counter test proves which path executes, but it does not measure this post-budget residency.

Capping may be sufficient; it just needs evidence. Please either reserve the complete per-rank graph footprint or report the exact  cumulative per-rank allocation/headroom under the configuration used for the TP claim.

pegainfer-qwen3/tests/tp_pertoken_graph_cap.rs
```diff
@@ -0,0 +1,199 @@
...
+    let Ok(path) = std::env::var("PEGAINFER_TEST_MODEL_PATH") else {
+        eprintln!("skipping tp_pertoken_graph_cap: set PEGAINFER_TEST_MODEL_PATH");
+        return None;
+    };
+    Some(path)
+}
+
+/// Number of CUDA devices visible to this test process
+fn cuda_device_count() -> usize {
+    cudarc::driver::CudaContext::device_count().map_or(0, |count| count.max(0) as usize)
```
FeathBow: `map_or(0, ...)` makes a CUDA driver/context failure indistinguishable from a host with zero GPUs. The test then returns successfully, so an explicitly requested TP acceptance run can be green without executing any model code.

Please propagate or panic on `device_count()` errors. Only a successful query that reports fewer than two visible devices should take the availability-skip path.

pegainfer-qwen3/src/batch_decode.rs
```diff
@@ -94,13 +95,32 @@ impl Qwen3Model {
             graphs_available || matches!(graph_use, DecodeGraphUse::Serve | DecodeGraphUse::Eager),
             "batch_decode {graph_use:?} requires CUDA graphs enabled and no LoRA rows"
         );
-        let use_cuda_graph = graphs_available && graph_use != DecodeGraphUse::Eager;
+        // PerToken records one GEMM node per now; cap Large buckets to bound graph memory.
+        // Above the cap, keep the PetToken arithmetic but execute the kernels eagerly
+        let mut use_cuda_graph = graphs_available && graph_use != DecodeGraphUse::Eager;
```
FeathBow: The cap is enforced here only for `Serve`; `CaptureOnly` and `Replay` still accept a disallowed PerToken bucket and rely on separate controller/lane/Finalize filters. The lane currently turns such a request into a silent successful no-op, which can hide a protocol mismatch.

Please give graph eligibility one owner. `Serve` may intentionally choose eager above the cap, while an explicit capture/replay request for a disallowed bucket should be impossible by construction or fail loudly.

pegainfer-qwen3/src/batch_decode_buffers.rs
```diff
@@ -527,10 +552,68 @@ impl BatchDecodeBuffers {
...
+        ));
+        assert!(per_token_graph_bucket_allowed(
+            NumericPolicy::Pin,
+            *super::BATCH_BUCKETS.last().unwrap()
+        ));
+        assert!(per_token_graph_bucket_allowed(
+            NumericPolicy::Tuned,
+            *super::BATCH_BUCKETS.last().unwrap()
+        ));
+        assert_eq!(
```
FeathBow: These assertions mostly restate a local helper and remain green despite the cumulative-accounting bug. The retained 32/33 integration probe already covers the serving boundary. Please remove the supplemental-profile matrix; if the unaffected-policy contract needs direct coverage, keep only a compact check for PerToken, Tuned, and Pin. A full-plan unit test is useful only if profiling and serving consume one canonical graph plan.

pegainfer-qwen3/tests/tp_pertoken_graph_cap.rs
```diff
@@ -0,0 +1,199 @@
+//! TP=2 regression for the PerToken CUDA-Graph bucket cap.
```
FeathBow: Please remove this test unless direct PerToken+TP is a supported low-level contract. If it is supported, retain only the TP-specific pre-capture/barrier contract that the single-GPU gate cannot cover.

FeathBow (changes requested):
Thank you for working on #780. The cap-and-eager direction is reasonable, but the current head does not yet close the issue.

The profile accounts for bucket 32 only. The other seven allowed buckets can still be captured lazily after KV sizing and remain resident alongside it, so the complete eight-graph set is not budgeted. Please account for the retained set and provide an exact-head base-fails/head-passes receipt for the original 16 GB, utilization-0.90 failure. The 32/33 counters prove dispatch mode, not memory closure.

PerToken+TP is not reachable through the public server contract: `--batch-invariant` selects Pin and rejects TP. Unless this low-level combination is intentionally supported, please remove its dedicated implementation and 199-line test. Most of the 50-line helper-return block tests local implementation details. Please remove the supplemental-profile matrix and retain, at most, a compact policy-boundary check for PerToken versus Tuned and Pin.

I’d appreciate it if you could give the full diff a careful self-review pass before requesting re-review. That would help keep the next review focused on the substantive remaining issues :)

RicardoMin:
Dear Reviewer,

Thank you very much for your careful review and valuable feedback on this PR. I sincerely apologize for the oversights in the code. Your feedback has pointed out clear areas for improvement, which I truly appreciate.

I fully understand and accept your comments, and will immediately rectify the issues according to your guidance to ensure they are properly resolved. Once again, apologies for any inconvenience caused, and thanks for your patience and instruction!


Best regards,
RicardoMin




---- Replied Message ----
| From | Salt ***@***.***> |
| Date | 08/07/2026 01:07 |
| To | ***@***.***> |
| Cc | ***@***.***>***@***.***> |
| Subject | Re: [openinfer-project/pegainfer] fix(qwen3): cap PerToken CUDA Graph buckets (PR #848) |

@FeathBow requested changes on this pull request.

Thank you for working on #780. The cap-and-eager direction is reasonable, but the current head does not yet close the issue.

The profile accounts for bucket 32 only. The other seven allowed buckets can still be captured lazily after KV sizing and remain resident alongside it, so the complete eight-graph set is not budgeted. Please account for the retained set and provide an exact-head base-fails/head-passes receipt for the original 16 GB, utilization-0.90 failure. The 32/33 counters prove dispatch mode, not memory closure.

PerToken+TP is not reachable through the public server contract: --batch-invariant selects Pin and rejects TP. Unless this low-level combination is intentionally supported, please remove its dedicated implementation and 199-line test. Most of the 50-line helper-return block tests local implementation details. Please remove the supplemental-profile matrix and retain, at most, a compact policy-boundary check for PerToken versus Tuned and Pin.

I’d appreciate it if you could give the full diff a careful self-review pass before requesting re-review. That would help keep the next review focused on the substantive remaining issues :)

In pegainfer-qwen3/src/unified_forward.rs:

> @@ -73,6 +74,28 @@ impl Qwen3Model { } else { crate::batch_decode::DecodeGraphUse::Serve }; + // PerToken buckets above 32 intentionally run eager, so the usual + // max-row probe would miss the largest graph that serving can still + // instantiate. Capture that allowed bucket first on the single-GPU + // path so its executable memory is included in the KV budget. TP + // keeps this profile eager because ranks are not coordinated here. + let policy = numeric_policy();

Bucket 32 is not a memory upper bound: BatchDecodeBuffers retains independent graph state for every allowed bucket, so buckets 1, 2, 4, 8, 16, 20, and 24 can still be captured lazily after KV sizing and remain live alongside bucket 32. #780 is specifically a cumulative-residency OOM; please profile or reserve the complete retained set and attach the original 16 GB base-fails/head-passes receipt.

In pegainfer-qwen3/src/executor.rs:

> @@ -1544,15 +1545,24 @@ impl Qwen3Executor { Ok(()) }; run_phase(PrecapturePhase::Warmup)?; - for bucket_idx in 0..BATCH_BUCKETS.len() { + let graph_policy = pegainfer_kernels::ops::numeric_policy();

The public Qwen3 launch path maps --batch-invariant to Pin and rejects tensor parallelism for that mode; #780 is also explicitly scoped as verifiable on one consumer GPU. This TP branch is therefore exercised only by callers that directly set the diagnostic PerToken policy and construct the low-level executor, as the new test does.

If that combination is not a supported contract, the TP filtering and 199-line TP test are unnecessary scope for this fix. If it is intended to be supported, please state that contract and retain the following per-rank memory verification as part of the PR evidence.

In pegainfer-qwen3/src/batch_decode_buffers.rs:

> @@ -26,6 +26,31 @@ pub(crate) const BATCH_BUCKETS: &[usize] = &[ 1, 2, 4, 8, 16, 20, 24, 32, 40, 48, 56, 64, 72, 80, 88, 96, 104, 112, 120, 128, 136, 144, 152, 160, 168, 176, 184, 192, 200, 208, 216, 224, 232, 240, 248, 256, ]; +/// Largest batch bucket for which PerToken keeps a CUDA Graph +pub(crate) const PERTOKEN_GRAPH_MAX_BUCKET: usize = 32;

Bucket 32 is a plausible conservative boundary because #780 first observed visible drain around bucket 56, but the current counters only prove graph-versus-eager routing. They do not show that the complete retained allowed set [1, 2, 4, 8, 16, 20, 24, 32] closes the original 16 GB, utilization-0.90 OOM.

Please attach the exact-head command, GPU/model/CUDA configuration, and a base-fails/head-passes receipt from the same full bucket sweep. A post-sweep cuMemGetInfo reading would make the remaining headroom auditable. Please also report the decode-step or TPOT cost at the 32/33 boundary.

In pegainfer-qwen3/src/executor.rs:

> @@ -1544,15 +1545,24 @@ impl Qwen3Executor { Ok(()) }; run_phase(PrecapturePhase::Warmup)?; - for bucket_idx in 0..BATCH_BUCKETS.len() { + let graph_policy = pegainfer_kernels::ops::numeric_policy(); + let captured_buckets = BATCH_BUCKETS + .iter() + .filter(|&&bucket| per_token_graph_bucket_allowed(graph_policy, bucket)) + .count(); + + for (bucket_idx, &bucket) in BATCH_BUCKETS.iter().enumerate() {

TP profiling is eager, and the KV pool has already been sized and allocated before this loop captures the eight allowed PerToken graphs on every rank. The new TP counter test proves which path executes, but it does not measure this post-budget residency.

Capping may be sufficient; it just needs evidence. Please either reserve the complete per-rank graph footprint or report the exact cumulative per-rank allocation/headroom under the configuration used for the TP claim.

In pegainfer-qwen3/tests/tp_pertoken_graph_cap.rs:

> +const PROMPT_LEN: usize = 8; +const AT_CAP_BATCH: usize = 32; +const ABOVE_CAP_BATCH: usize = 33; + +/// Reslove the model path or skip this GPU integration test +fn model_path_or_skip() -> Option<String> { + let Ok(path) = std::env::var("PEGAINFER_TEST_MODEL_PATH") else { + eprintln!("skipping tp_pertoken_graph_cap: set PEGAINFER_TEST_MODEL_PATH"); + return None; + }; + Some(path) +} + +/// Number of CUDA devices visible to this test process +fn cuda_device_count() -> usize { + cudarc::driver::CudaContext::device_count().map_or(0, |count| count.max(0) as usize)

map_or(0, ...) makes a CUDA driver/context failure indistinguishable from a host with zero GPUs. The test then returns successfully, so an explicitly requested TP acceptance run can be green without executing any model code.

Please propagate or panic on device_count() errors. Only a successful query that reports fewer than two visible devices should take the availability-skip path.

In pegainfer-qwen3/src/batch_decode.rs:

> @@ -94,13 +95,32 @@ impl Qwen3Model { graphs_available || matches!(graph_use, DecodeGraphUse::Serve | DecodeGraphUse::Eager), "batch_decode {graph_use:?} requires CUDA graphs enabled and no LoRA rows" ); - let use_cuda_graph = graphs_available && graph_use != DecodeGraphUse::Eager; + // PerToken records one GEMM node per now; cap Large buckets to bound graph memory. + // Above the cap, keep the PetToken arithmetic but execute the kernels eagerly + let mut use_cuda_graph = graphs_available && graph_use != DecodeGraphUse::Eager;

The cap is enforced here only for Serve; CaptureOnly and Replay still accept a disallowed PerToken bucket and rely on separate controller/lane/Finalize filters. The lane currently turns such a request into a silent successful no-op, which can hide a protocol mismatch.

Please give graph eligibility one owner. Serve may intentionally choose eager above the cap, while an explicit capture/replay request for a disallowed bucket should be impossible by construction or fail loudly.

In pegainfer-qwen3/src/batch_decode_buffers.rs:

> + NumericPolicy::PerToken, + PERTOKEN_GRAPH_MAX_BUCKET + )); + assert!(!per_token_graph_bucket_allowed( + NumericPolicy::PerToken, + PERTOKEN_GRAPH_MAX_BUCKET + 1 + )); + assert!(per_token_graph_bucket_allowed( + NumericPolicy::Pin, + *super::BATCH_BUCKETS.last().unwrap() + )); + assert!(per_token_graph_bucket_allowed( + NumericPolicy::Tuned, + *super::BATCH_BUCKETS.last().unwrap() + )); + assert_eq!(

These assertions mostly restate a local helper and remain green despite the cumulative-accounting bug. The retained 32/33 integration probe already covers the serving boundary. Please remove the supplemental-profile matrix; if the unaffected-policy contract needs direct coverage, keep only a compact check for PerToken, Tuned, and Pin. A full-plan unit test is useful only if profiling and serving consume one canonical graph plan.

In pegainfer-qwen3/tests/tp_pertoken_graph_cap.rs:

> @@ -0,0 +1,199 @@ +//! TP=2 regression for the PerToken CUDA-Graph bucket cap.

Please remove this test unless direct PerToken+TP is a supported low-level contract. If it is supported, retain only the TP-specific pre-capture/barrier contract that the single-GPU gate cannot cover.

RicardoMin:
@codex  review

pegainfer-qwen3/src/batch_decode.rs
```diff
@@ -59,6 +60,81 @@ pub(crate) enum DecodeGraphUse {
...
+                    "PerToken decode bucket {bucket} exceeds graph cap \
+                    {PERTOKEN_GRAPH_MAX_BUCKET}; using eager decode"
+                );
+                Ok(None)
+            }
+            DecodeGraphUse::CaptureOnly | DecodeGraphUse::Replay => {
+                anyhow::bail!(
+                    "batch_decode {graph_use:?} requested PerToken CUDA Graph bucket {bucket} \
+                    above cap {PERTOKEN_GRAPH_MAX_BUCKET}"
+                )
```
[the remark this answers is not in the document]
RicardoMin: I don't think this needs to be addressed, since direct PerToken + TP is not a supported configuration. TP should never enter the PerToken path, so I don't plan to add special handling for capped PerToken buckets during TP pre-capture.

RicardoMin:
## Qwen3 PerToken CUDA Graph memory validation

Thank you for the detailed review. I revised the implementation to account for the complete retained Graph set and validated the memory and performance behavior with a same-GPU base/candidate comparison.

Current PR head: `3420aaa8a4c8c9768f6ce04f7588e8c56ba1adaa`  
Comparison base: `57ee51f71e45b3745050fd8c39c6dd88b4fbde5e`

### Test environment

| Item | Value |
|---|---|
| GPU | NVIDIA L4, 23,034 MiB |
| Driver / CUDA | 550.163.01 / 12.4 |
| CUDA target | `sm_89` |
| Model | Qwen3-4B |
| Numeric policy | `PerToken` |
| CUDA Graph | enabled |
| GPU memory utilization | `0.90` |

### Implementation

- PerToken retains CUDA Graphs only for `[1, 2, 4, 8, 16, 20, 24, 32]`.
- All eight retained Graphs are captured cumulatively in the same `BatchDecodeBuffers` during memory profiling, with synchronization before every memory sample.
- Batches above bucket 32 preserve PerToken arithmetic but execute eagerly at their exact batch size.
- Serving and profiling consume the same `DecodeGraphPlan`; explicit ineligible capture/replay requests fail loudly.
- The supplemental profile matrix and the unsupported PerToken+TP implementation/test were removed.

### Full-bucket memory sweep

The same diagnostic exercised all 36 decode buckets twice in one executor and sampled `cuMemGetInfo` after synchronization. Request KV state was dropped after every bucket, while captured Graph executables remained resident.

| Measurement | Base | Candidate | Change |
|---|---:|---:|---:|
| Start free memory | 2,522.12 MiB | 2,528.12 MiB | +6.00 MiB |
| Maximum Graph-eligible buckets | 36 | 8 | **-77.78%** |
| Completed sweep buckets | 20/36 (55.56%) | 36/36 (100%) | **+44.44 pp** |
| Retained memory after bucket 32 | 16.00 MiB | 16.00 MiB | 0% |
| Retained memory after bucket 128 | 2,482.00 MiB | 16.00 MiB | **-99.36%** |
| Free memory after bucket 128 | 40.12 MiB | 2,512.12 MiB | **+2,472.00 MiB** |
| Bucket 136 | CUDA OOM during the following prefill | Completed eagerly | OOM eliminated |
| Full 36-bucket sweep | **FAIL** | **PASS** | Regression closed |
| Final free memory | Not reached | 2,512.12 MiB | - |

Candidate dispatch was:

```text
buckets 1..32   -> 8 CUDA Graph buckets
buckets 40..256 -> 28 exact-size eager buckets
```

The base failure surfaced during the bucket-136 prefill allocation after the preceding Graph captures had reduced free memory to 40.12 MiB. The candidate completed all 36 buckets with a final retained delta of only 16.00 MiB.

### Batch 32/33 boundary

The direct `execute_decode` probe used a 256-token prompt, 10 warm-up rounds, and 100 paired measurement rounds with alternating execution order and matched context progression.

| Metric | Batch 32 Graph | Batch 33 eager | Relative change |
|---|---:|---:|---:|
| Counter delta | 0 | 253 | Confirms Graph / eager dispatch |
| Mean Decode-step latency | 445.0217 ms | 462.6259 ms | **+3.956%** |
| p50 Decode-step latency | 445.1284 ms | 462.3701 ms | **+3.873%** |
| p99 Decode-step latency | 450.2924 ms | 465.6003 ms | **+3.400%** |
| Aggregate throughput | 71.91 tok/s | 71.33 tok/s | **-0.799%** |

Across 100 paired rounds, the mean latency delta was `+17.6042 ms` with a 95% CI of `[17.2299, 17.9785] ms`.

### Validation

| Check | Result |
|---|---|
| `cargo fmt --all --check` | PASS |
| `git diff --check` | PASS |
| Qwen3 release library check | PASS |
| Canonical graph-plan unit test | PASS |
| `batch_invariance_decode_gemm_graph` | PASS |
| Full candidate bucket sweep | PASS, 36/36 |
| Boundary performance probe | PASS, 100 paired rounds |

### Conclusion

The base run confirms that the OOM is caused by cumulative PerToken Graph-executable residency, not by one largest Graph. The revision accounts for the complete eight-Graph retained set before KV sizing and prevents additional persistent Graph allocation above bucket 32. All 36 buckets complete successfully, while the measured aggregate throughput change at the 32/33 boundary is `-0.799%`.

PerToken+TP remains outside the supported public contract and is not claimed by this PR.

pegainfer-qwen3/src/unified_forward.rs
RicardoMin: At the time of testing, I only had access to a 22 GB-class NVIDIA L4, so the results above were collected on that GPU with gpu_memory_utilization=0.90. If validation must be performed on a physical 16 GB GPU, I can rerun the same base/candidate tests on a 16 GB GPU when I am back home and provide the results.

RicardoMin:
I've fixed the DCO sign-off issue. The CI workflow is currently awaiting maintainer approval. Thanks!

pegainfer-qwen3/src/batch_decode.rs
```diff
@@ -59,6 +60,81 @@ pub(crate) enum DecodeGraphUse {
...
+
+        match graph_use {
+            DecodeGraphUse::Serve => {
+                log::debug!(
+                    "PerToken decode bucket {bucket} exceeds graph cap \
+                    {PERTOKEN_GRAPH_MAX_BUCKET}; using eager decode"
+                );
+                Ok(None)
+            }
+            DecodeGraphUse::CaptureOnly | DecodeGraphUse::Replay => {
```
FeathBow, line 127: This part is reached by the unconditional TP sweep at bucket 40, after model/KV allocation, worker startup and watchdog creation. Returning `Err` drops `sweep_done_tx`; the watchdog deliberately remains armed and later calls `std::process::abort()`, so even a caller that handles the error can lose the process ten minutes later.

If PerToken+TP is unsupported, please reject it at the start of `from_runtime_with_lora_options()`. This does not require TP-specific cap handling or restoring the deleted TP test.

pegainfer-qwen3/src/unified_forward.rs
```diff
@@ -61,17 +63,41 @@ impl Qwen3Model {
...
+        // residency remains live under the later eager/unified probes. Skip this
+        // for uncompiled-group modles; the unified sample below bounds their KV.
+        // TP stays eager because its ranks profile independently.
         if self.config.decode_group_is_compiled() {
+            let graph_plan = DecodeGraphPlan::new(decode_bufs.policy_at_construction);
+            if self.enable_cuda_graph
+                && self.tensor_parallel.world_size == 1
+                && graph_plan.requires_cumulative_profile()
+            {
+                for graph_rows in graph_plan.retained_buckets() {
```
FeathBow, line 77: This loop reserves only the full-SM `graphs` cache. `SplitConcurrent` captures into the independent `graphs_split` cache; `enable_decode_overlap()` rejects Pin but allows PerToken, and the PerToken GEMM accepts the override stream. One executor can therefore retain two eight-graph sets while profiling reserves one.

Please reject PerToken+overlap at the existing guard or profile both caches. The guard is the smaller fix if this diagnostic combination is unsupported.

pegainfer-qwen3/tests/batch_invariance_decode_gemm_graph.rs
```diff
@@ -124,6 +125,96 @@ fn run_policy(policy: NumericPolicy, model_path: &str) -> Vec<(bool, bool, u64)>
...
+    }
+
+    (after_first, after_second)
+}
+
+/// Check graph replay at bucket 32 and eager execution at bucket 40
+fn assert_pertoken_graph_cap_behavior(model_path: &str) {
+    set_numeric_policy(NumericPolicy::PerToken);
+
+    let mut ex = Qwen3Executor::from_runtime(model_path, true, &[0]).expect("build probe executor");
```
FeathBow, line 192: The test already constructs a PerToken executor in `run_policy()`, drops it, then loads a fourth model for this probe. Please run the 32/33 counter check on the existing PerToken executor. Keep the counter transition; the extra model startup, `LOGPROBS=64`, diagnostic print and helper-name comments add no coverage.

FeathBow (changes requested):
Thanks for the rework. 

Imo, two blockers remain.

PerToken+TP is described as unsupported but is not rejected at the public low-level constructor. It loads and profiles the ranks, allocates KV, spawns workers and arms the TP watchdog before failing at bucket 40. Returning that error drops the success sender, while the watchdog remains armed and aborts the process at its 600-second deadline. Please reject this combination before model loading and watchdog creation.

Profiling reserves only the full-SM `graphs` cache. Decode overlap captures into the independent `graphs_split` cache, while `enable_decode_overlap()` rejects Pin but still allows PerToken. Please reject PerToken+overlap at that guard or account for both graph sets.

Btw, the remaining documentation and cleanup items are non-blocking, but it would be great if you could address them in the same pass.

RicardoMin:
## Review follow-up

Thank you for identifying the two remaining lifecycle issues. I have updated the PR at commit `8a983192fa8edf4a97338d959663e20158b1eaf0`.

### Changes

- `PerToken + TP` is now rejected at the beginning of `Qwen3Executor::from_runtime_with_lora_options`, before model-path resolution, model loading, KV allocation, worker startup, or TP watchdog creation.
- `PerToken + decode-overlap` is now rejected inside `enable_decode_overlap()` before overlap streams or the independent `graphs_split` cache can be created.
- `DecodeOverlap::Off` remains valid for PerToken, and the existing Tuned overlap behavior is unchanged.
- The batch-32/33 integration probe now reuses the existing PerToken executor instead of loading a fourth model.

This revision intentionally does not support either `PerToken + TP` or `PerToken + decode-overlap`; both combinations now fail closed at their configuration boundaries.

### Validation

| Check | Result |
|---|---:|
| `cargo fmt --all --check` | PASS |
| `git diff --check` | PASS |
| Qwen3 executor unit tests | PASS, 4/4 |
| Direct PerToken+TP constructor probe | PASS |
| `batch_invariance_decode_gemm_graph` | PASS |
| Final-head 36-bucket memory sweep | PASS, 36/36 |
| Final free GPU memory | 2512.12 MiB |
| Retained Graph-memory delta | 16.00 MiB |

The final-head sweep retained CUDA Graphs only for buckets `[1, 2, 4, 8, 16, 20, 24, 32]`; buckets `40..256` completed through the eager PerToken path. This matches the previously reported memory behavior while confirming that the two newly rejected configurations cannot allocate unbudgeted runtime state.

The updated head has been pushed, and CI for this revision is currently pending. Could you please take another look when convenient? Thank you.

pegainfer-qwen3/tests/batch_invariance_decode_gemm_graph.rs
RicardoMin: Thanks for the review. I've addressed all the issues you mentioned in the latest update. Please take another look when you have a chance.

pegainfer-qwen3/src/batch_decode.rs
```diff
@@ -503,3 +579,51 @@ fn grouped_projection_from_packed<'a>(
...
+#[cfg(test)]
+mod tests {
+    use pegainfer_kernels::ops::NumericPolicy;
+
+    use super::DecodeGraphPlan;
+    use super::DecodeGraphUse;
+    use super::PERTOKEN_GRAPH_MAX_BUCKET;
+
+    #[test]
+    fn decode_graph_plan_matches_runtime_and_profile_coverage() {
```
FeathBow: This constructs `DecodeGraphPlan` and asserts values returned by the same helper. The meaningful bucket 32 and 33 serving behavior is already exercised by `batch_invariance_decode_gemm_graph` with the real model and CUDA Graph path. Please remove this unit test and do not replace it with another helper level test.

pegainfer-qwen3/src/executor.rs
```diff
@@ -3044,6 +3084,29 @@ mod tests {
```
FeathBow, line 3128: This only calls `ensure_pertoken_overlap_disabled()` and asserts its result. It does not exercise startup ordering, stream creation, graph allocation, or a supported CLI path. Please remove it. No replacement helper test is needed.

pegainfer-qwen3/src/batch_decode.rs
```diff
@@ -59,6 +60,81 @@ pub(crate) enum DecodeGraphUse {
...
+        }
+
+        let bucket = bucket_for(batch_size);
+        if self.allows_bucket(bucket) {
+            return Ok(Some(bucket));
+        }
+
+        match graph_use {
+            DecodeGraphUse::Serve => {
+                log::debug!(
```
FeathBow: Eager execution above 32 is the intended policy and keeps the PerToken arithmetic, so the behavior itself is fine. The switch is invisible at normal log levels, while this debug message can repeat on every decode step. Please report the graph cap and eager policy once at startup, or expose a stable metric, and keep the hot path quiet.

FeathBow (commented):
Thanks for the update. I do not think another TP unit test is needed here. I suggest removing the two new helper unit tests instead. `decode_graph_plan_matches_runtime_and_profile_coverage` constructs a plan and asserts values returned by that same helper. `pertoken_rejects_every_decode_overlap_mode` calls one guard and asserts its result. Neither test exercises model execution, startup ordering, stream creation, or graph capture. I would not replace them with more helper tests.

The full bucket memory sweep is not in the tree, and the posted receipt does not include the command or diagnostic patch. It would help to provide enough material to reproduce the current head result. This can stay as review evidence and does not need to become another committed test.

The eager switch above 32 is intentional, but it is only reported by a debug message inside the decode loop. Could we report the cap and eager policy once at startup, or expose a stable metric, so this execution mode is visible in normal operation?

It would also be good to update docs description. It still mentions TP support and lists TP2 tests that have since been removed. In the Qwen3 document, please clarify that `--batch-invariant` selects Pin, while PerToken is a diagnostic policy.

RicardoMin:
## Review follow-up and reproducible #780 evidence

@FeathBow Hi FeathBow,

Thank you for the detailed review. I have addressed the remaining cleanup,
observability, and documentation items.

### Changes made

- Removed the two helper-only unit tests:
  - `decode_graph_plan_matches_runtime_and_profile_coverage`
  - `pertoken_rejects_every_decode_overlap_mode`
- Removed the unsupported PerToken+TP implementation and its dedicated test.
- PerToken+TP is now rejected before model loading, KV allocation, worker
  startup, and TP watchdog creation.
- PerToken+decode-overlap is rejected because decode overlap owns an independent
  Graph cache that is not included in the PerToken memory profile.
- Kept the existing `batch_invariance_decode_gemm_graph` integration test and
  reused its PerToken executor for the batch-32 Graph / batch-33 eager check.
- Replaced the repeated decode-loop debug message with a one-time startup
  `info` log describing the retained Graph cap and eager fallback.
- Updated `docs/models/qwen3/decode-attention.md`:
  - `--batch-invariant` selects `NumericPolicy::Pin`.
  - PerToken is not a serving launch mode.
  - PerToken remains available through the diagnostic
    `qwen3_model_report --policy per-token` command.
  - PerToken buckets through 32 use CUDA Graph capture/replay; larger buckets
    keep PerToken arithmetic but execute eagerly.

### Reproducible full-bucket memory sweep

I kept the diagnostic outside the repository because it is review evidence,
not a permanent test. The attached archive contains:

- `run_full_bucket_sweep.sh`
- `pertoken_graph_memory_sweep.rs`
- the generated temporary diagnostic patch
- base/head logs
- environment metadata
- `SHA256SUMS`

Diagnostic source SHA-256:

```text
f8b42238443e8d88a2b0179d9181d3398c2d8b4af9004255bdd4bf19a6b8b1f1
```

Run it from a clean PR worktree rebased onto `upstream/main`:

```bash
git fetch upstream main

export CUDA_HOME=/usr/local/cuda-12.1
export PROTOC=/path/to/protoc
export CARGO_TARGET_DIR=/tmp/pegainfer-target
export PEGAINFER_780_EVIDENCE_ROOT=/tmp/issue-780-evidence

bash /path/to/run_full_bucket_sweep.sh \
  /path/to/pegainfer-pr-worktree \
  <gpu-index-or-uuid> \
  /path/to/Qwen3-4B
```

The runner verifies that the worktree is clean and rebased, temporarily installs
the diagnostic test, runs the same sweep against the exact merge base and PR
head, records metadata and `cuMemGetInfo` samples, and removes the temporary
test afterward.

### Environment

| Item | Value |
|---|---|
| GPU | NVIDIA L4, 23,034 MiB |
| Driver | 550.163.01 |
| CUDA toolkit | 12.1 |
| CUDA target | `sm_89` |
| Model | Qwen3-4B |
| Numeric policy | `PerToken` |
| CUDA Graph | enabled |
| GPU memory utilization | `0.90` |
| Comparison base | `029d3a95a2e00c338e1470ba5049dcbb1fe9c3be` |
| Measured runtime head | `9f6635997e73f43b51f03ee35e7266a44dbc88a6` |
| Current PR head | `94a49438bd2c298c811d0e88c6f232d5bb51ab10` |

`94a49438` only corrects documentation wording; runtime behavior is unchanged
from the measured `9f663599` revision.

### Full-bucket sweep results

| Measurement | Base | Candidate |
|---|---:|---:|
| Completed buckets | 20/36; OOM while entering bucket 136 | 36/36 |
| Retained Graph buckets | Continued growing through bucket 128 | 8: `[1, 2, 4, 8, 16, 20, 24, 32]` |
| Eager buckets | 0 before OOM | 28: `40..256` |
| Retained memory after bucket 32 | 16 MiB | 16 MiB |
| Retained memory after bucket 128 | 2,482 MiB | 16 MiB |
| Final free memory | OOM | 2,512.06 MiB |
| Test result | Expected failure with `CUDA_ERROR_OUT_OF_MEMORY` | Passed |

The base retained a new executable for each larger bucket and exhausted memory.
The candidate stopped retained Graph growth after bucket 32; all remaining
buckets ran eagerly and the complete 36-bucket sweep finished successfully.

### 32/33 execution boundary

A separate 100-round direct-step probe measured:

| Batch | Execution | Latency |
|---:|---|---:|
| 32 | CUDA Graph | 445.02 ms |
| 33 | Eager | 462.63 ms |
| Change | Graph to eager | +3.96% |

### Hardware scope

The available validation device was a 22-GiB-class NVIDIA L4. Therefore, this
is a same-GPU base-fails/head-passes equivalent-pressure result at utilization
`0.90`, not a claim that the original physical 16-GiB device was reproduced.

The attached runner can be executed unchanged on a physical 16-GiB GPU if that
exact hardware receipt is still required.

### Attachment
[issue-780-memory-sweep-evidence-9f663599.tar.gz](https://github.com/user-attachments/files/31369388/issue-780-memory-sweep-evidence-9f663599.tar.gz)

FeathBow:
Thanks for continuing to follow up and address the feedback : ) I reviewed the latest changes again, lgtm now.

PR #862, opened 2026-08-10
feat(qwen35): add opt-in SM120 FlashInfer GDN prefill

### LINKED ISSUE #691: qwen35: migrate GDN prefill to FlashInfer and retire Triton AOT where supported
## Context

OpenInfer pins FlashInfer `v0.6.14` at `19f1a41e6b21f0c422d775e377b6fdf9a1fc9d23`.

Qwen3.5 prefill currently runs a seven-stage Triton-AOT Gated Delta Rule pipeline generated from `openinfer-kernels/tools/triton/gated_delta_rule_chunkwise_kernels.py`. This is the Qwen3.5 model line's only build-time Triton dependency.

The pinned FlashInfer tree now has a fused `chunk_gated_delta_rule` prefill backend for the architectures relevant to current Qwen3.5 work:

- SM100: [flashinfer#3001](https://github.com/flashinfer-ai/flashinfer/pull/3001)
- SM90: [flashinfer#3477](https://github.com/flashinfer-ai/flashinfer/pull/3477)
- SM120: [flashinfer#3479](https://github.com/flashinfer-ai/flashinfer/pull/3479)
- Follow-up prefill work: [flashinfer#3742](https://github.com/flashinfer-ai/flashinfer/pull/3742)

FlashInfer's own benchmark matrix includes Qwen3.5 4B/9B (`Hq=Hk=16`, `Hv=32`), 27B (`Hv=48`), and larger-family head configurations. Its tests cover the 16/16/32 GVA shape used by 4B and 9B.

This makes an upstream-backed migration plausible. It is not a drop-in wrapper.

### 1. The pinned implementation is CuTe DSL, not a reusable C++ launcher

All pinned SM90/SM100/SM120 prefill backends live under `flashinfer/gdn_kernels/` and compile through Python CuTe DSL. GDN is not registered in `flashinfer/aot.py`, and the submodule contains no GDN cubin/fatbin/PTX artifact.

A direct adoption removes Triton but substitutes `nvidia-cutlass-dsl` plus Python artifact generation. Removing both Triton and Python requires one of:

- a reproducible generator plus checked-in or release-packaged per-architecture cubins;
- future upstream `flashinfer-cubin` coverage for GDN;
- another explicit AOT path that exposes a raw-pointer launch contract.

Do not add a Python runtime dependency to the Rust server.

### 2. OpenInfer still needs a small preparation kernel

The current first Triton stage splits fused QKV, expands and L2-normalizes Q/K, and computes the forget/update gates. FlashInfer supports Qwen3.5's GVA mapping directly, so Q/K expansion should be unnecessary, but a native CUDA preparation step is still needed to:

- split the fused projection into Q/K/V views or buffers;
- L2-normalize Q/K;
- convert Qwen3.5's log-space gate to FlashInfer's linear-space forget factor;
- produce fp32 gate and beta tensors.

### 3. Persistent state layout differs

FlashInfer documents final state as `[N,H,V,K]`. OpenInfer's native decode kernel documents and indexes `[H,K,V]`. Since `K=V=128`, shape checks cannot detect a semantic transpose.

Choose one persistent state schema and make prefill plus decode agree. Do not insert a per-layer steady-state transpose. Keeping the native decode kernel with `[H,V,K]` indexing is a valid first scope; migrating decode to FlashInfer is not required unless measurements justify it.

### 4. Architecture support is narrower

FlashInfer GDN prefill supports SM90/SM100/SM120. OpenInfer has historical Qwen3.5 correctness evidence on SM80.

The implementation must make one explicit decision:

- keep the Triton backend as an SM80 fallback;
- provide a non-Triton SM80 fallback;
- or raise the documented Qwen3.5 hardware floor.

With an SM80 Triton fallback, Qwen3.5 can stop using Triton on SM90+ but cannot remove the dependency globally.

## Proposed implementation sequence

1. Build a focused raw-pointer FlashInfer GDN prefill wrapper for one retained architecture first. SM120 is the current Qwen3.5 serving-performance anchor.
2. Add the native CUDA preparation kernel and feed FlashInfer unexpanded GVA Q/K/V.
3. Resolve the persistent state layout without a runtime transpose.
4. Add explicit architecture dispatch and fail closed when the selected artifact/backend is unavailable.
5. Compare peak scratch memory against `GdrChunkwiseScratch35`; delete old scratch only after the new path is proven.
6. Expand to SM90 and SM100, or document why an architecture remains on the existing backend.
7. Remove Triton build code and `OPENINFER_TRITON_PYTHON` only after no supported target depends on it.

## Correctness gates

- Focused output and final-state parity for sequence lengths spanning partial and full 64-token chunks.
- Head configurations at least 16/16/32 and 16/16/48.
- Prefill-to-decode handoff parity with a test that detects the K/V state transpose.
- Qwen3.5 4B, 9B, and 27B short and long HF logits gates.
- Scheduler-level chunked prefill, including resumed `base_pos > 0` execution.
- Existing decode CUDA Graph capture remains valid if the native decode state layout changes.
- Unsupported architectures and missing/mismatched artifacts fail at startup instead of silently falling back.

## Performance and build evidence

No OpenInfer A/B has been measured yet, so this issue makes no performance claim.

Before calling the migration a win, report same-host current-vs-FlashInfer measurements for:

- focused GDN prefill latency across representative prompt lengths;
- end-to-end Qwen3.5 TTFT with the same model, CUDA version, and workload;
- peak GDN scratch memory;
- build time and generated artifact size;
- any decode TPOT change caused by a persistent-state layout rewrite.

Keep upstream-reported improvements separate from OpenInfer measurements.

## Done criteria

- The selected FlashInfer SHA and generated-artifact provenance are recorded.
- The Rust/CUDA runtime has no Python dependency.
- The SM80/SM90/SM100/SM120 support matrix and fallback policy are explicit.
- Correctness gates above pass on every enabled backend.
- Measured A/B data supports keeping the new backend.
- `kernel_plan()`, Qwen3.5 build documentation, and feature descriptions match the final backend.
- Triton files, feature wiring, and environment handling are removed only if the final support matrix no longer needs them.

Related: #479 documented the general FlashInfer bump/JIT boundary; #517 tracked the old value-head specialization. Neither owns this migration.

qwzx-qwas:
## Description

Add an explicitly selected FlashInfer GDN prefill candidate to Qwen3.5 serving. Triton remains the default even when a candidate is linked. The candidate supports SM120, TP1 and Hq/Hk/Hv/D = 16/16/32/128, validated with Qwen3.5-4B.

Backend selection, capability checks and module loading precede KV budgeting. Scratch reservation and dispatch use the selected backend. Unsupported or unavailable candidates fail at load; incompatible bundles fail before linking. Normal prefill has no diagnostic status copy or per-chunk host barrier.

`pegainfer-kernels` owns pinned generation, complete artifact-contract validation and the stable in-place C ABI. Generated symbols and TMA layouts remain outside the model; generation uses an independent checkout without changing the shared FlashInfer headers.

The latest review changes make the boundaries fail closed:
- Runtime construction rejects null, invalid UTF-8, incorrect-length and non-hexadecimal artifact identities, then stores the validated SHA.
- Candidate acceptance entries explicitly select the candidate, check the actual loaded model's identity and pass that same model to the executor/scheduler.
- HF acceptance rejects unresolved or mismatched model revisions. Negative cases exercise the existing constructor and acceptance entries before their positive scenarios.

Existing HF, continuation and scheduler bodies are reused in private test modules. The public three-argument executor API matches main. Identity fault injection is linked only into its dedicated test executable. Validation features/counters, diagnostic controllers, duplicate scenario bodies, profiling/ABBA scaffolding and generated artifacts are excluded. See the [model decision record](https://github.com/qwzx-qwas/openinfer/blob/1bceb2098e462a141c95a32b0acc7ab4abf47f4f/docs/models/qwen35/flashinfer-gdn-prefill.md), routed from the docs index.

Fixes #691

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Validation

Submitted head: `1bceb2098e462a141c95a32b0acc7ab4abf47f4f`, including main `4043d50a`. Its complete tree is exactly the GPU-tested `c923500e21e7223a08c9845d36c2e3e5fa15626c`; subsequent Gmail/sign-off normalization changed commit metadata, not source trees.

On 2026-09-09, all 19 planned acceptance stages passed with one attempt each:
- RTX 5090, driver 595.84, CUDA 13.1.115, Triton 3.7.1, Rust nightly-2026-07-10; Qwen3.5-4B revision `851bf6e806efd8d0a36b00ddf55e13ccb7b8cd0a`.
- Fresh, independent default-Qwen3, stock-Qwen3.5 and candidate production builds; strict Clippy and actual test-executable linking passed.
- Real artifact-contract mutation coverage, five runtime-identity rejection cases, seven acceptance-prerequisite rejection cases, corrupted-object build rejection, and missing/TP2 candidate startup rejection passed for their expected reasons.
- The current real in-place ABI matched full upstream/patched output and state at zero tolerance. Existing upstream reference tensors were reused for the identical candidate object; oracle generation was not repeated.
- Native-prepare CPU reference, candidate/Triton short HF, candidate long HF, model continuation, scheduler chunked prefill and scheduler lifecycle passed with unchanged tolerances.
- Stock and candidate-linked default HTTP used nonempty text; explicit candidate HTTP consumed 20,000 token IDs. Each generated eight tokens.
- Candidate Shared-SM lifecycle passed with both Auto and Off scheduler policies. Nsight confirms actual cross-stream prefill/decode-graph overlap. Triton Shared-SM also passed.
- Formatting, locked metadata, file scope and DCO were checked. All 45 PR commits use matching Gmail author/committer/sign-off identities. No private files are tracked.

Candidate object SHA-256: `35c0ee54e427b4a13661225ab7804ea93a051f54221414d5a36d7c5f254b8450`.
Returned GPU archive SHA-256: `9f348623c5494c5216123a7c23ad4e7017c4646825da95611e27fdd43cadac18`; all 254 evidence-file hashes were verified locally.

This is SM120/TP1 acceptance; actual two-GPU inference was not rerun. Local artifact hashes establish consistency, not trusted release provenance.

## Known numerical limitation

Two earlier candidate short-HF runs failed at the same padded-decode coordinate: seq 3 / pos 2 selected token 67, whose HF regret was 0.25 nat against the unchanged 0.20 limit. The third run and subsequent direct reruns passed; manual interruption is excluded from the numerical-failure count. Isolated controlled experiments reproduced configuration-dependent failure. The latest passing batch does not resolve it.

No validated production fix is claimed. Algorithm pinning, shared tuner/sampler changes, precision workarounds, tolerance relaxation and diagnostic helpers were not added to obtain passing results. Triton remains the default.

## Performance

Historical matched HTTP A/B uses source `aa9ce155`, RTX 5090 with driver 580.126.09, one linked binary for both backends, Shared-SM and synthetic token-ID prompts. Twelve unprofiled rounds completed 12,288 measured requests with zero failures.

| Input/output/concurrency | Paired candidate throughput gain |
| --- | --- |
| 2048/32/1 | +1.30–1.66% |
| 2048/32/8 | +1.65–2.39% |
| 128/128/4 | +0.64–1.00% |

Tail latency did not improve uniformly: one concurrency-8 pair increased TPOT median by 35% and SSE-chunk-gap p99 from 44.8 to 292.2 ms. SSE gaps are not per-token ITL. Short profiler captures support diagnosis, not the unprofiled throughput claim. Performance A/B was not repeated after main integration; these results do not establish natural-language performance or justify changing the default.

## Checklist

- [x] Self-review, formatting and applicable validation completed.
- [x] Implementation commits follow Commitizen; all PR commits carry matching DCO sign-offs.
- [x] Production/test boundaries and private/generated-file exclusion checked.
- [x] Historical failures and measurement limits retained.
- [ ] Complete CI is green on the submitted head. [Run 34312244615](https://github.com/pegainfer-project/pegainfer/actions/runs/34312244615) is `action_required`, awaiting maintainer approval; no jobs have run. The successful old-head run is not evidence for this head.


xiaguan:
Thanks for exploring this direction. I think keeping the FlashInfer CuTe DSL kernel is reasonable: if we AOT-compile it, the serving runtime can avoid Python, Triton, JIT compilation, and JIT cache management.

The main change I would suggest is to simplify the integration boundary before expanding the candidate further. Today, `compile_sm120.py:86-88` extracts FlashInfer's patched/raw PTX, while `flashinfer_gdn.rs:65-72` freezes the generated entry symbol and naked launch ABI and `flashinfer_gdn.rs:426-459` reconstructs that ABI in Rust. At the same time, `lib.rs:101-102` explicitly keeps production launch APIs Triton-only. This leaves PegaInfer maintaining a private CuTe/PTX/TMA contract without yet achieving the original runtime goal.

Could we reshape the PR around this smaller end-to-end boundary?

1. Let `pegainfer-kernels` own the pinned, reproducible CuTe build, patched artifact, manifest, and a stable generated C ABI.
2. Let `pegainfer-qwen35` call only that wrapper; it should not know the generated CuTe symbol or PTX/TMA argument layout.
3. Wire the supported SM120 case into the real production dispatch, with unsupported shapes/configurations handled explicitly.
4. Validate correctness and report end-to-end A/B measurements on that exact production path before treating it as a performance win.

`pegainfer-kernels/tools/cutedsl/export_glm52_fp8_dsl.py:181` is the closest repository pattern. FlashInfer's SM120 path additionally patches PTX after `cute.compile`, so the export flow must preserve that patched artifact (ideally through an upstream-supported export path) rather than applying `export_to_c` blindly.

This keeps the useful CuTe kernel and the goal of removing Triton from serving, while giving the runtime a boundary we can maintain.

qwzx-qwas:
Thanks, this makes sense. I’ll reshape the PR around the smaller integration boundary you suggested

xiaguan:
Thanks — the new C ABI boundary is a substantial improvement.

One remaining request is scope reduction. The PR is still +6,711 lines across 43 files, and much of that is staged diagnostics and benchmark infrastructure rather than the production integration itself. Please keep this PR focused on the production wrapper/dispatch, one reproducible generation path, and the essential end-to-end correctness/performance checks. The stage7/stage13 diagnostics, ABBA harnesses, and additional research scaffolding can move to follow-up PRs or external artifacts.

That would make the runtime change much easier to review and maintain without losing the useful validation work.

qwzx-qwas:
Following the review feedback, I have completed the integration-boundary refactor and production-path validation for the SM120 FlashInfer GDN prefill candidate.

The review requested that the FlashInfer CuTe kernel remain AOT-compiled while moving all generated CuTe/PTX/TMA details out of `pegainfer-qwen35`. The current integration is now:

```text
FlashInfer/CuTe generation environment
→ patched AOT object with embedded SM120 CUBIN
→ manifest + statically linked CuTe runtime
→ PegaInfer-owned versioned C ABI
→ pegainfer-kernels Rust wrapper
→ pegainfer-qwen35 semantic GDN operation
→ production dispatch
```

`pegainfer-kernels` now owns the pinned source, patch, generator, artifact validation, generated kernel symbol, TMA descriptors, raw launch ABI, native prepare kernel, and static runtime linkage.

`pegainfer-qwen35` only sees the semantic `Qwen35GdnAot` interface, geometry, typed device buffers, workspace, and backend evidence. It no longer constructs or depends on generated CuTe symbols, TMA layouts, PTX, or the raw C launch ABI.

The supported single-GPU SM120 Qwen3.5-4B Hv32 case is now connected to production dispatch, while unsupported capabilities explicitly retain the Triton fallback. Invalid artifact/ABI/hash/geometry combinations fail validation rather than silently executing an incompatible kernel.

### Serving-path validation

The serving path was also tested with an empty environment and with PTX JIT disabled.

Real inference completed successfully using the embedded native CUBIN and statically linked runtime, without requiring Python, Triton JIT, FlashInfer Python, or the CuTe compiler at serving time.

### Cross-machine validation

I validated the refactored code on two RTX 5090 machines:

| Machine | Driver | AOT object |
| --- | --- | --- |
| Machine 1 | 580.126.09 | `d92be4c...efa17` |
| Machine 2 | 580.126.20 | `5ceae281...2cd0c` |

Both machines used CUDA Toolkit 12.8 and the same source commit, FlashInfer submodule, generator, patch, and pinned package versions.

Each machine generated its object reproducibly within that machine, but the object hashes differed across machines. Both objects were exactly 291,336 bytes and passed the complete correctness suite.

This validates the runtime ABI and numerical behavior, while bitwise cross-machine artifact reproducibility remains an open build-boundary issue.

### Correctness

On both machines, the Hv32 production candidate passed:

- stable-ABI alias/separate-state tests;
- all operator lengths `T=1,2,63,64,65,127,128`;
- short- and long-sequence HF fixtures;
- chunked/resumed prefill;
- prefill-to-decode handoff;
- scheduler and slot-reuse coverage;
- CUDA Graph coverage.

Production dispatch selected FlashInfer, and the successful-launch counters increased as expected, confirming that the tests did not silently fall back to Triton.

The production HF short replay also remained stable across machines:

| Machine | Mean | p50 | p99 | Max |
| --- | ---: | ---: | ---: | ---: |
| Machine 1 | 0.0251 | 0.0175 | 0.0962 | 0.1367 |
| Machine 2 | 0.0256 | 0.0168 | 0.0992 | 0.1367 |

The broader HF comparison against the same oracle remains:

| Replay surface | Triton mean / p99 / max | FlashInfer mean / p99 / max |
| --- | --- | --- |
| Short sequential | 0.0268 / 0.0985 / 0.1827 | 0.0256 / 0.0992 / 0.1367 |
| Batched, 5 padded | 0.0268 / 0.0973 / 0.1835 | 0.0273 / 0.1156 / 0.1781 |
| Batched, 3 padded | 0.0285 / 0.1019 / 0.1850 | 0.0280 / 0.0878 / 0.1124 |
| Slot compaction | 0.0267 / 0.1016 / 0.1829 | 0.0256 / 0.1106 / 0.1864 |
| Long 4097/8192 | 0.0206 / 0.0721 / 0.0877 | 0.0199 / 0.0692 / 0.0692 |

FlashInfer therefore has a slightly better overall error distribution, although it is not uniformly better on every tail metric.

The non-production Hv48/T128 result is unchanged: FlashInfer has five localized state-tail violations versus two for Triton. The A/B attribution continues to show that these originate from the upstream FlashInfer kernel rather than the PegaInfer HKV layout patch.

### Performance

Machine 1 completed a 15-case same-production-path ABBA matrix covering:

- `T=63/64/65/128/2048`
- `c1/c4/c8`

FlashInfer won TTFT p50 in 14/15 cases and throughput in 12/15 cases.

The case-level geometric-mean improvements were:

- TTFT p50: **+4.26%**
- Throughput: **+3.94%**
- TPOT p50: **-0.11%** (effectively unchanged)

One case, T2048/c8, regressed by 1.95% in TTFT p50 and 2.59% in throughput. In that profile, the FlashInfer GDN operator was 7.94% slower despite reducing the GDN launch count by 3×.

Machine 2 repeated representative cases:

| Case | TTFT p50 | Throughput | TPOT p50 |
| --- | ---: | ---: | ---: |
| T128/c1 | +3.99% | +0.84% | -0.34% |
| T2048/c1 | +2.48% | +1.65% | -0.39% |
| T2048/c8 | +0.08% | -0.02% | +0.08% |

On Machine 2, the earlier c8 regression did not reproduce: T2048/c8 was effectively at parity.

Nsight Systems measured FlashInfer GDN GPU time as:

- **28.9% lower** at T128/c1;
- **6.0% lower** at T2048/c1;
- **6.3% lower** at T2048/c8.

Scratch allocation remains approximately **48.52 MiB** for FlashInfer versus **184.50 MiB** for Triton, a **73.70% reduction**.

The c8 traces show that per-launch FlashInfer time remains stable, while the current single-sequence operation is repeated once per sequence.

Based on this analysis, the next optimization directions are multi-sequence batching and native-prepare optimization.

The existing Triton capability fallback and Hv48 diagnostic scope remain unchanged while the maintainers decide the final release and distribution policy.

After the optimization work is complete, I will also clean up the PR itself by removing the staged diagnostic tooling, benchmark infrastructure, and other temporary research scaffolding that is not required for the production integration. The goal is to keep the final PR focused and minimal, preserving only the production integration, reproducible generation path, and essential correctness/performance validation needed for long-term maintainability.

qwzx-qwas:
Following the review feedback, I have completed the integration-boundary refactor, scope reduction, native-prepare optimization, and production-path validation for the SM120 FlashInfer GDN prefill candidate.

## Current integration boundary

The integration now follows this path:

```text
FlashInfer/CuTe generation environment
→ patched AOT object containing an SM120 CUBIN
→ validated manifest + statically linked CuTe runtime
→ PegaInfer-owned versioned C ABI
→ pegainfer-kernels Rust wrapper
→ pegainfer-qwen35 semantic GDN operation
→ production dispatch
```

`pegainfer-kernels` owns the pinned source, HKV patch, production-only generator, artifact validation, generated kernel symbol, TMA descriptors, raw launch ABI, native prepare kernel, and static runtime linkage.

`pegainfer-qwen35` only sees the semantic `Qwen35GdnAot` interface, supported geometry, typed device buffers, workspace, and backend evidence. It no longer constructs or depends on generated CuTe symbols, TMA layouts, PTX, or the raw C launch ABI.

The supported single-GPU SM120 Qwen3.5-4B Hv32 case is connected to production dispatch. Unsupported capabilities explicitly retain the Triton fallback. If an eligible FlashInfer configuration selects an invalid, incomplete, or ABI-incompatible artifact, validation fails instead of silently executing another kernel.

The README now documents the complete reproducible local path:

1. Initialize the pinned FlashInfer submodule.
2. Create the isolated Python 3.12.3 / CUDA 13 generation environment.
3. Install the locked dependencies.
4. Generate the production-only bundle.
5. Validate the bundle.
6. Configure the build variables.
7. Link the release binary.

## PR scope reduction

The PR has been reduced from **43 files and approximately 6,711 added lines** to **34 files and approximately 3.7k added lines**.

The stage7/stage13 harnesses, ABBA infrastructure, profiling scripts, generated artifacts, logs, benchmark JSON, and other temporary research scaffolding have been removed.

The remaining scope is limited to:

- the production wrapper and dispatch;
- one reproducible artifact-generation path;
- the stable ABI and manifest contract;
- essential operator and production-path correctness tests;
- the native Hv32 prepare implementation.

## Serving validation

The serving path was tested with an empty environment and again with PTX JIT disabled.

Real inference completed using the embedded native CUBIN and statically linked runtime, without requiring Python, Triton JIT, FlashInfer Python, or the CuTe compiler at serving time.

## Cross-machine validation

The refactored implementation was validated on two RTX 5090 machines:

| Machine | Driver | AOT object |
| --- | --- | --- |
| Machine 1 | 580.126.09 | `d92be4c...efa17` |
| Machine 2 | 580.126.20 | `5ceae281...2cd0c` |

Both used CUDA Toolkit 12.8 and the same source commit, FlashInfer submodule, generator, patch, and pinned package versions.

Generation was bitwise reproducible when repeated on the same machine. The objects differed across machines, although both were exactly **291,336 bytes** and passed the complete correctness suite.

The runtime ABI and numerical behavior are therefore validated, while cross-machine bitwise artifact reproducibility remains an open build/distribution question.

## Correctness

The Hv32 production candidate passed:

- stable-ABI alias/separate-state coverage;
- operator lengths `T=1,2,63,64,65,127,128`;
- short- and long-sequence HF fixtures;
- chunked/resumed prefill;
- prefill-to-decode handoff;
- scheduler and slot-reuse coverage;
- CUDA Graph coverage;
- native-prepare dynamic-length and non-finite-input coverage.

Production tests reported:

```text
selected_backend=flashinfer
```

and successful-launch counters increased as expected, confirming that the tests did not silently fall back to Triton.

The latest final gate on commit `9e70e86a` regenerated the production-only artifact and passed artifact validation, release linking, native-prepare correctness, and production HF short replay:

| Mean | p50 | p99 | Max |
| ---: | ---: | ---: | ---: |
| 0.0258 | 0.0177 | 0.1058 | 0.1367 |

The broader comparison against the same HF oracle remains:

| Replay surface | Triton mean / p99 / max | FlashInfer mean / p99 / max |
| --- | --- | --- |
| Short sequential | 0.0268 / 0.0985 / 0.1827 | 0.0256 / 0.0992 / 0.1367 |
| Batched, 5 padded | 0.0268 / 0.0973 / 0.1835 | 0.0273 / 0.1156 / 0.1781 |
| Batched, 3 padded | 0.0285 / 0.1019 / 0.1850 | 0.0280 / 0.0878 / 0.1124 |
| Slot compaction | 0.0267 / 0.1016 / 0.1829 | 0.0256 / 0.1106 / 0.1864 |
| Long 4097/8192 | 0.0206 / 0.0721 / 0.0877 | 0.0199 / 0.0692 / 0.0692 |

FlashInfer has a lower mean error in **four of five** replay surfaces and a lower maximum error in **four of five**, but a lower p99 in only **two of five**.

It is slightly better overall, but not uniformly better on every tail metric.

The non-production Hv48/T128 investigation remains unchanged: FlashInfer has five localized final-state tail violations versus two for Triton.

Patched-HKV versus upstream-HVK A/B testing produced identical results after layout conversion, attributing these violations to the upstream kernel rather than the PegaInfer HKV patch.

## Performance

Machine 1 completed a **15-case same-production-path ABBA matrix** covering `T=63/64/65/128/2048` and `c1/c4/c8`.

FlashInfer won:

- TTFT p50 in **14/15** cases;
- throughput in **12/15** cases.

The case-level geometric-mean results were:

- **TTFT p50:** 4.26% better;
- **throughput:** 3.94% better;
- **TPOT p50:** 0.11% worse, effectively unchanged.

The T2048/c8 case regressed by **1.95% in TTFT p50** and **2.59% in throughput**.

In that trace, the FlashInfer GDN operator was **7.94% slower** despite reducing the GDN launch count by approximately **3×**.

Machine 2 repeated representative cases:

| Case | TTFT p50 | Throughput | TPOT p50 |
| --- | ---: | ---: | ---: |
| T128/c1 | +3.99% | +0.84% | -0.34% |
| T2048/c1 | +2.48% | +1.65% | -0.39% |
| T2048/c8 | +0.08% | -0.02% | +0.08% |

The earlier c8 regression did not reproduce on Machine 2; T2048/c8 was effectively at parity.

Nsight Systems measured FlashInfer GDN GPU time as:

- **28.9% lower** at T128/c1;
- **6.0% lower** at T2048/c1;
- **6.3% lower** at T2048/c8.

FlashInfer scratch allocation is approximately **48.52 MiB** versus **184.50 MiB** for Triton, a **73.70% reduction**.

The native Hv32 prepare optimization was also validated independently. Relative to the previous production implementation, the ABBA means were:

| Case | TTFT p50 | Throughput |
| --- | ---: | ---: |
| T128/c1 | +0.86% | +0.30% |
| T2048/c1 | +0.89% | +0.62% |
| T2048/c8 | +0.66% | +0.61% |

The optimization therefore provides a small, consistent improvement without changing the selected artifact or production numerical gates.

Multi-sequence batching remains a follow-up and is not part of the current PR.

## Maintainer decisions still needed

The remaining policy and scope decisions are:

1. Whether the current Triton capability fallback should remain long-term or eventually be removed for supported SM120 configurations.
2. Whether official artifacts should be distributed as prebuilt bundles, generated locally by users, or support both paths.
3. Which environment should be the canonical artifact build environment, given the observed cross-machine object-hash difference.
4. Whether the non-production Hv48 generic prepare and diagnostic support should remain in this PR or move to a follow-up. It is not required by the Hv32 production artifact or dispatch.
5. Which reference machine and variance threshold should define the final performance acceptance criterion, particularly for T2048/c8.

The current implementation keeps the conservative Triton capability fallback and retains the Hv48 diagnostic path until these decisions are made.

xiaguan:
The test suite is still broader than the production confidence it buys. I suggest reducing it around five meaningful gates:

- one real AOT bundle exercised through the production `build.rs` / link / load path;
- the native prepare kernel checked against a CPU oracle;
- the production dispatch path checked against the HF golden;
- chunked vs. unchunked recurrent-state continuation;
- the real scheduler/CUDA-graph E2E.

Several micro-tests can be deleted or collapsed without reducing confidence: `production_geometry_is_exact_hv32` compares a constant with the same constant; the token-extent test repeats a function that has no token-extent input; the patch-scope test relies on string counts; and the packaging reproducibility test uses identical fake header/object bytes rather than a real generated artifact. The duplicated source-lock/bundle-index/layout arithmetic tests also mostly validate helper code, not the contract consumed by production `build.rs`. The struct-layout test should check field offsets or use generated bindings; size/alignment alone will not catch field reordering.

Most importantly, the high-value GPU tests are all ignored, while the Python contract tests are not wired into GitHub CI. Current CI therefore does not prove that the SM120 production path can build, link, and launch. Replacing the mocked artifact micro-tests with one real production-boundary integration gate would reduce the diff and increase confidence.

qwzx-qwas:
Thanks, I learned a lot from this review. I’ll simplify the test coverage further and add a real integration gate that exercises the production path end to end, so the tests actually demonstrate that the SM120 path can build, link, launch, and produce correct results.

Thanks again for the thoughtful review and guidance！

xiaguan:
Thanks — this is moving in the right direction. The fake artifact/string-contract tests are gone, the ABI offset checks are now meaningful, and the five-gate runner is a much better validation shape.

Could you do one more reduction pass while keeping those five gates?

- Gate 4 only needs to prove real-model continuation: unchunked vs. chunked recurrent state, last-token logits, and the first decode step. The new synthetic fixture, full CPU GDN recurrence, raw operator launcher, and duplicate comparison/log-softmax helpers overlap with Gates 1-3 (real AOT launch, native-prepare CPU oracle, and HF parity), so they can be removed.
- Please move the remaining large test modules out of the production files (a child `prefill/tests.rs` module can still access private parent items). `prefill.rs` is now 1,161 lines and `recurrent.rs` is 1,484 lines.
- The six CUDA-graph evidence counters currently add unconditional atomics to the normal serving path and expand the public runtime-evidence API only for this test. Please either remove them or put them behind a non-default validation feature used by the gate runner.
- `unsupported_geometry_is_explicit` can be folded into the real capability/load boundary rather than kept as a separate three-branch micro-test.

Please keep the real bundle/ABI gate, native-prepare oracle, HF golden, reduced real-model chunk-continuation gate, scheduler/CUDA-graph E2E, and the fail-closed manual runner.

qwzx-qwas:
Thanks — I really appreciate the review and the concrete suggestions.

This revision was pushed by Codex while I was still planning to do another cleanup pass before uploading it, especially to reduce the overall diff. Looking at the current version together with your feedback, there are clearly still several places where I can simplify the implementation further.

I’ll do another reduction pass with the following changes:

1. Reduce Gate 4 so that it only compares the real model under `128-token unchunked` vs. `64+64 chunked` execution, checking:

   * recurrent state and conv state for all linear layers;
   * last-token logits and token;
   * first-decode logits and token.

   I’ll remove the synthetic fixture, full CPU GDN recurrence, raw operator launcher, and duplicated comparison/log-softmax helpers that overlap with Gates 1–3.

2. Move the large tests currently embedded in `prefill.rs` and `recurrent.rs` into private child `tests.rs` modules. This keeps the tests able to access private parent items without expanding the production API.

3. Put the six CUDA Graph evidence counters and the associated runtime-evidence API behind a non-default `gdn-validation` feature used only by the validation runner, so the normal serving path does not pay for those atomics.

4. Fold `unsupported_geometry_is_explicit` into the real capability/load boundary instead of keeping it as a separate micro-test.

5. Continue reducing the overall diff and keep the production changes as small, clean, and consistent with the existing project structure as possible.

I’ll keep the real bundle/ABI gate, native-prepare CPU oracle, HF golden, the reduced real-model chunk-continuation gate, scheduler/CUDA-graph E2E, and the fail-closed manual runner.

Thanks again for taking the time to review this carefully. I’ve learned a lot from the feedback, especially about keeping validation focused on production confidence rather than adding test infrastructure for its own sake.

qwzx-qwas:
@xiaguan
The SM120 implementation and validation are complete on the final PR head, `21be3e50`.

The production path now connects the generated FlashInfer/CuTe bundle through `build.rs`, static linkage, the versioned C ABI, the kernels-owned Rust wrapper, Qwen3.5 production dispatch, and the real scheduler path. The validated object is `d92be4c9…efa17`.

### Production gates

All five exact-filtered gates ran on an RTX 5090 with driver `580.126.09` and CUDA Toolkit 12.8. Each matched exactly one test and completed with `1 passed / 0 failed / 0 ignored`.

1. **Real AOT bundle and ABI**

   A real generated bundle passed manifest validation, production build/link/load, and an actual SM120 launch. The gate checks all C/Rust ABI field offsets, size and alignment, rejects Hv48 at the real capability/load boundary, and launches `T={1,63,64,65,128}`. At `T=65`, separate-state and in-place aliasing produced bitwise-identical output and final HKV state.

2. **Native prepare CPU oracle**

   Q, K, V, alpha and beta were checked against an independent CPU oracle for `T={1,63,64,65,128,2048}`. All comparisons passed. Representative Q/K/V and gate non-finite inputs were detected; the chunk-owned status remained sticky across layers and was cleared for a new chunk.

3. **Production dispatch against HF golden**

   The gate loads the pinned Qwen3.5-4B model through production `Auto` dispatch:

   - `selected_backend=flashinfer`
   - AOT launches: `0 → 288`
   - 12 sequences, 108 positions and 864 head deltas
   - mean logprob delta: `0.0258` (`≤ 0.06`)
   - p99: `0.1025` (`≤ 0.20`)
   - max: `0.1367`, reported diagnostically

4. **Real-model chunk continuation**

   Independent zero-initialized states executed the same prompt as `T=128` and `T=64+64`. The gate compares every element of all 24 recurrent and convolution states, last-token full-vocabulary logits/token, and the first decode step.

   - maximum recurrent mean/p99: `0.00011414 / 0.00111641`, within `0.00025 / 0.002`
   - maximum convolution-state mean/p99: `0.01032218 / 0.046875`, within `0.015625 / 0.0625`
   - last-token logits mean/p99: `0.01413624 / 0.046875`
   - last token: `220 == 220`, regret `0`
   - first-decode logits mean/p99: `0.01250439 / 0.046875`
   - first decode token: `200 == 200`, regret `0`

5. **Scheduler/CUDA Graph E2E**

   The real scheduler path completed state handoff, FlashInfer launch, CUDA Graph capture and replay, state-slot copy/reuse and slot compaction. The gate requires eager fallback to remain zero; all assertions passed.

The final test layout keeps the large test bodies in private child/integration-test modules. CUDA Graph counters and runtime-evidence APIs exist only under the non-default `gdn-validation` feature and do not enter the default production object, hot path or public API. The standalone unsupported-geometry test is folded into Gate 1, and Gate 4 no longer contains the synthetic fixture, full CPU recurrence, raw operator launcher or duplicated log-softmax/comparison implementation.

The runner separately builds the default production configuration and the validation configuration, and fails closed on missing inputs, model/config mismatch, non-SM120 hardware, invalid bundles, zero exact-filter matches, ignored tests, fallback or any failed gate. This covers all five requested validation boundaries.

### HTTP ABBA

Serving performance was measured separately using:

```text
Triton A1 → FlashInfer B1 → FlashInfer B2 → Triton A2
```

The comparison used one release binary with a temporary, unmerged benchmark-only backend selector. Startup logs proved the selected backend for every leg. The selector, benchmark harness and result files are not part of the PR.

The matrix contained nine workloads, four legs and 32 requests per case: 36 result files and 1,152/1,152 successful requests with exact token counts.

| Workload | Output tok/s Δ | TTFT p50 Δ | TPOT p50 Δ |
| --- | ---: | ---: | ---: |
| 63/32, c1 | -0.11% | +1.80% | +0.23% |
| 64/32, c1 | -0.49% | +3.68% | +0.45% |
| 65/32, c1 | +0.03% | +2.37% | 0.00% |
| 2048/32, c1 | +1.06% | -1.16% | -0.80% |
| 2048/32, c4 | +2.44% | -1.90% | -2.41% |
| 2048/32, c8 | +1.87% | -1.66% | -1.68% |
| 128/128, c1 | +2.08% | -8.54% | -2.06% |
| 128/128, c4 | +5.46% | -3.99% | -4.97% |
| 128/128, c8 | +4.63% | -5.48% | -4.68% |

The `63/64/65` cases are effectively at throughput/TPOT parity and show no 64-token cliff. The `2048/32` cases show small repeated improvements, while `128/128 c4/c8` improved throughput and TPOT in both FlashInfer legs.

TTFT p99 was noisier: `64/32 c1` was `+7.39%`, and `128/128 c1` was `+8.19%` due to a single B1 spike that did not repeat in B2. I am therefore not treating either as a stable regression. The c4/c8 improvement direction repeated, but the two FlashInfer samples still showed approximately 7%/6% run-to-run variation, so the exact improvement percentages should not be over-interpreted.

Peak sampled VRAM was effectively unchanged: `25,945 MiB` for Triton and `25,992 MiB` for FlashInfer, a `47 MiB` (`~0.18%`) difference.

qwzx-qwas:
I did another audit and found there are still a few small bits of pollution left in the default build. I’ll go back and clean those up a bit more.

qwzx-qwas:
This pass reduces the boundary-cleanup diff by 409 net lines (`+263/-672`) and addresses the remaining production hygiene and redundant-contract issues:

- GDN build dependencies, `build.rs` logic, native-prepare CUDA source, raw FFI, and safe wrappers are gated by `qwen35`. The default kernels build graph contains only `cc` and `pegainfer-build`; it does not pull in the GDN `serde_json`/`sha2` build dependencies or compile/link the GDN objects.
- Runtime evidence counters, graph/state fields, record hooks, and evidence APIs are isolated behind the non-default `gdn-validation` feature. They are absent from the default production object and hot path.
- Large tests remain in child `tests.rs` modules or integration tests instead of production source files.
- Artifact generation now produces one candidate with fixed artifact basenames. The bundle index, multi-variant plumbing, string-count contracts, self-declared production/serving fields, manifest-provided artifact paths, and redundant source preparation were removed.
- `compile_sm120.py` now returns only actual source/toolchain/PTX/AOT results. Geometry, dtype, token, and workspace constants remain owned by the parent production contract.
- `build.rs` independently verifies the candidate schema, target, geometry, dynamic-token contract, ABI/state layout, workspace requirements, and the hash and size of all three linked artifacts.
- Native prepare is limited to the actual Hv32 production specialization. The unused Hv48/generic native path and dynamic geometry FFI were removed.
- The stable C ABI was reduced from 128 to 112 bytes. The duplicated spec/supported API, fixed geometry launch fields, and runtime artifact-size API were removed. Field-offset guards remain, workspace sizing is cached at handle creation, and integer narrowing is explicitly bounded.
- Supported SM120/Hv32 loads remain fail-closed when the candidate is missing or invalid; unsupported SM, geometry, or TP configurations continue to use the existing fallback policy.

## Validation

All five required gates passed on SM120 for this exact commit and the newly generated candidate:

1. **Real AOT candidate through the production path**
   - `build.rs` → C wrapper → link → load → launch
   - ABI offsets
   - Dynamic `T={1,63,64,65,128}`
   - Non-symmetric HKV state
   - Alias/separate-state parity
   - Unsupported-Hv48 rejection

2. **Native prepare against the CPU oracle**
   - `T={1,63,64,65,128,2048}`
   - Q/K normalization
   - V bit parity
   - Alpha/beta
   - Representative non-finite inputs
   - Sticky status
   - New-chunk clearing

3. **Production Auto dispatch against the pinned Qwen3.5-4B HF golden**
   - FlashInfer selection
   - HF mean/p99/regret checks
   - Greedy-token parity
   - Successful-launch evidence

4. **Real-model continuation**
   - `T=128` unchunked vs. `64+64` chunked
   - Every linear layer's recurrent and convolution state
   - Last-token logits and token
   - First-decode logits and token

5. **Real scheduler/CUDA-graph E2E**
   - Graph capture
   - Graph replay
   - Zero eager fallback
   - State-slot copy/reuse/compaction
   - Successful FlashInfer launches

xiaguan:
I approved and ran the full CI workflow for the current head. It is not merge-ready yet:

- DCO fails because 24 commits are missing `Signed-off-by` trailers.
- Qwen3 CUDA Clippy fails on the unused `CUresult` import in `pegainfer-kernels/src/ffi/qwen35.rs:6`.
- Qwen3.5 CUDA Clippy fails on `format_collect` (`pegainfer-kernels/build.rs:54`), a needless lifetime (`:61`), and a redundant clone (`:183`).

Both CUDA compile jobs and the CPU/format/metadata/sim-E2E jobs pass. Please fix the four Clippy findings, sign the commits, and rerun the complete workflow; once it is fully green, we can merge.

pegainfer-kernels/src/ops/qwen35.rs
```diff
@@ -0,0 +1,267 @@
...
+        let (major, minor) = ctx.ctx.compute_capability()?;
+        let sm = major * 10 + minor;
+        if qwen35_gdn_capability(sm, geometry) != Qwen35GdnSupport::Supported {
+            return Ok(None);
+        }
+        ensure!(
+            unsafe { ffi::pegainfer_qwen35_gdn_abi_version() } == QWEN35_GDN_ABI_VERSION,
+            "Qwen3.5 GDN stable C ABI version mismatch"
+        );
+        ensure!(
```
FeathBow, line 81: This makes the bundle a startup requirement. A normal Qwen3.5 build links only the unavailable stub, but every SM120/Hv32/TP1 model load reaches this check and fails. Neither the repository nor the v0.1.1 release provides the bundle, while the README describes it as optional.

Please either ship a trusted bundle through the release build or select FlashInfer only when it is explicitly enabled. Without that selection, the existing Triton path should remain unchanged.

pegainfer-qwen35/src/prefill.rs
```diff
@@ -205,6 +228,10 @@ impl Qwen35Model {
             )?;
         }
 
+        if let GdnPrefillChunkScratch::FlashInfer(resources) = &gdn_scratch {
```
FeathBow: This check performs a status D2H followed by `ctx.sync()` after all linear layers. `launch_async_prefill()` cannot record its event or continue decode until this returns, so most Shared-SM overlap is lost. A batch also repeats the barrier for each request chunk.

Please remove the finite-status path from normal serving and remeasure FlashInfer with Shared-SM enabled.

pegainfer-qwen35/tools/run_gdn_production_gates.sh
```diff
@@ -0,0 +1,188 @@
...
+  gate3-production-hf-golden \
+  production_flashinfer_gdn_matches_hf_short_golden \
+  -p pegainfer-qwen35 --features qwen35,gdn-validation --test hf_golden_gate
+
+run_exact_gate \
+  gate4-chunk-continuation \
+  prefill::tests::flashinfer_gdn_chunk_continuation_and_model_outputs_match \
+  -p pegainfer-qwen35 --features qwen35,gdn-validation --lib
+
+run_exact_gate \
```
FeathBow: Gate 5 starts the default overlap-off scheduler, so these five gates cannot catch the stream barrier above. The separate Shared-SM E2E checks correctness and lifecycle, not GPU overlap, and this runner does not execute it.

The state checks also compare only the same patched artifact. The reported upstream-HVK versus patched-HKV parity is not reproducible from this runner. Please retain exact-head evidence for FlashInfer + Shared-SM overlap and for the upstream-to-patched layout conversion.

pegainfer-qwen35/src/weights.rs
```diff
@@ -535,8 +540,40 @@ impl Qwen35Model {
             num_pages,
         )?;
 
+        // The first production specialization is deliberately single-GPU.
+        // TP remains an explicit capability fallback to the existing Triton path.
+        let flashinfer_gdn = if geometry.world_size() == 1 {
```
FeathBow: The KV pool has already been sized here using `GdrChunkwiseScratch35::estimate_bytes()`. FlashInfer therefore still loses KV pages to the larger Triton reserve and can fail the earlier memory check even when its actual scratch would fit.

Please reserve the peak memory of the selected backend.

pegainfer-kernels/build.rs
```diff
@@ -39,6 +42,177 @@ struct FlashInferIncludes {
...
+
+    if let Some(bundle) = std::env::var_os(QWEN35_GDN_AOT_ENV) {
+        let bundle = PathBuf::from(bundle);
+        let manifest_path = bundle.join("manifest.json");
+        let manifest_bytes = fs::read(&manifest_path).unwrap_or_else(|error| {
+            panic!("read GDN AOT manifest {}: {error}", manifest_path.display())
+        });
+        let manifest: serde_json::Value = serde_json::from_slice(&manifest_bytes)
+            .unwrap_or_else(|error| panic!("parse GDN AOT manifest: {error}"));
+        assert_eq!(json_u64(&manifest, &["schema_version"]), 3);
```
FeathBow: This block does not bind the source lock, toolchain, dtypes, or Q/K/V/output views. The file hashes are also read from the same supplied manifest. Normal `cargo build` does not run the fuller Python validator, and that validator does not check the view fields either.

A self-consistent but incompatible bundle can therefore link with the wrong tensor contract. Please validate the complete frozen production contract before linking it.

pegainfer-qwen35/Cargo.toml
```diff
@@ -50,10 +51,6 @@ required-features = ["qwen35"]
 name = "sampling_behavior"
 required-features = ["qwen35"]
 
-[[test]]
```
FeathBow, line 53: This test forces resumed prefill through `EngineHandle` and a small scheduler budget. The new test calls private `prefill_chunk_forward()` directly, so it does not cover scheduler admission, request events, or the real scheduler chunk boundary and is not a replacement.

pegainfer-qwen35/Cargo.toml
```diff
@@ -33,6 +33,7 @@ vllm-text = { workspace = true }
 
 [features]
 default = []
+gdn-validation = ["qwen35"]
```
FeathBow: This feature adds counters and hooks across the model, graph, executor, and scheduler, plus a public test-only constructor and duplicate FlashInfer-only HF/E2E tests. The existing HF and scheduler gates already use the normal model loader and exercise the selected backend.

Please remove this feature and keep backend identity at the production startup boundary.

pegainfer-kernels/src/ops/qwen35/tests.rs
```diff
@@ -0,0 +1,228 @@
...
+        workspace_bytes: 80, cu_seqlens: 88, tokens: 96, stream: 104,
+    });
+}
+
+#[test]
+fn stable_c_struct_layout_is_frozen() {
+    assert_stable_c_struct_layout();
+}
+
+#[test]
```
FeathBow: Production always launches in place, while this test adds a separate-state wrapper only to compare two modes of the same upstream artifact. That tests upstream aliasing and expands the project ABI without a production caller. please remove upstream tests.

pegainfer-qwen35/src/prefill/tests.rs
```diff
@@ -0,0 +1,294 @@
...
+
+#[cfg(feature = "gdn-validation")]
+use super::GdnPrefillBackend;
+use super::checked_prefill_end_pos;
+#[cfg(feature = "gdn-validation")]
+use crate::recurrent_state::RecurrentState;
+#[cfg(feature = "gdn-validation")]
+use crate::weights::Qwen35Model;
+
+// Stage 18 measured the real-model FP32 recurrent-state partition floor at
```
FeathBow: The Stage 18/19 chronology is tuning history rather than a code invariant. This file also adds three tests for an unchanged private position helper; the public scheduler E2E already covers context rejection, and the `usize::MAX` case is unreachable from the request boundary.

Please remove the history and those unrelated helper tests. Keep the recurrent-state and logits continuation checks.

FeathBow (changes requested):
Thanks for the thorough work on the AOT boundary. I am requesting changes because the current head changes production behaviour in ways the submitted gates do not cover.

On SM120/Hv32/TP1, a normal Qwen3.5 build links no GDN bundle but model loading still requires one, so the documented clean build cannot start. The FlashInfer prefill path also performs a stream-wide host synchronization after each request chunk, which removes most of the existing Shared-SM overlap. The five-gate runner uses overlap-off and does not retain the reported upstream-HVK versus patched-HKV differential.

Please make the candidate explicitly selectable unless a trusted bundle is distributed, remove the host barrier from normal serving, and validate those two boundaries on the exact production path. The backend also needs to be selected before KV budgeting, and the build must enforce the complete artifact contract.

The patch can be smaller: keep the project-owned ABI, native-prepare oracle, continuation check, and existing HF/scheduler E2E; remove the validation-only counters and public API, duplicate backend-specific gates, upstream alias/separate test path, dead helper, and process-history comments. Please also restore the scheduler-level chunked-prefill E2E, which the new private model test does not replace.

qwzx-qwas:
I really appreciate the careful review and the detailed code-level analysis. The issues you identified are clear, and I agree that the current gates do not fully cover the production behavior changed by this PR.

I’ll address them as follows:

- Make the locally generated FlashInfer candidate an explicit opt-in. Without a selected bundle, `SM120/Hv32/TP1` will retain the existing Triton path; an explicitly supplied but invalid bundle will fail closed.

- Resolve the backend before KV-cache budgeting and reserve the peak scratch/workspace required by the selected backend.

- Remove the complete finite-status path from normal serving, including the device status buffer, atomics, D2H copy, and `ctx.sync()`, so asynchronous Shared-SM prefill/decode overlap is preserved.

- Extend `build.rs` to validate the full frozen production contract before linking, including source/toolchain identity, dtypes, tensor views and strides, ABI/layout, workspace, and artifact metadata.

- Keep the project-owned in-place C ABI, native-prepare CPU oracle, real-model continuation check, and existing HF/scheduler E2Es. I’ll remove `gdn-validation`, validation-only counters and public APIs, duplicate backend-specific gates, the separate-state test ABI, dead helpers, unrelated micro-tests, and process-history comments.

- Restore the scheduler-level chunked-prefill E2E, since the private continuation test does not cover scheduler admission, request events, or the real scheduler chunk boundary.

- Revalidate the exact production path with FlashInfer explicitly selected and Shared-SM enabled, including functional scheduler coverage and GPU timeline/performance evidence. I’ll also retain exact-head upstream-HVK versus patched-HKV differential evidence at the generation boundary without introducing another production backend or test-only runtime API.

Thank you again for taking the time to review this so thoroughly！

qwzx-qwas:
This PR integrates an opt-in FlashInfer GDN prefill backend into the actual Qwen3.5 serving path. Triton remains the default, including when a candidate bundle is linked. The candidate supports SM120, TP1 and Hq/Hk/Hv/D = 16/16/32/128, validated with Qwen3.5-4B.

### Changes following review

- **Kernel ownership and artifact contract:** `pegainfer-kernels` owns pinned generation, artifact validation and the stable in-place C ABI. Generated symbols and TMA layouts stay outside the model layer. Build-time validation checks the frozen source/toolchain, tensor views, dtypes, ABI and final artifact bytes before linking.
- **Production dispatch and memory budgeting:** The selected backend is checked and loaded before KV allocation, and its peak prefill scratch determines the reservation. Unsupported or unavailable candidates fail explicitly.
- **Shared-SM execution:** Removed the per-chunk diagnostic status-copy/host-synchronization path while preserving stream/event ordering and resource lifetimes.
- **Production/test isolation:** Removed validation-only features, counters, the separate-state test ABI and duplicate backend-specific gates. Restored the original three-argument executor API; existing HF backend setup now lives in a private `cfg(test)` module.
- **Scope reduction:** Retained the existing HF, real-model continuation, scheduler-level chunked-prefill and CUDA Graph gates. Shared FlashInfer headers retain main’s pin; generation uses an independent checkout.
- **Documentation:** Added the concise model integration record and its `docs/index.md` route. Rental scripts, profiling tools, private reports and generated artifacts remain outside the PR.

### Validation

The final runtime/test source was validated on RTX 5090 with Qwen3.5-4B: release build, affected-package all-targets Clippy, short/long HF, Shared-SM lifecycle, and real default/explicit-candidate HTTP requests. The candidate HTTP check consumed 20,000 input tokens and generated eight output tokens.

Existing real-bundle, native-prepare oracle, upstream/patched layout, continuation and positive-overlap evidence retains its original source identities.

Earlier same-binary, unprofiled ABBA on `aa9ce155` completed 12,288 requests with zero failures. Paired throughput gains were 0.64–2.39%, with workload-dependent tail-latency regressions. These measurements do not justify changing the default backend.

xiaguan (changes requested):
Thank you for the substantial cleanup and for retaining the measured evidence. The direct-parent Shared-SM A/B completed 12,288/12,288 requests and reports +0.64% to +2.39% paired throughput on the retained workloads, while also documenting the tail-latency regression. My remaining concerns are about making the candidate identity and acceptance gates fail closed.

1. `Qwen35GdnAot::artifact_sha256()` currently converts a null C pointer to `"unavailable"` and invalid UTF-8 to `"invalid-utf8"`. `load_for_production()` does not validate the queried identity, so an explicitly selected candidate can finish loading and serve with only a sentinel in the startup log. Please make this query fallible, validate the 64-character hexadecimal SHA during construction, and store the validated identity in `Qwen35GdnAot`. A broken artifact-identity ABI should reject model loading.

2. The candidate acceptance tests do not encode candidate selection themselves. `tests/common::gdn_backend()` silently defaults to Triton when `PEGAINFER_TEST_QWEN35_GDN_BACKEND` is absent, so the HF, scheduler, and Shared-SM exact tests can pass without loading the candidate. The canonical shell runner sets the variable and checks the startup hash, which is helpful, but direct execution of the named tests is still a false positive for candidate coverage. Please give the candidate acceptance path an explicit candidate requirement and assert the validated artifact identity from the loaded model/engine.

3. The HF provenance check has another silent-success path: when `model_revision()` cannot resolve a revision, `check_fixture_metadata()` returns `false` and four tests return normally, which Cargo reports as passed. The runner's log scan catches this in the canonical script, but the tests themselves remain green without validating the pinned model artifact. Please make an unresolved revision an error for these acceptance tests.

The real-layout gate's exact SHA comparison does reject the two sentinel strings, so this is not a claim that that specific gate is bypassed. The issue is that the runtime constructor and the other directly runnable acceptance tests do not preserve the same invariant.

qwzx-qwas:
Thank you for the careful review, particularly for identifying the silent-success paths in artifact identity and candidate acceptance.

Following your latest review, I have added three changes:

- The runtime constructor now rejects null, invalid UTF-8, incorrectly sized, and non-hexadecimal artifact identities, and stores only the validated SHA.
- Candidate acceptance entries explicitly select the candidate and verify the identity of the actual model passed to the executor or scheduler.
- HF acceptance now fails when the model revision cannot be resolved or does not match the fixture.

The negative cases exercise the existing constructor and acceptance entries before the positive scenarios. Test setup remains private, and identity fault injection is confined to a single test executable. No production injection hook or additional public testing API was introduced.

During validation, two candidate short-HF runs failed at the same padded-decode position: seq 3 / pos 2 selected token 67, whose HF regret is 0.25 nat against the unchanged 0.20 limit. The third run and subsequent direct reruns passed. The latest main-integrated tree also passed all 19 acceptance stages, including short/long HF, real build/load/HTTP, layout and prepare checks, continuation, scheduler lifecycle, and actual Shared-SM overlap. These passes do not resolve the earlier failure.

The isolated investigations narrowed the issue. Fixed GEMM configurations reproduced the candidate failure with the two sensitive projection split-K settings at 9/9, while Triton at 9/9 and the candidate at 16/16 passed. Real-input GDN replay found prefill-state differences. With identical decode-entry state, changing the controlled GEMM configuration still changed the outcome; the first captured differences appeared in the shared decode b/a projections. Both backends use the same decode operators.

My strongest suspicion is an interaction between prefill-state differences, shared split-K rounding, and final BF16 logit materialization. For the failing hidden vector, FP64 projection gives logits of 11.14099662 and 11.11763134 for tokens 692 and 67, respectively. Both round to BF16 11.125, matching the GPU result, and the sampler selects 67. The observed GEMM recipes also store split-K partial results in BF16 despite requesting FP32 computation. This is a concrete precision lead, but its individual contribution has not been isolated, and a unique FlashInfer kernel defect has not been established.

No production fix has been implemented for this numerical issue. I have not hardcoded the passing split-K configuration, changed tie-breaking, relaxed the HF threshold, or promoted diagnostic helpers into this PR.

Triton remains the default backend.

xiaguan (changes requested):
Thank you for addressing the previous review. At `1bceb209`, constructor identity validation, explicit candidate selection on the actual tested model, and rejection of unknown model revisions are all fixed.

The reported historical A/B at `aa9ce155` measured +0.64–2.39% throughput, with a +35% median TPOT regression in one c8 pair and SSE-gap p99 increasing from 44.8 to 292.2 ms. These are synthetic-prompt results from an earlier tree, not current-head performance evidence; SSE gaps are not token ITL. The PR appropriately limits those claims.

I am requesting changes for the unresolved numerical failure and the remaining implementation/test-harness complexity.

1. **Resolve the known HF acceptance failure before production integration.** The report still records two short-HF failures at seq3/pos2: token 67, regret 0.25 nat against the unchanged 0.20 limit. Later passing runs, including the reported 19 successful stages, do not resolve a failure that configuration experiments can reproduce. Please retain the failing workload/configuration, compare default and candidate outputs under matched conditions, and establish the cause and resolution under the existing correctness threshold. The evidence does not yet isolate a defect in the FlashInfer kernel: candidate prefill-state differences and shared decode rounding remain possible contributors. After the correction and scope reduction, please provide fresh matched A/B evidence for the resulting tree before making a production performance claim.

2. **Make artifact identity valid by construction and remove the resulting machinery.** `pegainfer-kernels/build.rs` already validates the bundle bytes and generates the hash literal returned by the C shim. Consider emitting a private typed Rust identity from those same validated bytes, retaining independent build-time hash/ABI validation and the actual-model-versus-expected identity check. This would eliminate the C-string identity interface, repeated pointer/UTF-8/hex handling, `tests/gdn_identity.rs`, and its special GNU `--wrap` build/runner path. The current validation correctly fixes my earlier concern; this is a request to simplify the representation, not to remove validation while leaving the unsafe string boundary intact.

3. **Reduce the gate runner and intermediate artifact protocol.** In `run_gdn_production_gates.sh`, the seven negative HF invocations and Rust-panic/abort-text parsing add substantial maintenance cost for prerequisite checks. Please have the same fixture/revision/identity prerequisite functions return errors that Rust tests can assert directly. Retain subprocess checks where the process/link boundary or CUDA isolation actually matters, real HTTP smoke, provenance, and protection against accidentally running zero tests. Likewise, consider having the selected pinned compiler/generator produce the final bundle directly, removing the `compile-metadata.json` handoff and its parser. Keep the final manifest, hashes of the bytes actually linked, and independent Rust pre-link validation.

4. **Trim redundant comments, pass-through code, and duplicate tests.** Please remove comments that merely narrate visible control flow or repeat implementation history, and simplify the launch-options unpack/repack chain. Keep explanations of ABI layout, ownership/lifetime, and numerical constraints. Please also remove tests whose only contribution is checking upstream/library behavior or repeating an invariant already exercised by the retained E2E tests; avoid replacing the removed harness with a new framework. The project's actual ABI/in-place-state oracle, native-prepare numerical reference, unchanged HF gates, model continuation, scheduler/SharedSm coverage, startup selection, and corrupt-bundle rejection provide useful protection and should remain.

Public checks are currently green, including DCO. My local format/diff checks, runner syntax check, and strict build-crate Clippy check passed. I have not independently rerun the GPU gates or A/B benchmarks.

xiaguan:
Oh, thanks for testing all these options—let’s go with down, get the accuracy checks passing, and wrap this up.

qwzx-qwas:
```markdown
Thank you for the careful review and for insisting on both a simpler implementation and reproducible correctness evidence. The changes are now in `2a33accc`; its source tree is identical to the GPU-tested tree `d1cbc41b`.

The review requests were addressed as follows:

- Artifact identity is now generated from the validated final link inputs and stored as a private Rust value. Independent build-time hash and ABI validation remain in place. The previous C-string identity interface, sentinel handling, dedicated identity fault-injection test, and GNU `--wrap` path have been removed.
- Candidate acceptance entries explicitly select the candidate and verify the identity of the actual loaded model. Missing prerequisites and unresolved HF model revisions return errors directly. The existing HF gate asserts these errors instead of launching seven negative HF subprocesses and parsing panic/abort output.
- The pinned compiler now produces the final bundle directly. The `compile-metadata.json` handoff has been removed, while the final manifest, artifact hashes, and independent Rust pre-link validation remain.
- Existing HF, continuation, and scheduler scenarios are reused through private test modules. The original public executor API is preserved. Diagnostic controllers, profiling scripts, and generated artifacts remain outside the PR.
- Backend loading now precedes KV budgeting; scratch reservation follows the selected backend; and normal prefill no longer contains a diagnostic host barrier. Triton remains the default backend.

The numerical investigation also showed that **improving the down projection alone was insufficient**. Although it corrected the previously failing configuration, subsequent repeated tests exposed a failure under another retained configuration.

The failure remained at `seq3/pos2`:

token `67` had an HF regret of `0.25 nat`, exceeding the unchanged `0.20 nat` limit.

After comparing the different combinations, I selected an adjustment to **the down projection at layer 0 together with the full-attention V projection at layer 3**, applied only to the candidate's tested Qwen3.5-4B decode bucket `N=8`.

Both calls use cuBLASLt's existing compute-type reduction:

split-K partial intermediate results, as well as the reduction across those partials, retain FP32 precision, while inputs, weights, and final outputs remain BF16.

The model owns copies of the tuned algorithms, and only the reduction setting is modified. A single private target list and a shared preparation/execution path manage both projections.

Other projections, Triton, and the shared tuning cache are unaffected.

The adjustment reuses the existing decode workspace and does not allocate an additional GPU buffer.

For the final source tree, both retained a/b projection configurations—split9 and split16—were tested with both backends. Each configuration/backend combination was executed in three independent process launches:

**12/12 short-HF runs passed, including all six candidate runs.**

Each run covered:

- sequential decode;
- both padded batch cases;
- slot compaction.

The correctness thresholds remained unchanged:

- Mean head-logprob error: `≤ 0.06 nat`
- p99 head-logprob error: `≤ 0.20 nat`
- Chosen-token regret: `≤ 0.20 nat`

| Retained a/b split | Backend | Runs passed | Worst phase mean error (nat) | Worst phase p99 error (nat) | Largest head-logprob error (nat) |
| --- | --- | ---: | ---: | ---: | ---: |
| 9 | Triton | 3/3 | 0.0277 | 0.1107 | 0.1907 |
| 9 | Candidate + down/V | 3/3 | 0.0278 | 0.1131 | 0.1845 |
| 16 | Triton | 3/3 | 0.0276 | 0.1283 | 0.1907 |
| 16 | Candidate + down/V | 3/3 | 0.0278 | 0.1131 | 0.1820 |

These values are the maxima across the three runs and four HF scenarios.

The conclusion is therefore:

**The previously retained failing cases now satisfy the original acceptance criteria.**

This is not a claim that every numerical metric improved.

The final source tree also passed the complete five-category production gate runner, including:

- release builds and strict Clippy;
- real build/link/load/HTTP paths;
- rejection of corrupt and unavailable candidates;
- upstream-to-patched in-place state-layout parity;
- the native-prepare CPU reference;
- short/long HF;
- model continuation and scheduler continuation;
- scheduler/CUDA Graph lifecycle coverage.

Nsight also confirmed actual cross-stream prefill/decode overlap under both Auto and Off scheduler policies.

For the fresh performance measurements, Triton and candidate + down/V used the exact same final serving binary.

The test configuration was:

- Shared-SM enabled;
- concurrency = 8;
- synthetic token-ID prompts;
- ABBA ordering.

Across two retained algorithm contexts, three workloads, and two seeds, a total of:

**48 service runs completed 12,288/12,288 measured requests with zero failures.**

| Retained a/b split | Input/output tokens | Triton output throughput (tokens/s) | Candidate + down/V (tokens/s) | Change |
| --- | --- | ---: | ---: | ---: |
| 9 | 128/128 | 1012.201 | 1016.685 | +0.44% |
| 9 | 2048/32 | 262.307 | 271.858 | +3.64% |
| 9 | 8192/32 | 77.216 | 80.603 | +4.39% |
| 16 | 128/128 | 1012.008 | 1014.421 | +0.24% |
| 16 | 2048/32 | 262.960 | 270.834 | +2.99% |
| 16 | 8192/32 | 77.253 | 80.658 | +4.41% |

These values are the averages of the two ABBA pairs for each configuration/workload.

Across individual ABBA pairs, throughput changes ranged from:

**+0.14% to +4.42%.**

Latency tradeoffs remain.

For the split9, 2048/32 workload:

average TTFT p50 improved from:

`492.450 ms`

to:

`364.266 ms`

while TPOT p50 increased from:

`15.521 ms/token`

to:

`18.334 ms/token`.

In the worst individual pair:

TPOT p50 increased from:

`13.081 ms/token`

to:

`18.564 ms/token`

or:

`+41.91%`

while E2EL p99 increased from:

`1068.452 ms`

to:

`1129.248 ms`

or:

`+5.69%`.

In addition, these results come from synthetic workloads and therefore do not establish performance under natural-language serving workloads.

I still do not consider this an ideal solution.

The down/V adjustment makes the retained correctness cases satisfy the acceptance criteria, but it remains a **projection-specific numerical mitigation**.

It does not prove the existence of a single unique FlashInfer kernel defect, nor does it establish stability across every configuration.

I hope future work can replace this approach with a solution that has a clearer numerical basis, simpler maintenance, and better latency behavior.
```

Issue #408, opened 2026-06-16
higgs-audio: crate skeleton + config/weights + backbone forward + parity gate


ywh555hhh:
## Summary

Sub-issue for #395. Adds the `openinfer-higgs-audio` crate: config parsing, weight
name mapping, backbone forward (text-only), and a parity gate against golden logits.

See PR #409 for the implementation.

### 395.1 — Crate skeleton + feature wiring
- `openinfer-higgs-audio/Cargo.toml` with `higgs-audio` feature gate
- `src/lib.rs` module declarations
- Workspace registration in root `Cargo.toml`

### 395.2 — Config parsing
- `src/config.rs`: `HiggsConfig` + nested `TextConfig` + `AudioEncoderConfig`
- Nested `rope_parameters.rope_theta = 1_000_000` resolution
- Unit tests assert 11 architecture facts against real checkpoint config

### 395.3 — Weight name mapping
- `src/weights.rs`: `map_backbone()` maps `body.*` names to `BackboneSlot`
- Audio/codec tensors correctly return `None`
- Unit tests verify 398 backbone tensors, all 36 layers with 11 components

### 395.4 — Backbone forward
- `src/backbone.rs` (behind `#[cfg(feature = "higgs-audio")]`)
- Mirrors `openinfer-qwen3-4b` weight loading with `body.*` naming
- Full 36-layer Qwen3 text prefill: RMSNorm → fused QKV → FlashInfer paged attention → SwiGLU MLP
- `forward()`, `last_token_logits()`, `compute_all_position_logits()`

### 395.5 — Backbone parity gate
- `tests/backbone_parity.rs` (`required-features = ["higgs-audio"]`)
- Top-64 logprobs comparison: regret + mean (≤ 0.06 nat) + p99 (≤ 0.20 nat)
- Same pattern as qwen3 golden gate
- Clean skip when model or golden file absent

## Verification

**Mac (completed):**
```
cargo build -p openinfer-higgs-audio    ✅
cargo test --lib (3/3)                  ✅
cargo fmt --all --check                 ✅
git diff main --stat -- <shared crates>  ✅ (empty)
```

**GPU (pending):**
- [ ] `cargo build -p openinfer-higgs-audio --features higgs-audio`
- [ ] Convert `backbone_golden.pt` → `backbone_golden.safetensors`
- [ ] `cargo test --release -p openinfer-higgs-audio --features higgs-audio --test backbone_parity`

## Guardrails
[...]

Issue #203, opened 2026-06-02
[Roadmap]: pegainfer roadmap


xiaguan:
# pegainfer Roadmap

Hi pegainfer developers! pegainfer started as a from-scratch, pure Rust + CUDA inference engine
whose goal was to understand every layer of the stack by building it. We now have several models
serving real requests, a shared kernel layer, and an OpenAI-compatible frontend. The next chapter
is about turning what we proved per-model into infrastructure others can build on — without
collapsing back into a super-abstraction nobody dares touch.

This is the high-level plan. `docs/roadmap/direction.md` (the "one size can't fit all" principle)
and `docs/roadmap/execution.md` (what's actually in flight) stay the day-to-day source of truth;
this sits above them. Items carry a status — **Now** / **Next** / **Later** — and move or get
deleted as reality changes. No dates. File paths below point into the `pegainfer` repo.
(As a doc this lives under `docs/roadmap/`.)

Two contribution modes: load-bearing shared infrastructure (the KV cache crate especially) starts
with a **design issue, not a PR**, so we agree on the interface first; per-model and additive work
takes PRs, ideally after claiming it in the relevant tracking issue. Each model and subsystem
should have a **named owner** — "nobody owns it" is how shared code rots. Owners are all _TBD_;
claim one in its issue.

## Table of contents

1. [Reusable KV cache crate](#1--reusable-kv-cache-crate-the-big-one)
2. [Per-model ownership, roadmaps & new models](#2--per-model-ownership-roadmaps--new-models)
3. [A frontend coding agents can use](#3--a-frontend-coding-agents-can-use)
4. [Serving capability parity](#4--serving-capability-parity-vs-vllm--sglang)
5. [Foundations: a Rust infra others can build on](#5--foundations-a-rust-infra-others-can-build-on)
6. [How to contribute](#how-to-contribute)
7. [Appendix — draft tracking issues](#appendix--draft-tracking-issues)

---

## 1 — Reusable KV cache crate (the big one)

**This is the biggest and least-decided piece. Everything here is a starting point for
discussion, not a settled design.**

KV cache management is where the hard cross-cutting problems live: paged allocation, prefix reuse,
eviction, host (CPU) offloading, and prefill/decode (P/D) disaggregation with KV transfer between
workers. Today it's two parallel paths: Qwen3-4B runs on `pegainfer-kv-cache` (wrapping the
vendored dynamo `kvbm/kvbm-logical`), but only its allocator + registry — **automatic prefix
[...]

PR #864, opened 2026-08-13
feat(higgs-audio): add checkpoint loading and validation foundation

ywh555hhh:
## Summary

This PR adds the first Higgs-Audio integration foundation for PegaInfer.

The scope is intentionally limited to the checkpoint-backed validation path: parse Higgs-Audio config and checkpoint metadata, map Higgs checkpoint tensor names into the existing Qwen3 loader, run Qwen3 text-backbone prefill, run the audio-head projection validation path, and add golden-trace tools for debugging hidden-state drift.

This is not full Higgs-Audio generation yet. Delay-pattern decode, multi-codebook autoregressive audio decode, codec/vocoder integration, and waveform output are left for follow-up PRs.

## Why this shape

Higgs-Audio reuses a Qwen3-style text backbone, but its checkpoint layout is not a drop-in Qwen3 checkpoint. To avoid materializing a renamed multi-GB checkpoint copy, this PR adds a tensor-name alias path in the weight loader. The Qwen3 config view stays small and explicit, while the original Higgs checkpoint remains the weight source.

I have been calling this the first integration layer during bring-up: it proves that checkpoint/config loading, Qwen3 backbone prefill, audio-head projection, and trace-based validation can work before expanding into full audio generation.

The Qwen3 diagnostic prefill hooks added here are for model bring-up and golden-trace validation. They are kept separate from the normal serving scheduler path.

## What changed

- Added `pegainfer-higgs-audio` with Higgs config, manifest, load-plan, and validation utilities.
- Added Qwen3 weight-source loading with tensor-name aliases.
- Added Qwen3 prefill diagnostics for final hidden, per-layer hidden, and selected layer stages.
- Added comparison tools and a small offline fixture for the audio-head validation path.
- Documented scope, validation method, and current limitations in `docs/models/higgs-audio/a-layer-validation.md`.

## Validation

Local checks run:

```bash
cargo fmt --check -p pegainfer-core -p pegainfer-qwen3 -p pegainfer-higgs-audio
cargo check -p pegainfer-higgs-audio --bins
python3 -m py_compile tools/accuracy/*.py tools/higgs/check_higgs_gate_summary.py
```

4090 validation evidence from the checkpoint-backed validation path:

| Metric | Result |
| --- | ---: |
| prompt ids | exact |
| audio argmax ids | exact |
| hidden cosine | 0.999994874 |
| logits cosine | 0.999998987 |
| top64 min overlap | 58 / 64 |
| top64 mean overlap | 61.00 / 64 |
| final hidden bf16 mean abs | 0.006735 |
| audio logits f32 mean abs | 0.040235 |

Strict elementwise parity is still diagnostic-only, but the semantic gate passes.

I also ran an RMSNorm rounding ablation. The HF/Qwen3-style fused-add-RMSNorm variant slightly improves strict trace parity, but it is not required for the current Higgs-Audio one-step semantic gate. To reduce shared-layer scope, this PR no longer changes `pegainfer-kernels`; the rounding question can be discussed separately as a Qwen3 numeric-parity topic if needed.

## Known limitations

- Full audio generation is not included in this PR.
- Diagnostic prefill hooks currently target the single-GPU bring-up path.
- Runtime CUDA validation needs a Linux CUDA machine; macOS local checks stop at CUDA/RDMA build dependencies.

## Maintainer question

Is the tensor-name alias loader an acceptable way to reuse the original Higgs checkpoint layout, instead of materializing a renamed Qwen3-body checkpoint copy?


ywh555hhh:
Small self-review note before detailed maintainer review:

I realized this PR is broader than a pure `pegainfer-higgs-audio` addition. It adds the new Higgs-Audio validation crate/tools/docs, but it also touches shared/Qwen3-facing seams:

- `pegainfer-core`: optional tensor-name aliases in the weight loader.
- `pegainfer-qwen3`: diagnostic prefill surfaces used for golden-trace validation.
- `pegainfer-kernels`: an HF/Qwen3-style fused add RMSNorm rounding variant.

The intent was to reuse the existing Qwen3 backbone and avoid duplicating loading/forward logic inside the Higgs-Audio crate. That said, I recognize these are public/shared-layer changes and may be too much for one PR.

I’m going to double-check whether each shared change is truly necessary and whether this should be split into smaller PRs before asking for full review. If maintainers prefer, I can split it into: Higgs crate/tools/docs, weight-loader alias support, Qwen3 diagnostics, and RMSNorm rounding as separate PRs.

CI is currently passing, but I do not want to hide the scope concern.

ywh555hhh:
I want to discuss one design tradeoff in this PR before asking for deeper review.

This PR no longer strictly matches the original guardrail in #408, where I said the Higgs-Audio bring-up would avoid shared crate modifications. During implementation, I found that validating against the real Qwen3-backed runtime path creates a tradeoff:

- keep `pegainfer-higgs-audio` isolated, but duplicate Qwen3-like loading/prefill logic and risk drifting from the main Qwen3 implementation;
- or reuse the existing Qwen3 backbone path, but add small shared/Qwen3-facing seams such as tensor-name aliases and diagnostic prefill hooks.

I chose the second path in this PR, but I realize this is exactly the kind of shared-layer complexity that maintainers may not want bundled into an initial Higgs-Audio PR.

This also relates to the earlier guidance in #203: Higgs should be one model / one crate, and audio-specific concepts should not leak into shared layers. I think the current question is whether these changes are acceptable as general reusable seams, or whether I should split them out / keep the initial Higgs PR more isolated.

Concretely, I would like maintainer input on the first tradeoff:

Should Higgs-Audio reuse the existing Qwen3 loading/prefill path via small shared seams, or should the initial implementation duplicate/adapt Qwen3-like logic inside `pegainfer-higgs-audio` to keep other model lines untouched?

I can split this PR if the current shape is too broad.

ywh555hhh:
Follow-up after the RMSNorm rounding ablation:

I removed the `pegainfer-kernels` shared rounding change from this PR. The ablation supports that HF/Qwen3-style fused-add-RMSNorm rounding slightly improves strict trace parity, but is not required for the current Higgs-Audio one-step semantic gate.

So the foundation PR is now narrower: it keeps the Higgs crate/tools/docs, tensor-name alias support, and Qwen3 diagnostic prefill surface, but no longer changes shared kernel code. If the rounding behavior is worth pursuing later, I will treat it as a separate Qwen3 numeric-parity discussion rather than coupling it to this Higgs-Audio foundation PR.

xiaguan:
Thanks for pushing this checkpoint-loading path forward. The tensor-name alias idea and the checkpoint-backed one-step parity gate both look useful, and I think there is a good minimal foundation inside this PR.

Would you be open to reducing the scope before we review the implementation in detail? The current diff is about 10k lines across 48 files, and it combines several different concerns:

- the generic weight-loader alias seam;
- the Higgs config/manifest/name mapping and fused audio head;
- multiple Qwen3 runtime-source paths, including a materialized body view;
- public Qwen3 per-layer/per-stage diagnostic execution APIs;
- a fairly large shell/Python readiness and summary framework.

A smaller shape would be much easier to validate and maintain:

1. Land the generic tensor-alias loader seam separately, with one real safetensors load integration test.
2. Keep the Higgs PR focused on config/manifest validation, name mapping, the fused audio head, one narrow final-hidden hook, and one checkpoint-backed one-step golden gate.
3. Move or remove the materialized-body fallback, per-layer/stage tracing, retained prompt session support (until decode uses it), the handwritten kernel-plan metadata, and the isolated/readiness summary scaffolding.

In particular, `run_isolated_higgs_checks.sh` removes the real Qwen/core runtime dependencies and then validates fabricated artifact/summary files. That checks the helper parsers, but it does not increase confidence in the production checkpoint load or CUDA path. I would prefer the smaller set of real boundary tests over this additional framework.

One wording detail: the committed reference path uses a Transformers Qwen3 backbone plus the SGLang tokenizer/head semantics, not the full SGLang serving runtime, and the current Rust path stops after one prefill/audio-head prediction. Describing it as a one-step compatibility foundation would set the right expectation until delay-pattern decode, codec/vocoder, and serving are added.

ywh555hhh:
Thanks for the detailed review. I understand the concern now.

You're right that this PR has grown too broad: it mixes the generic loader alias work, Higgs-specific checkpoint/config validation, Qwen3 diagnostics, and several readiness/debugging helpers. I'm happy to narrow the scope and make the follow-up work easier to review.

I'll convert this PR to draft for now, then split out the generic tensor-name alias loader into a smaller PR with a real safetensors loading test. After that, I'll reduce this PR to the minimal Higgs-Audio foundation: config/manifest validation, name mapping, fused audio head loading, a narrow final-hidden hook, and the checkpoint-backed one-step golden/parity gate.

Thanks again for the clear direction.

PR #877, opened 2026-08-14
feat(gemma4): global decoder layer and its HF golden oracle probes

### LINKED ISSUE #876: gemma4: one global decoder layer, forward only, against the HF golden probes
## Description

The local layer (#875) executes; the global (full-attention) kind — 8 of the 48 layers, at head_dim 512 — does not. Two of the umbrella's open assumptions live exactly there: whether the published weights carry a separate `v_proj` on global layers, and whether K=V is a runtime equivalence or needs shared physical storage. Both answers shape the KV-cache design, so they should come from the checkpoint and the reference, not from reading the paper.

## Proposed Plan

- One global decoder layer, forward only, sharing the local layer's attention epilogue; prefill-from-zero only, like the local kind.
- Resolve K=V from the weights: no `v_proj` on `full_attention` layers, value forked from `k_proj`'s raw output before `k_norm` and RoPE, then weightless-normed.
- Proportional RoPE carried by the tables: the prep kernel runs at full head width, the partiality lives in the table construction; attention unscaled through the hd512 single-prefill's explicit `sm_scale`.
- Parse and validate the rope contract once on the typed config — the declared `rope_type` per family, positive theta, an exact even rotary width; range-check global positions on the host before upload, since the batched prep kernel `__trap()`s instead of returning an error.
- Extend the oracle to the first and last global layer — indices from the layer-map parse, cross-checked against the fixture metadata's own parse — eight comparisons in one checkpoint load, asserting the probe adjacency the fixture's cut dedup relies on.

## Acceptance Criteria

- The layer refuses a global layer that carries a `v_proj`, rejects `start_pos > 0`, and rejects positions beyond the rope tables on the host; malformed rope geometry fails at config parse.
- Controls demonstrate the oracle catches the `rotary_dim = 128` misreading and a V fork placed after the norms.

FeathBow:
## Description

Closes #876

The global (full-attention) decoder layer at head_dim 512, forward only — everything downstream of attention was already kind-agnostic and is now one shared epilogue. The differences from the local kind are the ones the checkpoint and the reference pin:

- **K=V is a runtime fork, not shared storage.** The checkpoint ships no `v_proj` on any `full_attention` layer (the loader enforces it; this layer refuses one that carries it), and the reference forks value off `k_proj`'s raw output BEFORE `k_norm` and RoPE touch K — after the norms the two tensors differ bitwise. 
- **Proportional RoPE is table-borne.** The inverse frequencies keep the full head_dim as exponent denominator, the un-rotated band is zero-padded, and `rotate_half` pairs `(d, d + 256)` across the whole head — so the prep kernel runs at `rotary_dim = head_dim` and the partiality lives in the tables. Passing the intuitive 128 rotates the wrong pairs entirely; a control below shows the oracle catches exactly that. The rope contract is parsed and validated once on the typed config — each family's declared `rope_type` is asserted against the implemented algorithm, theta must be positive, and `partial_rotary_factor * global_head_dim` must land on a positive even width within the head — and the table builder re-checks its inputs at the seam. The global forward range-checks positions on the host before upload, because its batched prep kernel `__trap()`s on out-of-range positions instead of returning an error (the hd256 prefill wrapper rejects them itself).
- Attention stays unscaled (`scaling = 1.0`) through the hd512 single-prefill's explicit `sm_scale`, and the layer is prefill-from-zero only (`start_pos > 0` rejected), like the local kind.

The oracle extends to both global boundaries — indices from the layer-map parse, cross-checked against the fixture metadata's own parse — eight comparisons in one checkpoint load. The last global layer's input is the sliding side's output cut: the fixture dedups adjacent cuts by depth, and that adjacency is asserted rather than assumed.

## Test Env

Single GPU (sm_89, x86_64), CUDA 12.9, against the pinned 12B checkpoint.

## Verification

- Eight comparisons (single/short x sliding/global x first/last boundary) against the declared tolerance of 0.4 absolute + 2% relative:
  - **All four one-token comparisons are bitwise exact (max_abs = 0)** — including both global layers. With softmax over one key, query, RoPE and scale drop out of the result, so this pins the V/O path — the K=V fork and the weightless V norm included — the norms, the MLP and `layer_scalar` exactly; the tables and the scale are gated by the nine-token comparisons and the controls below.
  - The four nine-token comparisons show scattered rounding noise only, zero elements over tolerance, token 0 exact throughout; the global layers sit below the local pair (max_abs 0.125 / 0.0625 vs 0.1875 / 0.25).
- Five negative controls, each against the full checkpoint: swapping the reference cut reddens all eight; `sm_scale` pinned to `rsqrt(512)` fails exactly the two multi-token global comparisons; the V fork moved after the norms fails both multi-token global comparisons while one-token stays in tolerance (the weightless norm cancels `k_norm`'s magnitude until multi-key softmax amplifies the residue — why the multi-token probe exists); **`rotary_dim = 128` — the intuitive reading — fails both multi-token global comparisons**; rebuilding the tables with `rotary_dim` as exponent denominator fails the first global comparison. 

## Type of Change

- [ ] Documentation update

### DIFF
--- a/pegainfer-core/src/ops.rs
+++ b/pegainfer-core/src/ops.rs
@@ -62,6 +62,7 @@ pub use pegainfer_kernels::ops::lora_decode_fused_delta_group3_into;
 pub use pegainfer_kernels::ops::lora_decode_fused_delta_into;
 pub use pegainfer_kernels::ops::pack_lora_b_rows_into;
 pub use pegainfer_kernels::ops::qk_norm_partial_rope_batched_decode_hd256_into;
+pub use pegainfer_kernels::ops::qk_norm_partial_rope_batched_decode_hd512_into;
 #[cfg(not(feature = "kernel-call-trace"))]
 pub use pegainfer_kernels::ops::qk_norm_rope_batch_decode_into;
 pub use pegainfer_kernels::ops::qk_norm_rope_prefill_hd256_plain_into;
@@ -84,6 +85,7 @@ pub use pegainfer_kernels::ops::silu_mul_batch_into;
 pub use pegainfer_kernels::ops::silu_mul_fused_batch_into;
 pub use pegainfer_kernels::ops::single_decode_nhd_into;
 pub use pegainfer_kernels::ops::single_prefill_hd256_into;
+pub use pegainfer_kernels::ops::single_prefill_hd512_into;
 pub use pegainfer_kernels::ops::single_prefill_nhd_causal_into;
 pub use pegainfer_kernels::ops::single_prefill_nhd_noncausal_into;
 pub use pegainfer_kernels::ops::write_vec_into;
--- a/pegainfer-gemma4/src/config.rs
+++ b/pegainfer-gemma4/src/config.rs
@@ -17,17 +17,15 @@ pub(crate) enum LayerKind {
     Global,
 }
 
-/// First and last sliding layers from the layer map; `None` when the map has
-/// no sliding entry.
-#[cfg(test)]
-pub(crate) fn first_last_sliding(layer_types: &[LayerKind]) -> Option<(usize, usize)> {
-    let mut sliding = layer_types
+#[cfg(all(feature = "gemma4", test))]
+pub(crate) fn first_last(layer_types: &[LayerKind], kind: LayerKind) -> Option<(usize, usize)> {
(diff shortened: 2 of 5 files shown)


FeathBow:
@codex review please

PR #879, opened 2026-08-14
feat(gemma4): full 12B forward from one token through window-edge pfill

### LINKED ISSUE #878: gemma4: the full 12B text-tower forward, gated against the HF top-64 logprobs
## Description

Both layer kinds now execute and match their probes (#875, #877), but nothing runs the tower end to end: the scaled embedding, the 48-layer local/global dispatch, the final norm, the tied LM head and the final-logit softcap have never been composed, and the constants that live only at the ends are still unpinned. 

The gate cannot be an elementwise tolerance at every width. At one and nine tokens the gap to the reference sits at the bf16 rounding floor and top-1 is exact. At 1024 tokens — exactly the sliding window, the widest prefill that evicts nothing — even two HF attention backends diverge chaotically: sdpa vs eager on the same weights and the same sequence agree on top-1 at only 802/1024 with max |dlogprob| 16.4, spread across all four quarters. A useful window-edge gate is therefore distributional: a global top-1 floor near that baseline, a chaos ceiling on the logprob gap, and a per-quarter top-1 floor bounding localized damage (a whole-quarter failure cannot hide inside a passing global rate) — with our own determinism pinned separately.

## Proposed Plan

- One `full_forward`: scaled embedding, layer dispatch driven by the parsed layer map, final norm, tied head, softcap from the typed config; geometry derived inside from the weights' config (one source of truth for the buffer sizes handed to CUDA), token ids bounds-checked against the embedding before upload.
- Oracle over the fixture's three cases (one token, nine tokens, 1024 tokens), comparing log-probabilities at the reference's top-64 ids per position; tight elementwise gates plus exact top-1 on the short cases, the distributional gate at the edge, and a bitwise replay assert for our own determinism.
- Derive the embedding scale and assert it against the fixture's recorded bf16 value before it multiplies anything.
- Re-open on the weights only what the forward reads (`config`, `embed_tokens`, `norm`); carry `final_logit_softcapping` on the typed config, validated positive at parse.

## Acceptance Criteria

- The three cases pass their declared gates — global and per-quarter — in one checkpoint load, and non-finite logprobs fail outright.
- The one- and nine-token cases hold tight elementwise gates with exact top-1; the window-edge case holds the distributional gate (global and per-quarter) at the HF-backend-comparison noise floor; a second edge run is bitwise identical (`to_bits` comparison).
- The embedding scale equals the fixture's recorded bf16-rounded value.

FeathBow:
## Description

Closes #878.

The full text-tower forward, prefill form: scaled embedding, all 48 layers dispatched by the parsed layer map, the final norm, the tied LM head, and the final-logit softcap — every prompt at or below the 1024-token window, where causal attention IS sliding attention. Geometry and the end constants come from the weights' own config inside the forward — one source of truth for every buffer size handed to CUDA — and token ids are bounds-checked against the embedding before upload, since the embedding kernel validates neither its output width nor the row it reads.

The end constants are the ones a from-the-paper implementation gets wrong: the embedding multiplier is the bf16 rounding of `sqrt(hidden_size)` (exactly 62.0 at hidden 3840 — the reference casts the scale to the weight dtype before multiplying), derived in one place and asserted against the fixture's recorded value before use; the LM head is the embedding matrix itself; the softcap rides the typed config, validated positive at parse.

The oracle compares log-probabilities at the reference's top-64 ids per position (host log-softmax over the f32 upcast, f64 accumulation for the 262k-way sum), with per-case gates declared from measurement and the window-edge gate distributional — the reasoning and the HF-self-comparison baseline are in the issue.

## Test Env

Single GPU (sm_89, x86_64), CUDA 12.9, against the pinned 12B checkpoint. 

## Verification

- **single** (1 token): max |dlogprob| 0.162 over the top-64, top-1 exact — under the 0.3 gate with top-1 required exact.
- **short** (9 tokens): max |dlogprob| 0.638, top-1 exact at all nine positions — under the 0.9 gate with top-1 required exact. Both short cases sit at the bf16 rounding floor.
- **edge** (1024 tokens, exactly the window): top-1 agreement **821/1024** with max |dlogprob| 16.19, misses spread across all four quarters (229/190/204/198, first miss at position 27) — against the distributional gate: global top-1 >= 0.75, chaos ceiling 20, and per-quarter top-1 >= 0.65 bounding localized damage — a whole-quarter failure cannot pass on the strength of the others. The baseline making those numbers meaningful: two HF attention backends — sdpa vs eager, same weights, same sequence — agree at 802/1024 with max 16.4, per-quarter 224/186/199/193 (worst quarter 72.7%); our forward sits at that noise floor.

## Type of Change

- [ ] Documentation update

### DIFF
--- a/pegainfer-core/src/ops.rs
+++ b/pegainfer-core/src/ops.rs
@@ -88,6 +88,7 @@ pub use pegainfer_kernels::ops::single_prefill_hd256_into;
 pub use pegainfer_kernels::ops::single_prefill_hd512_into;
 pub use pegainfer_kernels::ops::single_prefill_nhd_causal_into;
 pub use pegainfer_kernels::ops::single_prefill_nhd_noncausal_into;
+pub use pegainfer_kernels::ops::softcap_bf16_in_place;
 pub use pegainfer_kernels::ops::write_vec_into;
 #[cfg(feature = "kernel-call-trace")]
 pub use traced::embedding_batch;
--- a/pegainfer-gemma4/src/config.rs
+++ b/pegainfer-gemma4/src/config.rs
@@ -62,6 +62,10 @@ pub(crate) struct Gemma4Config {
     /// proportional rope tables.
     #[allow(dead_code)]
     pub(crate) global_rotary_dim: usize,
+    /// Applied as `cap * tanh(x / cap)` over the final logits; every
+    /// published size declares one.
+    #[allow(dead_code)]
+    pub(crate) final_logit_softcapping: f32,
 }
 
 #[cfg(feature = "gemma4")]
@@ -138,6 +142,11 @@ impl Gemma4Config {
             sliding_window > 0,
             "Gemma 4: sliding_window must be positive"
         );
+        let final_logit_softcapping = f32_field(tc, "text_config", "final_logit_softcapping")?;
+        anyhow::ensure!(
+            final_logit_softcapping > 0.0,
+            "Gemma 4: final_logit_softcapping {final_logit_softcapping} must be positive"
+        );
         Ok(Self {
             hidden_size: usize_field(tc, "hidden_size")?,
             intermediate_size: usize_field(tc, "intermediate_size")?,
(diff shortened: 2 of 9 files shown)


FeathBow:
@codex review please

PR #888, opened 2026-08-16
docs(skills): point vllm-bench install at vllm-project/vllm rust workspace

xiaguan:
## Summary

The standalone `vllm-project/vllm-bench` repository is archived and read-only; `vllm-bench` is now developed as a crate in the vLLM Rust workspace (`vllm-project/vllm` → `rust/src/bench`). The old `cargo install --git` / prebuilt-binary routes still work but are frozen at the last standalone commit.

Update the vllm-bench skill to the maintained path:

- build: `git clone --depth 1 https://github.com/vllm-project/vllm.git && cd vllm/rust && cargo build --release -p vllm-bench`
- install onto PATH from the same directory: `cargo install --path src/bench`
- drop the frozen standalone-repo install routes
- point the PegaInfer smoke example's checkout directory at `<vllm-repo>/rust`

Verified locally: `cargo build --release -p vllm-bench` succeeds and `vllm-bench --version` runs.

### DIFF
--- a/.agents/skills/vllm-bench/SKILL.md
+++ b/.agents/skills/vllm-bench/SKILL.md
@@ -21,26 +21,30 @@ Use this workflow for multi-turn work:
 
 ## Install
 
-Install from the upstream repository:
+`vllm-bench` has moved into the main vLLM repository as a crate in the Rust
+workspace (`vllm-project/vllm` → `rust/src/bench`); the standalone
+`vllm-project/vllm-bench` repo is archived and read-only. Build the maintained
+version from the new home:
 
 ```bash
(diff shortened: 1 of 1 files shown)


xiaguan:
Review comment addressed in 5414e6eb: the install command now runs `cargo install --path src/bench` from the `vllm/rust` directory the build steps leave the shell in — the previous `rust/src/bench` did resolve to the nonexistent `vllm/rust/rust/src/bench`. Good catch.

Scope note: this PR is now docs-only (skill install update). The stepped-bridge fix commit has been split off — tracked in #896 instead.

Issue #746, opened 2026-07-23
perf(qwen3): make fused QKV and gate/up GEMMs parity-safe


xiaguan:
## Context

Qwen3 already stores related projection weights contiguously at load time:

- `qkv_proj = vstack(q_proj, k_proj, v_proj)`
- `gate_up_proj = vstack(gate_proj, up_proj)`

The forward path still reads row ranges from those matrices and launches separate GEMMs: three for Q/K/V and two for gate/up. `openinfer-qwen3/src/batch_decode.rs` records why the previous fused-QKV path was not retained: under shard-local tensor parallelism, its numerical differences were large enough to flip greedy tokens in parity tests.

This is a focused follow-up to #456, which identifies fused QKV and gate/up projection as known roofline levers.

## Goal

Make the contiguous weight layout useful for compute by providing parity-safe fused projection paths:

1. one GEMM for Q/K/V followed by a lightweight split/deinterleave step;
2. one GEMM for gate/up followed by the existing SwiGLU elementwise step.

The implementation must preserve the current eager and CUDA Graph execution contracts and must fail closed for unsupported TP shapes.

## Scope

- Cover Qwen3 TP=1 and TP=2 first.
- Diagnose the current fused-QKV numerical divergence before changing the production path.
- Reuse the existing contiguous `qkv_proj` and `gate_up_proj` allocations.
- Keep LoRA behavior correct for logical q/k/v and gate/up projections.
- Avoid adding a generic abstraction unless another model has the same immediate need.

### Correctness

- Existing Qwen3 HF golden gates pass for TP=1.
- TP=2 greedy/logits parity passes on the current tensor-parallel coverage.
- Eager and CUDA Graph paths produce equivalent results under the repository's existing tolerances.
- LoRA-enabled and LoRA-disabled paths remain covered.

### Performance

Report measured A/B data before claiming a win:

- current split GEMMs vs fused projection path;
- TP=1 and TP=2;
- prefill and decode;
- at least concurrency 1 and 8;
- kernel timings plus end-to-end TPOT/throughput.

Keep the fused path only where the measured end-to-end result improves without weakening correctness gates.

## Relevant code

- `openinfer-qwen3/src/weights.rs`: load-time `DeviceMatrix::vstack`
- `openinfer-qwen3/src/batch_decode.rs`: split Q/K/V and gate/up GEMMs; parity note
- `openinfer-qwen3/src/prefill.rs`: row-range projection GEMMs
[...]

Issue #456, opened 2026-06-25
Tracking: Qwen3-4B kernel-list performance vs roofline


xiaguan:
## Why

#455 swapped the prefill QK-norm+RoPE kernel for a warp-per-token version and got **2.45x cold** — the old kernel sat at ~40% of DRAM roofline (one-thread-per-dim + two shared-mem reductions on a 128-dim memory-bound op). Nobody was watching it; we found it only by measuring one kernel by hand. The full Qwen3-4B kernel list has no systematic roofline tracking, so other kernels with similar headroom are invisible.

## Current state

- `qwen3_kernel_report` already computes per-kernel latency + speed-of-light (`DevicePeakBandwidth::query`, CUPTI `sm__throughput`, `smsp__warps_active`, tensor-pipe metrics) — the roofline machinery exists.
  - `openinfer-qwen3-4b/src/bin/qwen3_kernel_report.rs`, `openinfer-qwen3-4b/src/kernel_bench.rs`
- But the manifest only covers the attention block: `prefill_qk_norm_rope`, `prefill_kv_scatter`, `prefill_attention_core`, `paged_{decode,prefill}_attention`.
  - `openinfer-qwen3-4b/kernel_manifests/qwen3-4b.toml`
- The GEMMs that dominate the forward (QKV / O / gate_up / down / logits), the RMSNorms, and sampling are **not in the list** — so they are never ranked against their bound.

## Asks

1. Extend the manifest to the full Qwen3-4B forward (GEMMs, norms, sampling), prefill + decode.
2. Add a roofline-ranking pass: every kernel as % of its bound (DRAM BW for memory-bound, tensor / cuBLAS peak for compute-bound), sorted by headroom.
3. File follow-up issues for the worst offenders. Known levers from the decode-roofline pilot: fuse QKV / gate-up; cuBLAS picks an Ampere `sm_80` WMMA tile on Blackwell at bs>=2.
4. Decide whether a regression gate belongs in CI (the `docs/conventions` bench-regression convention already exists).

## Acceptance

- A reproducible command that prints every Qwen3-4B forward kernel with latency + % of roofline (cold, sm_120).
- A ranked headroom list with >=1 follow-up issue per kernel below ~70% of its bound.

Relates to #241 (Qwen3-4B roadmap). First acted-upon win: #455.

n-WN:
[...]

PR #895, opened 2026-08-17
perf(qwen3): fuse qualified TP1 decode projections

Polaris3003:
Refs #746

Follow-up to #456.

### Historical component-wise suite

The earlier SM86 suite measured QKV and gate/up independently across TP1/TP2
and decode/prefill-unified paths.

- correctness: 20/20;
- projection reports: 3/3;
- topology reports: 16/16;
- end-to-end benchmarks: 64/64;
- command errors: 0.

| Projection | Phase | TP | Mean E2E improvement | Component decision |
| --- | --- | ---: | ---: | --- |
| QKV | prefill/unified | 1 | +2.40% | KEEP_SPLIT: below the 3% threshold and kernel direction disagreed |
| gate/up | prefill/unified | 1 | +1.30% | KEEP_SPLIT |
| QKV | decode | 1 | +2.32% | ENABLE: all repetitions and kernel evidence agreed |
| gate/up | decode | 1 | +0.61% | KEEP_SPLIT: below the 2% threshold |
| QKV | prefill/unified | 2 | -0.12% | KEEP_SPLIT |
| gate/up | prefill/unified | 2 | -2.13% | KEEP_SPLIT |
| QKV | decode | 2 | +11.69% mean | KEEP_SPLIT: repeated throughput changed direction (`+85.02%` and `-7.03%`) |
| gate/up | decode | 2 | -2.96% | KEEP_SPLIT |

This experiment identifies TP1 decode QKV as the main independently qualified
contributor. Gate/up alone was positive but below the independent component
threshold. Prefill/unified and TP2 stayed split.

### Current atomic-path validation

The current minimal implementation treats TP1 decode QKV and gate/up fusion as
one immutable topology: `FusedQkvGateUp`. It was validated against the existing
split baseline on Qwen3-4B and Qwen3-8B.

Environment:

- candidate: `8634d9df4457e7d3c9cf3b14bc135a0b6a2e5ed8`;
- baseline: `b0be2d01d22ccedb4ce6e19ba0dd239ccca47322`;
- GPU: 2× NVIDIA GeForce RTX 4090, SM89;
- driver: 595.58.03;
- CUDA build toolchain: 12.6;
- HTTP matrix: 900/900 requests successful, zero failures/timeouts;
- eager-equivalence requests included: 905 total;
- artifact SHA256:
  `ba48ea44b1364ebea7c63ca208517bf02763341feb1abcfafa3f42d4ad1a96a6`.

HTTP results for the atomic fused path:

| Model | Concurrency | TPOT p50 improvement | Output throughput improvement |
| --- | ---: | ---: | ---: |
| Qwen3-4B | 1 | +1.46% | +1.40% |
| Qwen3-4B | 8 | +1.80% | +1.69% |
| Qwen3-8B | 1 | +0.98% | +0.98% |
| Qwen3-8B | 8 | +1.25% | +1.21% |

Both retained candidate measurements were faster than both retained baseline
measurements in every TP1 decode cell.

Direct Qwen3-4B decode results:

| Execution mode | Batch 1 TPOT | Batch 8 TPOT |
| --- | ---: | ---: |
| Eager | +2.26% | +3.82% |
| CUDA Graph | +1.50% | +2.42% |

CUDA Graph trace results over 100 steps:

| Batch | Baseline | Candidate | Improvement |
| ---: | ---: | ---: | ---: |
| 1 | 9.522967 ms | 9.380392 ms | +1.50% |
| 8 | 10.323344 ms | 10.039357 ms | +2.75% |

Correctness results:

| Gate | Result |
| --- | --- |
| Projection path selection | passed |
| QKV split bitwise-copy test | passed |
| Fused/split BF16 SwiGLU rounding | passed |
| gate/up LoRA row offsets | passed |
| Qwen3-4B HF golden, TP1/TP2, eager/graph | passed |
| Qwen3-8B HF golden, TP1/TP2, eager/graph | passed |
| Five-target LoRA golden, TP1/TP2 | passed |
| TP2 concurrent decode | passed |
| HTTP greedy output A/B parity | passed |
| Eager/CUDA Graph output parity | passed |

| Golden gate | Maximum mean delta | Maximum p99 delta |
| --- | ---: | ---: |
| Qwen3-4B HF | 0.0316 | 0.1135 |
| Qwen3-8B HF | 0.0308 | 0.1189 |
| Five-target LoRA | 0.0330 | 0.1215 |

The non-target prefill paths remained unchanged, with a maximum measured
regression of 0.52%. TP2 resolved to `Split`; its decode differences stayed
within 0.63%.

The Nsight databases retained whole-graph timing, but projection nodes were not
materialized separately. These results therefore support the atomic
end-to-end path and do not claim a new per-projection kernel breakdown on SM89.


xiaguan (changes requested):
Thanks for the correctness work and the end-to-end data. I am requesting changes because the current evidence does not yet justify this production default.

- The measured SM89 gain is only 0.98–1.80% TPOT / 0.98–1.69% output throughput, with two retained A/B measurements per HTTP cell, while the PR explicitly says the projection nodes were not measured separately. Please add current-head CUDA-event A/B for `3× Q/K/V GEMM` vs `packed GEMM + split_qkv`, split vs merged gate/up, and `split_qkv` itself at batch 1 and 8, with raw order-balanced repetitions.
- `split_qkv_kernel` is a handwritten full-buffer BF16 read/write pass with per-element division/modulo and branching, added to turn packed QKV into three owning buffers. vLLM's Qwen3 path uses one `QKVParallelLinear` and then splits the packed result into views. Please use that data flow as a design reference: preserve view/stride semantics downstream or fuse materialization into the QK-norm/RoPE/KV consumers. If the copy remains, its measured cost must be part of the decision.
- The selector is GPU-agnostic, but the evidence covers SM86/SM89. cuBLASLt tuning chooses an algorithm inside the fused shape; it does not compare fused and split topologies. Please gate by measured architecture/shape or benchmark both topologies in the decision.
- Historical component data qualified TP1 decode QKV (+2.32%) but not gate/up (+0.61%). Please do not couple gate/up to QKV without component evidence. This scope reduction should also remove low-value test volume: delete the helper self-test and reduce the seven-case copy matrix to one production shape plus one tail case; keep the pre-launch shape rejection and LoRA row-offset coverage where applicable.

Polaris3003:
Thanks for the review. I narrowed the production change to QKV-only decode fusion.

### Production scope

- Qwen3 TP1 decode uses one packed QKV GEMM followed by `split_qkv`.
- Gate/up remains split and is no longer coupled to QKV.
- Prefill, unified, TP2, DFlash, overlap, Pin, and PerToken remain split.
- The selector fails closed by TP, numeric policy, overlap mode, DFlash state,
  SM, and exact Qwen3 projection geometry.
- Only the measured SM89 Qwen3-4B and Qwen3-8B geometries are qualified.
- Existing contiguous `qkv_proj` and `gate_up_proj` allocations are reused.
- LoRA logical projection boundaries remain unchanged.

### Component A/B

The component measurements use CUDA Events, ABBA/BAAB ordering, batch sizes 1
and 8, and 40 raw samples per comparison.

For the complete QKV topology, including `packed GEMM + split_qkv`:

| Model | Batch 1 | Batch 8 |
| --- | ---: | ---: |
| Qwen3-4B | +2.887% | +15.903% |
| Qwen3-8B | +11.945% | +10.066% |

The standalone `split_qkv` cost is 3.33–3.59 us. Packed GEMM-only numbers were
not used as the production claim. Gate/up was measured independently but remains
split in production.

### End-to-end validation

Latest SM89 RTX4090 decode A/B:

| Model | Eager bs1 | Eager bs8 | Graph bs1 | Graph bs8 |
| --- | ---: | ---: | ---: | ---: |
| Qwen3-4B | +1.06% | +2.45% | +0.85% | +1.83% |
| Qwen3-8B | +0.66% | +1.39% | +0.51% | +1.04% |

The HTTP A/B completed 720/720 requests successfully. All 360 paired
baseline/candidate requests produced identical output hashes.

The HTTP trace did not provide trusted completion-token counts, so output tok/s
is not claimed as a formal metric. Stream TPOT and QPS are reported only as
supplementary E2E evidence.

### Materialization and test scope

The current implementation still materializes compact Q/K/V buffers rather than
introducing a generic stride/view abstraction. The materialization cost is
included in the complete QKV A/B above; the packed GEMM-only result is not used
as the production decision.

xiaguan (approved):
LGTM. The revised path is narrowly qualified to QKV-only TP1 decode on the measured SM89 model geometries, includes split materialization in the component A/B, keeps gate/up and every unqualified runtime on the established split path, and supplies matching correctness and end-to-end evidence.

PR #897, opened 2026-08-17
refactor(frontend): id-based Scheduler contract — RequestLedger replaces typestate handles

xiaguan:
## What

The `Scheduler` trait now deals in plain `RequestId`s:

```rust
pub trait Scheduler: Send {
    fn submit(&mut self, id: RequestId, request: Request);
    fn step(&mut self, ledger: &mut RequestLedger) -> anyhow::Result<()>;
    fn metrics(&self) -> SchedulerMetrics;
}
```

`StepEmitter` becomes **`RequestLedger`**: the account book of live requests. The driver opens an account per submission (`register`), the scheduler writes verdicts/tokens by id, exactly one terminal transition closes each account, and remaining accounts are swept to `Failed` on drop — or on fatal, with the real step error attached (previously the drop bombs could only say "dropped by the engine").

## Why

The typestate handles (`QueuedRequest`/`ActiveRequest`) promised compile-time "terminal exactly once", but every real consumer stored them in a `HashMap<RequestId, HandleSlot>` and re-derived the guarantee at runtime with `unreachable!`/`expect` — per model crate. Both consumers also unpacked the payload on the first line of `submit`, leaving the public type carrying a double `Option`, a `Box`, and a hidden "payload already taken" state for a flexibility nothing used. qwen3 even grew a second id space to key its own registry.

This moves the registry into the frontend, written once:

- `QueuedRequest`/`ActiveRequest` deleted from the public API. The one remaining Option-for-`Drop` dance lives in the crate-private `Submission` channel envelope (drop bomb still covers requests stuck in the channel or sent to a dead scheduler).
- Model crates delete their `HandleSlot` enums, handle maps, and take/reinsert plumbing (~-680 net lines); qwen3's executor ids are now the contract ids end to end.
- Scheduler-facing ledger methods are `pub`; `register`/`commit_step`/`fail_all` are `pub(crate)`, so the "model code never calls this" comment is now a compiler guarantee.
- `load()`/`LoadSnapshot` → `metrics()`/`SchedulerMetrics` (it already carries spec-decode counters and will grow more; struct moved to `metrics.rs`; legacy `load_watch` family renamed to match).

Trade-off, stated plainly: "terminal exactly once / nothing after terminal" is now a runtime panic in the ledger instead of a move-check — the same failure mode consumers already had, but enforced at one place with the offending call site in the stack.

## Testing

- `cargo check --tests --examples` clean for pegainfer-frontend, qwen3, k3, glm52, qwen35, gemma4, deepseek-v2-lite, kimi-k2, server (`PEGAINFER_CUDA_SM=103`, kernel-lab container).
- `cargo test --lib` passes for frontend (ledger/driver/bridge, 64 tests), k3, qwen3, qwen35. Ledger tests port the emitter suite plus a new drop-sweep and touch-after-close panic test; the qwen3/k3 contract test suites pass unchanged (they were already black-box over the step stream).
- Known gap: `pegainfer-glm52 --lib` reported a test failure in my container; glm52 changes here are pure renames (compile-clean) and I have not yet confirmed whether that failure pre-exists on main — flagging rather than hiding it.


xiaguan:
Follow-up pushed (72140270): `submit` now takes one plain record instead of two loose args —

```rust
pub struct QueuedRequest {
    pub id: RequestId,
    pub request: Request,
}

fn submit(&mut self, request: QueuedRequest);
```

`QueuedRequest` returns as the name for "submitted, not yet admitted" — but as transparent data, not a handle: no Option, no Box, no drop bomb (that stays in the crate-private `RequestEnvelope`). Consumers stop re-pairing `(id, request)` themselves; k3's local `PendingRequest` dissolved into it. The id deliberately stays out of `Request`: payloads are constructed by the protocol stack and tests, ids only by the wiring — a `pub id` on `Request` would either lie until submit or force every caller to mint first.

PR #913, opened 2026-08-18
refactor(gemma4): native split KV decode for the global family

### LINKED ISSUE #912: gemma4: the 12B global family decodes through a prefill kernel because GQA group 16 has no decode dispatch
## Description

The 12B checkpoint's global family is MQA at GQA group 16, which FlashInfer's decode dispatcher does not compile — the supported groups are 1, 2, 3, 4 and 8. That is the reason its batched decode steps still run the prefill-shaped hd512 kernel at seq_len 1, fed by a prefill plan the decode path refills every step, re-deriving CTA tile scheduling for a kernel that gets one query row per request. (Other sizes carry 2 or 4 global KV heads — dispatchable groups that need none of this.)

A native decode read is reachable without touching the kernel dispatch table: `[b, 16*512]` q rows reinterpreted as `[2b, 8*512]` is an exact memory identity, presenting each request as two pseudo-requests of 8 heads over the same pages — GQA group 8, which dispatches, and a mapping that is only an identity because MQA gives every query head the same KV head. Presented that way the non-partitioned grid is still only `(2b, 1 kv head)` CTAs, so occupancy needs the split-KV form of the kernel, which has no head_dim 512 instantiation yet.

There is also a coverage gap this touches: the checkpoint oracles decode through the single-request path, so the batched serving entry — the path production requests actually take — has no reference-anchored numeric gate.

## Proposed Plan

- Make the split-KV batch decode launcher generic over head_dim and mask variant (the hd128 extern entries delegate unchanged) and instantiate it at head_dim 512. Reshape the hd512 decode wrapper to the serving contract: attention read only — the prep kernel owns the pool writes — taking the split plan and merge workspace, validated through `checked_paged_geometry` like the prefill entries.
- Route the global family's batched decode through it, presenting each request as a geometry-derived number of pseudo-requests (the 12B group 16 over one KV head splits in two; dispatchable groups pass through; anything else refuses before the weights load): factor-repeated page metadata uploaded per step at graph-stable padded shapes, a fixed 256-token chunk CSR from qwen3's builder lifted into pegainfer-core, and FlashInfer's merge for the partial states.
- Retire the decode path's global prefill plan; both preps read the step's uploaded tables.
- Drive the checkpoint oracles' decode segments through the batched serving entry at batch one, anchoring the path production takes to the HF references; the ragged determinism gate already anchors the graph replay to it.

## Acceptance Criteria

- The checkpoint oracles hold with their decode segments on the batched serving entry: serve-vs-oracle forward, window crossing vs HF, eviction footprint, greedy vs HF generate.
- The ragged batch determinism gate holds: replay, per-step row moves and the eager comparison agree bit-for-bit per request.
- The serving behaviour checks hold, and the A/B against the current default is recorded.

FeathBow:
## Description

Closes #912

The 12B global family is MQA at GQA group 16, which FlashInfer's decode dispatcher does not compile (groups 1,2,3,4,8 only) — the reason its batched decode steps ran the prefill-shaped hd512 kernel at seq_len 1, fed by a prefill plan refilled every step. This change gives it a native split-KV decode read. Measured end to end it is a workload trade, not a general win — the table below — carried for that trade and for the structural retirement it buys: the decode path's prefill plan machinery goes, the batched serving entry gains reference-anchored gates, and the geometry contract fails loud instead of silently misreading.

The first commit makes the split-KV batch decode launcher generic over head_dim and mask variant — the two hd128 extern entries delegate unchanged — and instantiates it at head_dim 512. The hd512 decode wrapper becomes attention read only: the serving prep kernel has already written the step's normed+roped K and its V fork, so the wrapper's scatter step and its k/v params go, and the positions they consumed leave the metadata aggregate. It takes the split plan and merge workspace, and validates through `checked_paged_geometry` like the prefill entries.

The second commit routes the batched global decode through it. Each request is presented as `global_split_factor` pseudo-requests over the same pages, derived from the geometry before the weights load: a dispatchable GQA group passes through whole (factor 1), the 12B family's group 16 over one KV head halves into two 8-head pseudo-requests (`[b, 16*512]` reinterpreted as `[2b, 8*512]` is an exact memory identity, and only because MQA gives every query head the same KV head), and any other shape refuses loudly rather than silently remapping query heads to the wrong KV head. The read runs split-KV over fixed 256-token chunks: the CSR comes from qwen3's builder, lifted into pegainfer-core and consumed by both lines, padded to bucket-stable shapes for the decode graphs. The decode path's global prefill plan and its per-step CTA tile scheduling retire; both preps read the step's uploaded tables. The local family's decode is unchanged.

The checkpoint oracles now drive their decode segments through the batched serving entry at batch one, so the path production requests take is anchored to the HF references — the ragged determinism gate anchors the graph replay to that same path — and the batched step honors the eviction gate's release toggle the way the single-request step does.

## Test Env

Single GPU (sm_89, x86_64), CUDA 12.9, Gemma 4 12B checkpoint.

## Verification

- Ten build/lint/test gates green (fmt, kernels+gemma4 builds, clippy `-D warnings` incl. the CI crate set, hd256/hd512 smokes, window-prefill reject, gemma4 lib tests).
- Checkpoint oracles green with their decode segments on the batched serving entry: serve-vs-oracle forward, window crossing vs HF (multi-chunk global split at 1100+ tokens), eviction footprint, greedy vs HF generate.
- Ragged batch determinism gate green: replay, per-step row moves and the eager comparison agree bit-for-bit per request.
- qwen3 `hf_golden_gate` green — the hd128 split launcher this change makes generic is qwen3's production decode path.
- Live serving checks green; the only server-log warning is the documented tokenizer fallback.
- A/B on a quiet card, five arms x three runs, same session, `main` re-measured as the cleanliness canary (matches its historical band). Short context = 5-14-token prompts; long context = 1761-token prompts, 128 out:

| arm | TPOT ms | c1 | c8 | c9 | lctx TPOT ms | lctx c8 | idle MiB |
| --- | --- | --- | --- | --- | --- | --- | --- |
| main (prefill-shaped global read) | 27.50 | 34.1 | 136.9 | 142.2 | 29.26 | 130.1 | 33002 |
| head, 256-token chunks (this PR) | 27.56 | 34.1 | 136.8 | 141.4 | **28.64** | 127.5 | 33034 |
| probe: 512-token chunks | 27.54 | 33.8 | 136.3 | 140.9 | 29.07 | 124.7 | 33034 |
| probe: 1024-token chunks | 27.60 | 33.9 | 135.5 | 138.7 | 29.95 | 122.7 | 33002 |
| probe: whole-context chunk (≈ non-partition head split) | 27.66 | 33.9 | 136.3 | 138.8 | 30.67 | 121.0 | 33002 |

  Attribution: the split is what earns the single-stream long-context win (lctx TPOT −2.1% at 256; the non-partition head split alone is the *worst* arm, +4.8% over main), and 256 dominates every other chunk size on every metric. The cost side: long-context batch throughput −2.0% (130.1→127.5) and c9 −0.6%; short context is neutral. Net: a symmetric ±2% latency-vs-throughput trade by workload, plus 32 MiB resident.
- Resident at the final tip, quiet card: 33034 MiB idle, 33386 MiB peak under the serving checks (card-scale, matching the serving doc's accounting); the doc's numbers are updated in this change.
- The final tip re-ran the full chain after the geometry-factor rework: build/clippy/lib/det green, oracles green, qwen3 golden green, live green, and the A/B-relevant paths are unchanged for the 12B factor-2 geometry (identical kernels and launch shapes).


FeathBow:
@codex review please

PR #931, opened 2026-08-19
feat(k3): speculative decoding with the RadixArk DSpark drafter

xiaguan:
## What

End-to-end speculative decoding for K3, drafted by the RadixArk community DSpark checkpoint (DFlash block-diffusion, block 7 — anchor + 6 drafts). Armed per rank with `--dflash-draft-model-path`; unarmed serving is untouched.

- **Verify step mode** (`K3StepMode::Verify`): packed per-slot segments — deferred-commit KDA replay rows then the speculative span. Chunkwise FlashKDA per segment with parity as a per-segment double buffer; absorbed paged MLA over a packed verify block table with store-semantics latent appends. Once a slot uses verify, all its decode goes through verify (a plain decode step would advance every row at the global parity).
- **Aux capture** (`K3AuxSink`): after each tap layer (`[7, 23, 51, 67, 83]`, checkpoint ids as spelled — no vLLM off-by-one), the step's leading rows land in a `[rows, 35840]` slab. Prefill chunks capture their live tokens; a verify step's accepted span rows become the drafter's next pending context.
- **Draft lane** (`dspark.rs`): eager 5-layer qwen3 GQA (64Q/16KV, head 64) at K3's hidden 7168, YaRN rope (factor 16, transformers semantics, attention_factor baked in), rank-256 anchor-drop Markov head, dual-source varlen attention over `[cached context ; block]`. Reuses the target's embed/lm_head (the checkpoint ships neither). Allocation-free after arm.
- **Scheduling**: `StepExecutor::decode_many` (multi-token rounds; default wraps plain decode), stop/length walked in order per round; per-request acceptance telemetry at release. Verify batches split into row-budget-sized steps (free-running EP peers pad); arming requires the budget to fit one slot's worst round (14 rows).

## Validation

- **Gates** (`tests/spec_verify.rs`, 224-expert checkpoint on GB300, all green): verify ≠ plain decode bitwise *by construction* (chunkwise FlashKDA vs fused core, packed-bucket projections — near-tie argmaxes flip), so the gates hold what is exact under identical launch geometry: determinism + history/page-scramble independence, rejected-draft-content non-leakage, packed-slot isolation, corruption-value invariance, margin-bounded oracle tracking, and a real-drafter round-trip gate.
- **Full-depth EP4 serve** (93 layers, 4×GB300): sequential and 8-way concurrent greedy smokes with zero errors; the concurrent round packs two slots per rank and exercises the verify split. Spec-vs-plain A/B: 2/4 prompts byte-identical, 2 fork on single-token near-ties. Acceptance 1.0–2.08 tokens/round against the pruned target (the drafter was trained against full K3 — judge quality on the 896-expert model).
- The first serve surfaced three bugs unreachable by single-thread/single-rank tests, fixed in `fix(k3)`: thread-local cuBLAS handle starved by the executor's thread handoff, a rows=0 aux sink on padding steps, and the verify row-budget overflow.

`cargo fmt --check` clean; k3 lib tests 41/41; workspace lib tests green except the known GPU-requiring `kv_pool` trio (no GPU in the build container).


xiaguan:
@codex review

PR #937, opened 2026-08-19
Restore the star history chart

Dessalines39394:
The star history chart was removed from the README because it had broken, not because it was unwanted. The upstream star-history.com endpoint no longer renders reliably under GitHub's stargazer API restrictions, and fixing it is necessary to keep the project's adoption visible.

This restores the Star History section to its original place after License, repointed to a working mirror that needs no API token and serves both the frontend and the chart image from the same host.


xiaguan:
Thanks for the contribution. We prefer to keep the Star History section removed rather than replace it with another external chart mirror. Please drop the Star History change; since it is the only change in this PR, the PR can then be closed.

xiaguan:
Closing per the earlier discussion — we'd rather keep the Star History section out of the README. Thanks for the contribution!

PR #814, opened 2026-08-01
PR: wire prefix-cache hit counters into /metrics 

1571859588:
## Summary

Expose real prefix-cache query/hit counters end-to-end so they surface in the
vLLM-compatible `/metrics` endpoint, matching vLLM's `prefix_cache_stats`
semantics (token granularity).

- `openinfer-engine`: add `prefix_cache_queries` / `prefix_cache_hits` to `LoadSnapshot`.
- `openinfer-vllm-frontend` (`bridge.rs`): map the new `LoadSnapshot` fields into
  `SchedulerStats.prefix_cache_stats`.
- `openinfer-qwen3` scheduler: aggregate `cached_tokens` from prefill resolution
  into per-step `StepEffects` and accumulate across both scheduler loops
  (non-LoRA and LoRA), publishing real query/hit counts.
- `openinfer-qwen35`: carry the counters through its scheduler publish path.
- `openinfer-sim` (`frontend_e2e`): assert the counters are emitted.

## Verified

Built and tested on A100 (SM 80, CUDA 12.4, nightly rust), card 5:

- `cargo build --release` (qwen3) — OK
- `cargo test --release --workspace --lib` — 74 passed
- `cargo test --release -p openinfer-vllm-frontend` — 27 passed
- `cargo test --release -p openinfer-sim --test frontend_e2e` — 13 passed (real GPU)
- `cargo fmt --all -- --check` — clean

## Notes

Cargo.lock reflects the verified local dependency resolution (bindgen 0.72.1)
required to build qwen3 on this environment.

## Test plan

- [ ] CI full CUDA build (qwen3) green
- [ ] `prefix_cache_queries` / `prefix_cache_hits` present in `/metrics` under load
- [ ] multi-turn repeat-prompt request shows hits > 0

PR #669, opened 2026-07-13
metric(qwen3): report prefix-cache query/hit

Kokoro2336:
## Description

Resolves #603 

Report prefix-cache query/hit

## Type of Change

- [x] New feature (non-breaking change which adds functionality)

## Checklist

- [x] My code follows the style guidelines of this project (see `docs/conventions/coding-style.md`).

- [x] I have performed a self-review of my own code.

- [x] I have formatted my commits according to Commitizen conventions.

- [x] I have run the local test suite and all tests pass (see `CLAUDE.md`).

Issue #603, opened 2026-07-07
metrics: report real prefix-cache query/hit counters from the qwen3 scheduler


xiaguan:
> **TL;DR:** `vllm:prefix_cache_queries_total` / `vllm:prefix_cache_hits_total` read 0 because the bridge sends `SchedulerStats::default()` for them. The qwen3 scheduler already knows the per-request hit (`Scheduled.cached_tokens` drives `prompt_tokens_by_source_total{source="local_cache_hit"}`), but the aggregate query/hit counters need interval deltas carried on `LoadSnapshot` (or a sibling channel). Part of #602.

Upstream semantics (`~/.cargo` checkout `rust/src/engine-core-client/src/metrics.rs`, `record_scheduler_stats_with_handles`): `prefix_cache_stats.base.queries/hits` are **interval deltas** recorded via `inc_by`, not cumulative values — sending cumulative totals every snapshot would over-count. Whatever carries them must reset per publish.

Scope:
- Count prefix-cache block queries/hits per scheduler step in the qwen3 scheduler.
- Extend the snapshot the bridge forwards (delta-safe), fill `SchedulerStats.prefix_cache_stats`.
- Assert the counters move in the bridge test and in a live `/metrics` scrape with a repeated ≥1-block prompt.

Done when: two identical ≥16-token prompts produce `queries > 0, hits > 0` on `/metrics`.

Kokoro2336:
Hi, I'd like to work on this issue

PR #952, opened 2026-08-21
fix(metrics): stamp spec-decode counters on the stepped bridge

scatyf3:
Issue #896 

## Type of Change

Bug fix (non-breaking change which fixes an issue)

## what's new

- Extract `SpecDecodeTracker::interval`: one cumulative-to-per-interval converter shared by both bridges to report SD status update. 
- Stamp `spec_decoding_stats` on the `SteppedEngineBridge`
- Add an opt-in pretend drafter to (`with_speculative_decoding(K, accepted)`) `pegainfer-sim` 
- New e2e `stepped_bridge_reports_spec_decode_counters_to_prometheus` test 
- remove spec-decode counters bug on document

## correctness check

green CI

E2E:
- model qwen3 4b+dflash
- Dataset/Protocol: ShareGPT 20 prompts、c4、--temperature 0 --ignore-eos、seed 0 via vllm bench.

The correctness check is our vllm frontend is wired to vllm bench:

```shell
vllm bench
Drafts:              1040
Draft tokens:        7023
Accepted tokens:     1500
Per-position acceptance (%):
  Position 0:  65.19    Position 4:   5.87
  Position 1:  36.83    Position 5:   3.27
  Position 2:  20.67    Position 6:   1.73
  Position 3:  10.67
```

## Checklist

- [ ] My code follows the style guidelines of this project (see `docs/conventions/coding-style.md`).
- [ ] I have performed a self-review of my own code.
- [ ] I have formatted my commits according to Commitizen conventions.
- [ ] I have run the local test suite and all tests pass (see `CLAUDE.md`).

PR #953, opened 2026-08-23
feat(observability): report real prefix-cache query/hit counters in /metrics

1571859588:
## Summary

Threads prefix-cache query/hit counters from the qwen3 scheduler through
`SchedulerMetrics` into the vLLM `SchedulerStats.prefix_cache_stats` surface,
so Prometheus `/metrics` no longer reports zeros for prefix-cache hit rate.

- `pegainfer-qwen3`: accumulate per-step `prefix_queries`/`prefix_hits` in
  `StepEffects` (one query per first-chunk request; hits = `cached_tokens`),
  fold into cumulative counters on the scheduler, expose via `metrics()`.
- `pegainfer-frontend`: add `prefix_cache_queries`/`prefix_cache_hits` to
  `SchedulerMetrics` and map them to `PrefixCacheStats` in the vLLM bridge.

## Relation to #814 / #669 / #603

This supersedes #814 (which was opened from `feat/sim-frontend-prefix-cache-metrics`,
head `7f24658e`, against the pre-rename tree and went CONFLICTING due to
`#841` PegaInfer rename + `#824` kv-store refactor). Per the maintainer note
on #814, this is the single canonical PR for the prefix-cache reporting surface.

Scope split to avoid two parallel implementations:
- **usage surface** (`cached_tokens` → OpenAI usage): already landed upstream
  via `TokenEvent::Scheduled` (issue #603 line). Not touched here.
- **`/metrics` surface** (`prefix_cache_queries/hits` → Prometheus): this PR.

## Verified

- `cargo test -p pegainfer-frontend --lib` — 65 passed (incl. bridge
  `prefix_cache_stats` mapping).
- `cargo clippy` + `cargo check --workspace --lib` — clean.

Note: a full qwen3 A100 e2e was blocked in this local environment by a
`rdma-mummy-sys` bindgen issue (missing vendored `rdma-core-mummy` headers) that
is unrelated to this change and does not touch rdma code; upstream CI with the
full rdma toolchain will validate the qwen3 build.

### DIFF
--- a/pegainfer-frontend/src/vllm/bridge.rs
+++ b/pegainfer-frontend/src/vllm/bridge.rs
@@ -37,7 +37,6 @@ use vllm_engine_core_client::protocol::request::EngineCoreRequest;
 use vllm_engine_core_client::protocol::request::EngineCoreRequestType;
 use vllm_engine_core_client::protocol::stats::BaseCacheStats;
 use vllm_engine_core_client::protocol::stats::PrefillStats;
-use vllm_engine_core_client::protocol::stats::PrefixCacheStats;
 use vllm_engine_core_client::protocol::stats::SchedulerStats;
 use vllm_engine_core_client::protocol::stats::SpecDecodingStats;
 use vllm_engine_core_client::protocol::utility::UtilityCallId;
@@ -591,6 +590,14 @@ fn stop_sentinel_id(eos_token_id: Option<u32>, stop_token_ids: &[u32]) -> Option
 /// vLLM `SchedulerStats` view of a load snapshot — what the frontend's
 /// Prometheus gauges (`scheduler_running`, `scheduler_waiting`,
 /// `kv_cache_usage`) and DP load balancer consume.
+///
+/// `prefix_cache_stats` is left at zero here on purpose: the running totals in
+/// `SchedulerMetrics` must NOT be shipped as-is, because the frontend's
+/// Prometheus logger increments its `prefix_cache_*_total` counters by the
+/// value of *every* `SchedulerStats` it receives. The bridge therefore exports
+/// per-send **deltas** (see [`PrefixCacheTracker`]) so a cached request does
+/// not re-add the whole history on each subsequent token batch. Callers fill
+/// `prefix_cache_stats` from [`PrefixCacheTracker::interval`] after this call.
 pub(crate) fn scheduler_stats_from(snapshot: &SchedulerMetrics) -> SchedulerStats {
     SchedulerStats {
         num_running_reqs: snapshot.num_running_reqs,
@@ -600,15 +607,37 @@ pub(crate) fn scheduler_stats_from(snapshot: &SchedulerMetrics) -> SchedulerStat
         } else {
             snapshot.kv_used_blocks as f64 / snapshot.kv_total_blocks as f64
         },
-        prefix_cache_stats: PrefixCacheStats {
-            base: BaseCacheStats {
-                queries: snapshot.prefix_cache_queries,
-                hits: snapshot.prefix_cache_hits,
-                ..BaseCacheStats::default()
-            },
-            ..PrefixCacheStats::default()
-        },
         ..SchedulerStats::default()
+    }
+}
+
+/// Prefix-cache counterpart of [`SpecDecodeTracker`]: the scheduler holds
+/// running totals while the wire must carry per-interval deltas, because the
+/// frontend increments its `prefix_cache_*_total` counters by the value of
+/// *every* `SchedulerStats` it receives. Shipping the running total would
+/// re-add the whole history on each subsequent batch, so both bridges convert
+/// through this type and cannot drift.
+#[derive(Default)]
+pub(crate) struct PrefixCacheTracker {
+    last_queries: u64,
+    last_hits: u64,
+}
+
+impl PrefixCacheTracker {
+    /// The delta to stamp on the next outgoing batch. Advances the baseline, so
+    /// a caller that declines to send after calling this drops only a no-op
+    /// interval.
+    pub(crate) fn interval(&mut self, snapshot: &SchedulerMetrics) -> BaseCacheStats {
+        let delta = BaseCacheStats {
+            queries: snapshot
+                .prefix_cache_queries
+                .saturating_sub(self.last_queries),
+            hits: snapshot.prefix_cache_hits.saturating_sub(self.last_hits),
+            ..BaseCacheStats::default()
+        };
+        self.last_queries = snapshot.prefix_cache_queries;
+        self.last_hits = snapshot.prefix_cache_hits;
+        delta
     }
 }
 
@@ -672,9 +701,11 @@ async fn publish_scheduler_stats(
     shutdown: CancellationToken,
 ) -> Result<()> {
     let mut spec = SpecDecodeTracker::default();
+    let mut prefix = PrefixCacheTracker::default();
     loop {
         let snapshot = *load_rx.borrow_and_update();
         let mut stats = scheduler_stats_from(&snapshot);
+        stats.prefix_cache_stats.base = prefix.interval(&snapshot);
         stats.spec_decoding_stats = spec.interval(&snapshot);
         let outputs = RequestBatchOutputs {
             engine_index,
--- a/pegainfer-frontend/src/vllm/bridge.rs
+++ b/pegainfer-frontend/src/vllm/bridge.rs
@@ -35,7 +35,9 @@ use vllm_engine_core_client::protocol::output::StopReason;
 use vllm_engine_core_client::protocol::output::UtilityCallOutput;
 use vllm_engine_core_client::protocol::request::EngineCoreRequest;
 use vllm_engine_core_client::protocol::request::EngineCoreRequestType;
+use vllm_engine_core_client::protocol::stats::BaseCacheStats;
 use vllm_engine_core_client::protocol::stats::PrefillStats;
+use vllm_engine_core_client::protocol::stats::PrefixCacheStats;
 use vllm_engine_core_client::protocol::stats::SchedulerStats;
 use vllm_engine_core_client::protocol::stats::SpecDecodingStats;
 use vllm_engine_core_client::protocol::utility::UtilityCallId;
@@ -597,6 +599,14 @@ pub(crate) fn scheduler_stats_from(snapshot: &SchedulerMetrics) -> SchedulerStat
             0.0
         } else {
             snapshot.kv_used_blocks as f64 / snapshot.kv_total_blocks as f64
+        },
+        prefix_cache_stats: PrefixCacheStats {
+            base: BaseCacheStats {
+                queries: snapshot.prefix_cache_queries,
+                hits: snapshot.prefix_cache_hits,
+                ..BaseCacheStats::default()
+            },
+            ..PrefixCacheStats::default()
         },
         ..SchedulerStats::default()
     }
--- a/pegainfer-frontend/src/vllm/bridge.rs
+++ b/pegainfer-frontend/src/vllm/bridge.rs
@@ -37,6 +37,7 @@ use vllm_engine_core_client::protocol::request::EngineCoreRequest;
 use vllm_engine_core_client::protocol::request::EngineCoreRequestType;
 use vllm_engine_core_client::protocol::stats::BaseCacheStats;
 use vllm_engine_core_client::protocol::stats::PrefillStats;
+use vllm_engine_core_client::protocol::stats::PrefixCacheStats;
 use vllm_engine_core_client::protocol::stats::SchedulerStats;
 use vllm_engine_core_client::protocol::stats::SpecDecodingStats;
 use vllm_engine_core_client::protocol::utility::UtilityCallId;
@@ -621,12 +622,14 @@ pub(crate) fn scheduler_stats_from(snapshot: &SchedulerMetrics) -> SchedulerStat
 pub(crate) struct PrefixCacheTracker {
     last_queries: u64,
     last_hits: u64,
+    last_external_queries: u64,
+    last_external_hits: u64,
 }
 
 impl PrefixCacheTracker {
-    /// The delta to stamp on the next outgoing batch. Advances the baseline, so
-    /// a caller that declines to send after calling this drops only a no-op
-    /// interval.
+    /// The local delta to stamp on the next outgoing batch. Advances the
+    /// baseline, so a caller that declines to send after calling this drops
+    /// only a no-op interval.
     pub(crate) fn interval(&mut self, snapshot: &SchedulerMetrics) -> BaseCacheStats {
         let delta = BaseCacheStats {
             queries: snapshot
@@ -638,6 +641,32 @@ impl PrefixCacheTracker {
         self.last_queries = snapshot.prefix_cache_queries;
         self.last_hits = snapshot.prefix_cache_hits;
         delta
+    }
+
+    /// The same conversion for the external (connector) side — blocks restored
+    /// from CPU offload or over P2P rather than found in local KV. Kept on a
+    /// separate baseline so the two families cannot pollute each other, and
+    /// `None` when nothing was restored: a line with no connector leaves
+    /// `connector_prefix_cache_stats` unset instead of reporting zeros.
+    pub(crate) fn external_interval(
+        &mut self,
+        snapshot: &SchedulerMetrics,
+    ) -> Option<PrefixCacheStats> {
+        let base = BaseCacheStats {
+            queries: snapshot
+                .prefix_cache_external_queries
+                .saturating_sub(self.last_external_queries),
+            hits: snapshot
+                .prefix_cache_external_hits
+                .saturating_sub(self.last_external_hits),
+            ..BaseCacheStats::default()
+        };
+        self.last_external_queries = snapshot.prefix_cache_external_queries;
+        self.last_external_hits = snapshot.prefix_cache_external_hits;
+        (base.queries > 0 || base.hits > 0).then(|| PrefixCacheStats {
+            base,
+            ..PrefixCacheStats::default()
+        })
     }
 }
 
@@ -706,6 +735,7 @@ async fn publish_scheduler_stats(
         let snapshot = *load_rx.borrow_and_update();
         let mut stats = scheduler_stats_from(&snapshot);
         stats.prefix_cache_stats.base = prefix.interval(&snapshot);
+        stats.connector_prefix_cache_stats = prefix.external_interval(&snapshot);
         stats.spec_decoding_stats = spec.interval(&snapshot);
         let outputs = RequestBatchOutputs {
             engine_index,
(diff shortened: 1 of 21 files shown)


xiaguan:
Please fix the token-vs-request query-unit mismatch and cumulative-counter double counting, then provide an exact-head, multi-batch scheduler-to-/metrics E2E result showing stable non-zero deltas across scrapes; the PR currently reports that Qwen3 E2E was blocked, so no E2E result has been provided.

1571859588:
## Summary
Fixes two issues flagged on the prefix-cache /metrics counters:

Token-vs-request query-unit mismatch (pegainfer-qwen3/src/scheduler/resolve.rs): prefix_queries was incremented per-request (+= 1) while prefix_hits was token-granular (+= result.cached_tokens), so repeated prompts reported >100% hit rates. Both are now token-granular: prefix_queries += req.prompt_tokens.len(), prefix_hits += result.cached_tokens (cached_tokens ≤ prompt_tokens ⇒ rate ∈ [0,1]).
Cumulative-counter double counting (pegainfer-frontend/src/vllm/bridge.rs + bridge/stepped.rs): dispatch_step shipped the scheduler's monotonic totals every batch; vLLM's Prometheus increments *_total by each value, over-counting until restart. Now it ships per-send deltas (prefix_cache_delta(last, cur)), mirroring the existing spec-decode delta path. Last-state is tracked in both the legacy publish_scheduler_stats loop and the &self stepped dispatch_step (via AtomicU64).

## Verification
cargo test --release -p pegainfer-frontend --lib vllm::bridge → 14/14 passed, including prefix_cache_stats_are_per_interval_deltas_not_running_totals (interval 1 ships (100,37); interval 2 after totals reach (200,74) ships delta (100,37) not (200,74); idle ships (0,0)).
Scheduler→metrics token-granular path verified via a line-for-line replica: 4 batches × 3 reqs, prompt=64/hit=37 → totals (768,444), rate 0.578, stable per-batch deltas (+192,+111), no overcount.

## Note on E2E
The live A100 /metrics multi-batch scrape requested in review could not be produced in this environment: the qwen3 build chain pulls rdma-mummy-sys, whose bindgen (0.66) generates _address fields under the local toolchain, breaking that third-party crate's compile — unrelated to this change. The overcount behavior is directly covered by the passing bridge delta test. If CI runs on this head, the bridge test goes green; I can supply a real /metrics scrape snapshot once a qwen3-buildable environment is available.

FeathBow (changes requested):
Thanks! please rebase on main :)

1571859588:
> Thanks! please rebase on main :)

Rebased on main (now 419ba7fa, based on 99b68f3b).

Worth flagging: while rebasing onto 95b02cc4 (#952 — "stamp spec-decode counters on the stepped bridge"), I found that PR introduced exactly the abstraction this branch needed. It extracted SpecDecodeTracker::interval as the shared cumulative→per-interval converter for the spec-decode counters. The prefix-cache counters had the identical bug (scheduler holds running totals, the wire must carry deltas, or the frontend's *_total counters re-add history every batch), so I mirrored it with PrefixCacheTracker and routed both bridges through it, rather than keeping my earlier free-function prefix_cache_delta plus the AtomicU64 last-state fields on SteppedEngineBridge. Both counters now go through one pattern and cannot drift.

Also fixed while rebasing: token-granular units in resolve.rs (prefix_queries was += 1 per request against token-granular prefix_hits), plus rustfmt and a missing DCO sign-off on the first commit.

Verified: cargo test --release -p pegainfer-frontend --lib → 68/68 passed, including prefix_cache_stats_are_per_interval_deltas_not_running_totals.

FeathBow (changes requested):
please fix CI probs.

1571859588:
Fixed. SchedulerMetrics gained prefix_cache_queries/prefix_cache_hits, but several model lines initialize it as a struct literal with no ..Default fallback, so the new fields broke their build. Zero-initialized the fields at every literal site — gemma4, qwen35 (×3, incl. one test), and glm52, which the CI matrix doesn't compile but would have broken identically.

Both lines report zeros for now, which is honest for them: qwen35's hybrid Gated DeltaNet state is linear and not prefix-reusable, and gemma4's conversation-tail cache resolves hits through send_scheduled's cached_tokens rather than through counters — wiring it up is a follow-up.

Verified locally: cargo check --release -p pegainfer-gemma4 --features gemma4 --lib and -p pegainfer-qwen35 --features qwen35 --lib --tests both pass (Triton AOT included), rustfmt clean, plus the frontend tests at 68/68.

FeathBow (changes requested):
I would appreciate it if you could rebase onto main first and keep an eye on the CI while working on the fix.

1571859588:
I've rebased onto the main branch and will keep an eye on CI while working on the fix.

Rebased on main (385ffc31, based on 70a600b7). Refactor from #968 splits qwen35's scheduler into `scheduler/mod.rs` + `tp.rs`. Rename detection cleanly carried over the SchedulerMetrics initializer fix. I re‑verified on the rebased head: `cargo check -p pegainfer-qwen35 --features qwen35 --lib --tests` passes, frontend suite 68/68, rustfmt is clean. I am now watching CI on the new head.

xiaguan (changes requested):
Thank you for rebasing the PR and fixing the new metrics-field initializers. I re-reviewed the current head (`385ffc31`), and the original metrics issues are still present:

1. `resolve_prefill_outputs` increments `prefix_queries` for every request's first chunk, while `schedule_prefill_chunk` only performs `match_and_add_prefix` when the prefix cache is enabled and `echo` is false. Cache-disabled and echo requests therefore report lookups that never happened. Please derive the counters from the executor's actual lookup result instead of inferring them from prompt length.

2. CPU-offload and P2P-restored blocks are absorbed by the same `match_and_add_prefix` call and then reported as local `prefix_cache_stats.base.hits`. Please preserve the source and populate the connector/external counters separately.

3. In the stepped bridge's `outputs.is_empty()` branch, `self.stats(prefix, spec)` advances the prefix tracker before the code decides whether to send. When there is a prefix delta but no speculative-decoding delta, the batch is dropped and that prefix delta is lost permanently. Please send a stats-only batch whenever either delta is present, and add a metadata-only cached-step followed by abort regression test.

4. The current tests validate synthetic scheduler totals and synthetic bridge deltas separately. Please add one live path through real lookup, the stepped bridge, the vLLM metrics logger, and a `/metrics` scrape. It should cover cache disabled/echo, local miss/hit, external restore attribution, and the metadata-only abort case.

The token-granularity and cumulative-to-interval conversions themselves look good, and the current public CI is green. Once the four cases above are covered, this should be ready for another focused review. Thank you.

1571859588:
Thanks — addressed 1 and 3, and I need your steer on 2 and 4.

1. Phantom queries — fixed. ModelExecutor gained prefix_cache_enabled() (defaulted, so the 8 test-side prefill_item constructors didn't need churning); Qwen3Executor delegates to its existing inherent method so the speculative-decoding override still applies. resolve_prefill_outputs now counts only requests the executor actually looked up:

```rust
if req.prefill_pos == 0 {
    effects.cached.push(...);                                  // unchanged
    if executor.prefix_cache_enabled() && !req.echo {
        effects.prefix_queries += req.prompt_tokens.len() as u64;
        effects.prefix_hits += result.cached_tokens as u64;
    }
}
```

I deliberately left effects.cached outside the gate — that drives the cached_tokens usage field from #603, which shouldn't change behaviour here. Regression test: prefix_cache_disabled_reports_no_lookups (fake executor with the cache off, drained to terminal, asserts both counters stay 0).

3. Stranded prefix delta — fixed. self.stats(prefix, spec) had already advanced the tracker while the send decision only looked at spec_decoding_stats. It now ships whenever either delta is non-zero. Frontend suite passes (68/68) locally.

2. External attribution — I need a pointer. I traced the wire contract and BaseCacheStats is { reset, requests, queries, hits }, with PrefixCacheStats adding only preempted_{requests,queries,hits} — there is no external/connector field to populate. Separating CPU-offload / P2P-restored hits from local ones would mean extending that struct in the vllm client crate (a git dependency), which feels out of scope here. Two options: (a) tell me where you want the external counts to land, or (b) I keep it honest by ensuring externally-restored tokens are simply not reported as local hits — though even that needs a source split out of match_and_add_prefix, which currently returns just usize. Happy to do (b) as a follow-up PR if you'd rather not grow this one.

4. Live e2e — needs a decision. pegainfer-sim has no prefix cache at all (grep for prefix|cached_tokens|match_and_add in pegainfer-sim/src returns nothing), so the stepped_bridge_reports_spec_decode_counters_to_prometheus pattern doesn't transfer: there's no real lookup to drive. On the qwen3 side I can't build locally — rdma-mummy-sys fails under the local libclang — so a genuine real-lookup → bridge → logger → /metrics test would have to be written and validated via CI. Is that acceptable, or would you prefer I add the lookup plumbing to sim so the e2e can run on CPU?

Note qwen3 is unverified locally for the same reason; the two qwen3 changes are small and mechanical but will need CI to confirm.

pegainfer-frontend/src/vllm/bridge.rs
```diff
@@ -602,6 +611,36 @@ pub(crate) fn scheduler_stats_from(snapshot: &SchedulerMetrics) -> SchedulerStat
...
 }
 
+/// Prefix-cache counterpart of [`SpecDecodeTracker`]: the scheduler holds
+/// running totals while the wire must carry per-interval deltas, because the
+/// frontend increments its `prefix_cache_*_total` counters by the value of
+/// *every* `SchedulerStats` it receives. Shipping the running total would
+/// re-add the whole history on each subsequent batch, so both bridges convert
+/// through this type and cannot drift.
+#[derive(Default)]
+pub(crate) struct PrefixCacheTracker {
```
FeathBow, line 618: On the pinned vllm rev, SchedulerStats carries connector_prefix_cache_stats: Option<PrefixCacheStats> (protocol/stats.rs:183) and the metrics crate registers vllm:external_prefix_cache_queries / vllm:external_prefix_cache_hits (metrics/src/scheduler.rs:168-169), so nothing in the git dependency needs extending. On the qwen3 side the split is available at the lookup site: schedule_prefill_chunk runs right after the prefetch settles and PrefetchState carries the probe's gpu_hit_blocks(); local = gpu_hit_blocks x block_size (the whole match when the request had no prefetch entry) and external = matched − local. I recommend a second pair of counters (prefix_cache_external_queries/hits) on SchedulerMetrics, a second PrefixCacheTracker stamping connector_prefix_cache_stats, and the split computed in schedule_prefill_chunk; match_and_add_prefix itself can stay a usize.

pegainfer-frontend/src/vllm/bridge/stepped.rs
```diff
@@ -224,9 +236,16 @@ impl SteppedEngineBridge {
...
         if outputs.is_empty() {
             // A drafted step with no batch to ride would strand its increment
-            // until the next batch, which may never come.
-            let stats = self.stats(spec);
-            if stats.spec_decoding_stats.is_some() {
+            // until the next batch, which may never come. A prefix-cache delta
+            // is stranded the same way, so ship on either one: `stats()` has
+            // already advanced both trackers, and anything not sent here is
+            // dropped for good.
+            let stats = self.stats(prefix, spec);
```
FeathBow, line 243: when either delta is non-zero, which is the right fix, but the only new bridge test drives publish_scheduler_stats (the legacy loop); nothing exercises this stepped branch. The requested case is: a step whose only update is a cached-tokens metadata record (no tokens, no terminal) followed by an abort, asserting the prefix delta reaches the output channel in a stats-only batch and is not re-sent later.

pegainfer-qwen3/src/scheduler/resolve.rs
```diff
@@ -100,13 +104,38 @@ fn resolve_prefill_outputs(
...
+            // The cached-token report drives the `cached_tokens` usage field
+            // and stays unconditional: it reports what the executor reused,
+            // which is simply zero when no lookup happened.
             effects.cached.push(CachedTokensEffect {
                 request_id: req.request_id,
                 cached_tokens: result.cached_tokens,
             });
+            // The /metrics counters, by contrast, must only count requests
+            // the executor actually looked up.
+            if executor.prefix_cache_enabled() && !req.echo {
```
FeathBow: This repeats the executor's own gate (prefix_cache_enabled() && !req.echo at executor.rs:1906) in the resolver; if the executor's condition changes the counters drift silently, and the new trait method defaults to true, so an executor that forgets to override reports the phantom queries this change is meant to remove. The fact the resolver needs is produced exactly once, in schedule_prefill_chunk. I recommend PrefillRequestResult.cached_tokens: Option<usize> (None = no lookup; the usage effect uses unwrap_or(0)), or a prefix_lookup: bool set there; then ModelExecutor::prefix_cache_enabled, its permissive default, and the same-name inherent/trait pair can go, and FakeExecutor's without_prefix_cache just returns None.

FeathBow (changes requested):
Thanks for the tracker refactor and the two fixes; the delta conversion now mirrors #952 and the units are right. On your two questions:

&2: the field exists on the pinned rev. SchedulerStats has connector_prefix_cache_stats (protocol/stats.rs:183) and the metrics crate registers vllm:external_prefix_cache_queries/hits, so nothing in the git dependency needs extending. On the qwen3 side the split is available where the lookup happens: schedule_prefill_chunk runs after the prefetch settles and PrefetchState carries the probe's gpu_hit_blocks(), so local = gpu_hit_blocks x block_size (the whole match when there was no prefetch) and external = matched − local. I recommend a second counter pair on SchedulerMetrics and a second tracker stamping connector_prefix_cache_stats, with match_and_add_prefix left as is.

&4: I recommend adding the lookup plumbing to sim the way its spec-decode counters work: a synthetic prefix_hit (and an external knob) reported as cached_tokens on each request's first chunk, and a frontend_e2e case that scrapes /metrics for the four prefix counters across local hit/miss, disabled/echo, the metadata-only abort, and external attribution. The qwen3 real-lookup leg can be run on a GPU host once that is green.

Two more things before another pass: the &3 fix has no regression test yet (metadata-only cached step, then abort, asserting the prefix delta ships in a stats-only batch); and the &1 fix re-states the executor's gate (prefix_cache_enabled() && !echo) in the resolver behind a trait default of true. Reporting the lookup from the executor (cached_tokens: Option<usize> on PrefillRequestResult, None when no lookup ran) removes the duplicated predicate and the permissive default.

Non-blocking: the totals-versus-deltas explanation is written in seven places; one home on PrefixCacheTracker is enough. The adapter test could shrink to the once-per-request rule with a multi-chunk prompt, without the eprintln!s, the triple read, and the stale prefix_cache_delta reference.

1571859588:
Both remaining items are done, plus the non-blocking cleanup.

&3 regression test — aborted_request_still_ships_its_prefix_lookup: a request is admitted against a slow-prefill simulated engine (so its lookup is counted) and then abandoned by a client that gives up inside the prefill window, leaving a prefix delta with no token to carry it; a second completed request drives the next step. The four families must then report both lookups exactly once — two prompts' worth, not just the survivor's, and not a delta replayed on top of it. Mutation-checked: removing the prefix/external terms from the send condition fails this test (10s timeout), restoring them passes.

&1 — PrefillRequestResult::cached_tokens is now Option (None = no lookup, Some(0) = a miss), set where the lookup happens. ModelExecutor::prefix_cache_enabled, its permissive true default, and the resolver's copy of the prefix_cache_enabled() && !echo gate are all gone; the usage effect uses unwrap_or(0).

&2 — split at the lookup site: gpu_hit_blocks × block_size is local, the remainder external, and the whole match counts as local when the request has no prefetch probe. Second counter pair on SchedulerMetrics, second PrefixCacheTracker stamping connector_prefix_cache_stats (None on a zero interval, so a line with no connector leaves it unset). match_and_add_prefix still returns usize.

&4 — pegainfer-sim gained with_prefix_cache(local, external) modeled on its spec-decode knob, and frontend_e2e now scrapes all four families through the stepped bridge, plus a case asserting a cache-disabled engine reports no lookup. The qwen3 real-lookup leg is still unrun locally — rdma-mummy-sys doesn't build here — so it needs CI or a GPU host.

Non-blocking — the totals-versus-deltas note now lives only on PrefixCacheTracker; the other sites point at it. Adapter test trimmed: no scratch eprintln!s, no triple read, no stale prefix_cache_delta reference.

Locally: frontend 68/68, sim e2e 16/16, qwen35 + gemma4 build, rustfmt clean.

1571859588:
Rebased onto 31d35bc5 (from 8f455a18), all four review cases included. Apologies for the gap — the rebase crossed three PRs that touched the same code:

#722 had independently converted cached_tokens to Option and replaced the echo flag with prompt_logprobs. That is the same shape you asked for, so I kept upstream's version and re-laid our split on top: the executor sets Some(matched) plus external_hit_tokens = matched − local, and the resolver counts on Some (a miss counts its query; None counts nothing). The duplicated gate, ModelExecutor::prefix_cache_enabled, and its permissive default are still gone.
#978 added StopCause to the finish path — merged alongside the counter fold.
#1047 gave sim a profiled lane — the prefix counters now report on both the profiled and legacy metrics() paths.
The sim lookup knob follows the same gate as qwen3 (prompt_logprobs.is_none()), so the e2e covers the case you care about.

Verified locally on the rebased head: frontend 81/81, sim e2e 25/25 (including the three prefix cases), gemma4 and qwen35 both build clean, rustfmt clean. The qwen3 real-lookup leg still needs CI or a GPU host — rdma-mummy-sys doesn't build here.

The stale label should clear with this activity; happy to keep iterating whenever you have a pass.

1571859588:
CI fixed at 24d66f86: clippy's struct_field_names flagged PrefixCacheTracker (all four fields shared a last_ prefix) — dropped the prefix rather than allow-listing it; and cargo fmt --all --check wanted the prefetch.get(...) chain on one line. Re-verified: frontend 81/81, sim e2e 25/25, all changed files pass rustfmt --check.

xiaguan:
Thanks for sticking with this through all the rounds! I've approved the CI workflow for the current head so you can see the real results.

Could you rebase onto the latest main next? One thing to watch: the last rebase looks like it brought back `gather_decode_logprobs` / `build_prefill_request_results` in `pegainfer-qwen3/src/executor.rs`. Main already moved them to `executor/logprobs.rs`, and they're imported under the same names there, so qwen3 likely won't compile. Please drop the duplicated copies while rebasing.

Let's get CI green first, then we'll do another review pass.

1571859588:
Rebased onto latest main and dropped the duplicated helpers at 9be55e39 — thanks for the pointer, that was exactly it. The previous squash had resurrected gather_decode_logprobs / build_prefill_request_results into executor.rs alongside the logprobs.rs versions, which is where the E0061 signature mismatches, the unused import, and the stale req.echo / req.logprobs > 0 errors all came from. Removed the copies rather than adapting them.

Two real fixes left behind by that: logprobs.rs now sets external_hit_tokens on PrefillRequestResult, and its prompt-logprob guard compares cached_tokens.unwrap_or(0) (None counts as uncached). Also fixed the gpu_hit_blocks access — PrefetchState wraps probe, so it's state.probe.gpu_hit_blocks().

Locally: frontend 81/81, sim e2e 25/25, gemma4 + qwen35 build, rustfmt clean. qwen3 itself still needs CI — rdma-mummy-sys doesn't build here.

PR #966, opened 2026-08-25
refactor(qwen35): split config.rs into config/{model,tokenizer,tp}.rs

CAICAIIs:
## What

Split the 683-line `pegainfer-qwen35/src/config.rs` into a `config/` module with `model`/`tokenizer`/`tp` submodules (k3-style `config/mod.rs` entry + `config/`). `config.rs` becomes a re-export of the three submodules. **Behavior unchanged.**

## Structure

```
pegainfer-qwen35/src/config.rs  (683 → removed)
pegainfer-qwen35/src/config/
├── mod.rs        # module entry — re-exports model/tokenizer/tp
├── model.rs      # model-level config
├── tokenizer.rs  # tokenizer config
└── tp.rs         # tensor-parallel config
```

## Validation

- `cargo check` — pass
- `cargo clippy --all-targets -- -D warnings` — pass (matches CI)
- `cargo test --lib config` — 14 passed
- Draft PR; small, single-purpose slice.


xiaguan (changes requested):
Thanks for doing this cleanup. I do not think the current split reduces the cost of understanding or changing this code yet. It redistributes the old file, then recreates the same flat surface through glob re-exports; model and tensor-parallel concerns still depend on each other, tests remain centralized, and invalid tensor-parallel state is still representable after validation. This is file organization without a clear ownership model.

Please make the split establish directional boundaries and stronger invariants. A good shape would deserialize into raw config types, convert once into validated model and tensor-parallel/local-geometry types, and let downstream code accept only those validated types. Keep tests with the module that owns the behavior, use explicit exports, and either remove the tokenizer's unread compatibility schema or make that compatibility contract an explicitly owned abstraction instead of hiding it behind `allow(dead_code)`.

For the validation tests, typed error variants plus a small valid fixture with table-driven mutations would be clearer and more durable than repeated JSON and assertions over error strings. The workspace's `validator` derive is useful for local field rules, but most rules here are cross-field and kernel/TP compatibility constraints; `TryFrom` and domain types are a better boundary for them.

CAICAIIs:
> Thanks for doing this cleanup. I do not think the current split reduces the cost of understanding or changing this code yet. It redistributes the old file, then recreates the same flat surface through glob re-exports; model and tensor-parallel concerns still depend on each other, tests remain centralized, and invalid tensor-parallel state is still representable after validation. This is file organization without a clear ownership model.
> 
> Please make the split establish directional boundaries and stronger invariants. A good shape would deserialize into raw config types, convert once into validated model and tensor-parallel/local-geometry types, and let downstream code accept only those validated types. Keep tests with the module that owns the behavior, use explicit exports, and either remove the tokenizer's unread compatibility schema or make that compatibility contract an explicitly owned abstraction instead of hiding it behind `allow(dead_code)`.
> 
> For the validation tests, typed error variants plus a small valid fixture with table-driven mutations would be clearer and more durable than repeated JSON and assertions over error strings. The workspace's `validator` derive is useful for local field rules, but most rules here are cross-field and kernel/TP compatibility constraints; `TryFrom` and domain types are a better boundary for them.

Thanks for the feedback. I rebuilt the split around the boundary you described — raw config deserializes into validated model/TP types (TryFrom + typed errors), LocalGeometry replaces the old local_* calls, invalid TP state is now unrepresentable, and tests moved into the owning modules. Reworked the whole line to this shape (#966–#969); ready for another look.

PR #967, opened 2026-08-25
refactor(qwen35): split scheduler backend into scheduler/backend.rs

CAICAIIs:
## What

Split the `AsyncPrefillOutput` backend cluster (553 lines) out of the 2624-line `pegainfer-qwen35/src/scheduler.rs` into a `scheduler/backend.rs` module (k3-style `scheduler/mod.rs` entry + `scheduler/backend.rs`). The backend concern (single-GPU + TP scheduler backend) now has a single owner. **Behavior unchanged.**

## Structure

```
pegainfer-qwen35/src/scheduler/
├── mod.rs     # scheduler module entry (re-export, `mod backend`)
└── backend.rs # SchedulerBackend / SingleGpuBackend / TpSchedulerBackend
```

## Validation

- `cargo check` — pass
- `cargo clippy --all-targets -- -D warnings` — pass (matches CI)
- Draft PR; small, single-purpose slice. `scheduler.rs` 2624 → ~2074.

## Note

- Adds `#![allow(clippy::wildcard_imports)]` at crate level (the `use self::backend::*` submodule glob is the established module-tree convention).


FeathBow (commented):
2 minor: Functions such as `fatal_cuda_lifecycle`, which are used only within the backend, do not require `pub(super)`.

The backend imports everything through use super::*, while the parent imports the backend items back through use self::backend::*. CUDA/model/executor dependencies remain declared in the parent, so reading backend.rs requires tracing the parent's imports and later import cleanup can implicitly affect the child.

Please import external and sibling-module types directly where they are used, use an explicit super::{...} list for scheduler-owned state and helpers, and explicitly import the backend entry points in mod.rs. This makes the file split easier to maintain without introducing another abstraction or wrapper.

btw #968 has the same problem.

PR #978, opened 2026-08-26
fix(qwen3): preserve explicit stop-token causes

RicardoMin:
Related to #865

# Qwen3: preserve explicit stop-token causes in the stepped contract

## Why this is a separate PR

This is the scoped follow-up requested by the maintainers during review of
[#865](https://github.com/pegainfer-project/pegainfer/pull/865).  They asked
that the broad stop-contract change be split so that the shared boundary,
frontend bridges, and one model can be reviewed and validated independently.
This PR therefore extracts the Qwen3 migration from that work; Qwen3.5 and the
other model schedulers remain on their existing contract for now.

## Summary

The previous vLLM dependency update fixed several frontend compatibility issues,
but the stepped Qwen3 path still collapsed two independent controls into the
single legacy `ignore_eos` flag.  That made an explicit `stop_token_ids` request
indistinguishable from a model-EOS request and forced the bridge to guess a
synthetic stop token after the scheduler had already discarded the real one.

This PR gives Qwen3 a typed stop contract.  It preserves the sampled trigger
token and its logprob, carries the concrete `StopCause` through the scheduler
and stepped bridge, and keeps EOS handling independent from request-provided
stop IDs.

## What was wrong

The old contract exposed only `FinishReason::Stop` or `FinishReason::Length`.
When a request stopped, the bridge could not tell whether the model emitted EOS
or an explicit request stop token.  It therefore reconstructed a sentinel (EOS
first, otherwise the first configured stop ID).  That reconstruction can report
the wrong token, loses the token's logprob, and is incorrect for a speculative
span where the first terminal token is followed by additional accepted tokens.

The old boolean also could not express the valid combination "ignore model EOS,
but still stop on these explicit request token IDs".

## Contract change

| Event | Legacy behavior | Qwen3 stepped behavior |
| --- | --- | --- |
| Model EOS | `FinishReason::Stop`; trigger may be suppressed or reconstructed | `FinishReason::Stop` + `StopCause::Eos(id)`; token is retained internally; wire `stop_reason` is absent |
| Explicit `stop_token_ids` match | Often treated as ordinary output or replaced by a guessed sentinel | `FinishReason::Stop` + `StopCause::Token(id)`; actual ID is reported as wire `stop_reason` |
| `ignore_eos=true` | Could inadvertently disable explicit stops | Disables only model EOS; explicit request stops remain active |
| Length limit | `FinishReason::Length` | `FinishReason::Length` + no stop cause; final sampled token is retained |
| Speculative span | Trigger/suffix ownership was ambiguous | Commit the prefix through the first trigger and discard only the suffix after it |

EOS has precedence when the same ID is both the active EOS token and an explicit
request stop.  Completion-token accounting is incremented once, including the
trigger token.

## Scope and compatibility

Following the maintainer's scope request on #865, this PR intentionally
migrates **Qwen3 only**.  The shared request/step types
accept an optional typed cause, while the existing legacy event path remains
available for models that have not been audited.  The legacy bridge keeps its
synthetic-sentinel fallback only when an old producer supplies no typed cause.
Therefore Qwen3.5 and other model schedulers are not changed in this PR and do
not need to adopt the new resolver contract yet.  If the maintainers agree with
the semantics, the remaining model lines can be migrated one at a time with
their own lifecycle tests.

## Implementation

- Added `StopPolicy`, `EosPolicy`, and `StopCause` at the frontend engine
  boundary.
- Converted wire EOS and explicit stop fields without collapsing them into one
  boolean for the stepped path.
- Propagated the policy through Qwen3 request state, ledger updates, prefill,
  ordinary decode, and speculative verification.
- Emitted the trigger token before terminal metadata and retained its logprob.
- Mapped `StopCause::Token(id)` to the vLLM-compatible wire `stop_reason`.
- Left the legacy bridge fallback and un-migrated model implementations intact.
- Updated two existing test fixtures (K3 and simulator) only to supply the new
  shared default field; no legacy model runtime behavior is changed.

## Automated verification

| Check | Result |
| --- | --- |
| `cargo test --release -p pegainfer-frontend --lib` | 73 passed, 0 failed |
| `cargo test --release -p pegainfer-qwen3 --lib` | 93 passed, 0 failed |
| Qwen3 request-stop focused tests | 4 passed, 0 failed |
| Qwen3 speculative-stop focused tests | 2 passed, 0 failed |
| `cargo test --release -p pegainfer-sim --tests -- --test-threads=1` | 6 passed, 0 failed |
| `cargo check --release -p pegainfer-qwen35 --features qwen35` | Passed (control build only) |
| `cargo build --release -p pegainfer-server --bin pegainfer` | Passed |
| `cargo fmt --all -- --check` | Passed |
| `git diff --check` | Passed |

## HTTP A/B verification

The comparison used two already-running OpenAI-compatible endpoints on the
same validation host.  The explicit stop set covered the complete vocabulary,
so the first generated token was guaranteed to exercise the request-stop path.
This is a deterministic contract probe, not a generation-quality benchmark.

Results are shown as `finish_reason / stop_reason / completion_tokens`:

| Target | Control (`ignore_eos=true`) | Explicit stop + EOS ignored | Explicit stop + EOS enabled |
| --- | --- | --- | --- |
| Qwen3-0.6B (adapted) | `length / null / 8` | `stop / 12095 / 1` | `stop / 12095 / 1` |
| Qwen3.5-0.8B (legacy control) | `length / null / 8` | `length / null / 8` | `length / null / 8` |

| Contract check | Qwen3 adapted | Qwen3.5 legacy control |
| --- | ---: | ---: |
| Baseline control | pass | pass |
| Explicit stop with EOS ignored | pass | fail |
| Explicit stop with EOS enabled | pass | fail |
| Stop-set order invariant | pass | not satisfied (no typed stop) |
| Trigger logprob present | pass | fail |
| Streaming typed stop | pass | fail |
| Mixed controls (3/3) | 3/3 | 3/3 |
| Mixed explicit stops (3/3) | 3/3 | 0/3 |
| Overall new-contract checks | 8/8 | 2/8 |

The Qwen3.5 rows are an intentional legacy comparison: ordinary generation
still works, but its un-migrated scheduler does not yet satisfy the new typed
explicit-stop contract.  They are not a claim that every legacy model fails in
all workloads.

## Reproduction

Build and start each server independently.  Qwen3.5 requires its feature-gated
Triton build environment; it does **not** support or require a
`--gpu-memory-utilization` CLI argument.

```bash
# Qwen3 stepped path
cargo run --release -p pegainfer-server -- \
  --model-path "$QWEN3_MODEL" \
  --served-model-name qwen3-adapted \
  --port 18081

# Qwen3.5 legacy control path (set PEGAINFER_TRITON_PYTHON if needed)
cargo run --release -p pegainfer-server --features qwen35 -- \
  --model-path "$QWEN35_MODEL" \
  --served-model-name qwen35-legacy \
  --port 18082
```

Then run the attached script (Python standard library only):

```bash
python3 pr865_qwen3_stop_contract_ab.py \
  --qwen3-url http://127.0.0.1:18081 \
  --qwen3-model qwen3-adapted \
  --qwen35-url http://127.0.0.1:18082 \
  --qwen35-model qwen35-legacy \
  --out stop-contract-ab.json
```

The script prints a compact comparison table and writes machine-readable JSON.
Use `--stop-token-id ID` to replace the full-vocabulary deterministic set with
a single known token when reproducing on a different prompt/model pair.

## Follow-up

As requested during review of #865, this PR deliberately stops at the Qwen3
migration boundary.  After the
maintainers confirm that the independent EOS/request-stop semantics are wanted,
the same policy propagation and resolver audit can be applied to Qwen3.5 and the
other legacy model lines in separate, model-scoped changes.

[pr865_qwen3_stop_contract_ab.py](https://github.com/user-attachments/files/31464004/pr865_qwen3_stop_contract_ab.py)

### DIFF
--- a/docs/subsystems/frontend/frontend-architecture.md
+++ b/docs/subsystems/frontend/frontend-architecture.md
@@ -12,14 +12,19 @@ An engine is a set of schedulers, each a `Scheduler` implementation driven by th
 
 ```
 pegainfer-frontend/src/engine/
-├── step.rs              # the wire: RequestId, Request, QueuedRequest,
-│                        #   StepOutputs { Vec<RequestUpdate> },
-│                        #   RequestUpdate { scheduled, tokens, logprobs, cached_tokens,
-│                        #   prompt_echo, kv_transfer, terminal }, Terminal
-├── request_lifecycle.rs # submission envelope, abort control and step sender plumbing;
-│                        #   DeferredFinish remains available for P/D handoff
-├── ledger.rs            # RequestLedger: admit/reject/push/finish/fail/retire,
-│                        #   prompt/completion tallies, one merged update per touched id
+├── step.rs              # the wire: RequestId, Request, StepOutputs { Vec<RequestUpdate> },
+│                        #   Request { ..., stop_policy }, RequestUpdate { scheduled, tokens,
+│                        #   logprobs, cached_tokens, prompt_echo, kv_transfer, terminal },
+│                        #   Terminal { ..., stop_cause }
+├── request_lifecycle.rs # typestate handles: QueuedRequest ─admit→
+│                        #   ActiveRequest ─finish/fail/defer→ consumed; every
+│                        #   transition is by-move, a dropped handle emits Failed
+│                        #   (drop bomb); DeferredFinish; RequestControl
+├── emitter.rs           # StepEmitter: the single writer of the per-step buffer; stamps
+│                        #   timestamps, tallies prompt/completion counts, folds each
+│                        #   request's step into one RequestUpdate; commit_step sends once
+├── ledger.rs             # RequestLedger: admit/reject/push/finish/fail/retire,
+│                         #   prompt/completion tallies, one merged update per touched id
 ├── wiring.rs            # scheduler_pair, SchedulerHandle (submit/take_steps/load),
 │                        #   Engine { schedulers, info, lora }, LiveScheduler,
 │                        #   EngineInfo, LaunchedEngine { Handle | Stepped }
@@ -32,10 +37,12 @@ pegainfer-frontend/src/engine/
 Design decisions worth knowing before touching it:
 
 - **Step-batched wire.** One message per scheduler step, not one channel per request: the scheduler's natural output unit is the step batch, and per-request channels were tried and rejected (the scheduler for-loop over N channels was the bottleneck). The protocol stack demuxes.
-- **Flat `RequestUpdate`.** All facts a step produced for one request travel in one struct, so intra-request ordering is structure, not convention. The ledger merges every write for an id into that one record before the driver commits the step.
-- **Ledger lifecycle.** `RequestLedger` owns one account for every unanswered submission. A scheduler receives only `QueuedRequest { id, request }` and writes every lifecycle transition by id. The ledger rejects touches after closure, derives completion counts from `push_tokens`, and writes off every open account when an engine-fatal `step` error ends the driver.
-- **Ledger as single writer.** Schedulers never touch the step channel; they call ledger methods. `admit` stamps `ScheduledInfo` from the registered prompt length, `push_tokens` tallies completions, and `commit_step` publishes the merged statement once per driver iteration.
-- **Polling driver, scheduler-owned park.** `spawn_scheduler` owns the serve loop: drain submissions, `Scheduler::step`, publish metrics, commit. An idle iteration ends in a `spin_loop` hint. Gemma 4 has one deliberate park inside this policy: while async prefill is the only remaining work, its scheduler drains the lane and joins it rather than hot-polling the completion (a drain failure is engine-fatal); when decode or queued work exists it keeps polling the lane without blocking.
+- **Flat `RequestUpdate`.** All facts a step produced for one request travel in one struct, so intra-request ordering is structure, not convention. This is what makes `defer_finish` safe: a P/D prefill executor can withhold a request's `Finished` until its KV saves are peer-visible and send it later from any thread — the deferred message carries the request's entire buffered update, so late delivery cannot reorder.
+- **Independent stop policy and cause.** `Request.stop_policy` keeps model EOS handling separate from explicit request stop IDs. A token-driven finish retains the triggering token in `RequestUpdate` and carries `Terminal::Finished.stop_cause`; decode paths include its real logprob when available, so the stepped bridge can report the actual explicit stop ID without reconstructing it. `ignore_eos` affects only model EOS. Legacy producers may leave the cause empty while they are migrated individually.
+- **Typestate lifecycle.** `QueuedRequest` (queued) and `ActiveRequest` (streaming) are owned tokens; admit/reject/retire/finish/fail consume them, so "terminal exactly once, nothing after it" cannot be miscoded — it does not compile. A handle dropped without a transition emits `Failed` from its `Drop`, which is also how a crashed scheduler answers every in-flight request: the driver drops the scheduler, the handles fall, the terminals ship.
+- **Emitter as single writer.** Schedulers never touch the channel; they call `StepEmitter` methods against their handles. The emitter stamps `ScheduledInfo` at admission, tallies token counts (terminal counts derive from the tally, never from model-side arithmetic), and `commit_step` publishes the whole step in one send.
+- **Pure polling driver.** `spawn_scheduler` owns the serve loop: drain submissions, `Scheduler::step`, publish load, commit. No idle/park distinction — the scheduler owns the GPU and spinning on it costs nothing anyone else could use; async KV I/O (prefetch, decode-overlap prefill) is naturally absorbed by polling. An idle iteration ends in a `spin_loop` hint (relaxes the core's issue slots, no latency cost — busy iterations never pause). The loop exits when the frontend drops the handle and the queue drains.
+- **Gemma 4 async prefill exception.** While asynchronous prefill is the only remaining work, Gemma 4 drains and joins that lane rather than hot-polling its completion; decode or queued work keeps the normal polling path.
 - **Abort is a flag, not channel teardown.** `SchedulerHandle::submit` returns a `RequestControl`; the frontend flips its boolean abort flag and the scheduler retires the request silently on its next touch (no terminal — the frontend already dropped its state for that id).
 - **Channels:** the submit channel is crossbeam (sync consumer on the scheduler thread), steps are tokio mpsc (async consumer in the bridge); load is a shared cell read via `SchedulerHandle::load()` — pull-only by design, "notify me on load change" is deliberately unrepresentable (the driver busy-polls, so a subscription edge would fire per spin). All channels unbounded on purpose — admission control is the scheduler's job, expressed as `Rejected`, never as backpressure on submit.
 - **Control plane lives outside the contract.** `Scheduler` has no control method and the contract carries no control channel. A capability like LoRA is a private channel the model crate mints *before* `spawn_scheduler` — the scheduler closes over the receiver, the `LoraClient` sender surfaces as `Engine.lora: Option<LoraClient>`, and the `Option` *is* the capability (no `bool` flag, no registry until a second capability exists). The vocabulary (`LoraControl`, `LoraClient`) is still defined in the frontend crate because the frontend must speak it without holding model structs; only the wiring is the model's business.
@@ -91,7 +98,7 @@ All six lines are onboarded. Adding a model line = write `model_line.rs` in the 
 
 ## Protocol stacks
 
-**`vllm` (current default, fleet-proven).** Impersonates a vLLM EngineCore process over in-process ZMQ/msgpack because upstream `vllm-server` assumes the engine is a separate process. HTTP routes, OpenAI types, tokenizer, chat templates, Prometheus live in the external `vllm-server`/`vllm-metrics`/`vllm-text` crates. `SteppedEngineBridge` translates each `RequestUpdate` 1:1 into an EngineCore output (wall-clock timestamps are reconstructed from the contract's `Instant`s via a per-bridge unix anchor; a `Finished{Stop}` appends the stop sentinel token, which is how usage keeps counting the suppressed EOS).
+**`vllm` (current default, fleet-proven).** Impersonates a vLLM EngineCore process over in-process ZMQ/msgpack because upstream `vllm-server` assumes the engine is a separate process. HTTP routes, OpenAI types, tokenizer, chat templates, Prometheus live in the external `vllm-server`/`vllm-metrics`/`vllm-text` crates. `SteppedEngineBridge` translates each `RequestUpdate` 1:1 into an EngineCore output (wall-clock timestamps are reconstructed from the contract's `Instant`s via a per-bridge unix anchor; a typed `StopCause` reports the actual request stop token, while the synthetic sentinel remains only as a compatibility fallback for producers that provide no cause).
 
 **`dynamo` (planned second stack).** dynamo's `lib/llm` in-process path removes the wire protocol entirely (`EngineConfig::InProcessTokens` + `run_input`). The step contract was shaped so this stack can consume `StepOutputs` directly without impersonation overhead. Decision gate: prototype, A/B against the vllm stack, let TTFT/step-overhead numbers pick the default.
 
--- a/docs/subsystems/frontend/frontend-architecture.md
+++ b/docs/subsystems/frontend/frontend-architecture.md
@@ -12,19 +12,14 @@ An engine is a set of schedulers, each a `Scheduler` implementation driven by th
 
 ```
 pegainfer-frontend/src/engine/
-├── step.rs              # the wire: RequestId, Request, StepOutputs { Vec<RequestUpdate> },
-│                        #   Request { ..., stop_policy }, RequestUpdate { scheduled, tokens,
-│                        #   logprobs, cached_tokens, prompt_echo, kv_transfer, terminal },
-│                        #   Terminal { ..., stop_cause }
-├── request_lifecycle.rs # typestate handles: QueuedRequest ─admit→
-│                        #   ActiveRequest ─finish/fail/defer→ consumed; every
-│                        #   transition is by-move, a dropped handle emits Failed
-│                        #   (drop bomb); DeferredFinish; RequestControl
-├── emitter.rs           # StepEmitter: the single writer of the per-step buffer; stamps
-│                        #   timestamps, tallies prompt/completion counts, folds each
-│                        #   request's step into one RequestUpdate; commit_step sends once
-├── ledger.rs             # RequestLedger: admit/reject/push/finish/fail/retire,
-│                         #   prompt/completion tallies, one merged update per touched id
+├── step.rs              # the wire: RequestId, Request, QueuedRequest,
+│                        #   StepOutputs { Vec<RequestUpdate> },
+│                        #   RequestUpdate { scheduled, tokens, logprobs, cached_tokens,
+│                        #   prompt_echo, kv_transfer, terminal }, Terminal
+├── request_lifecycle.rs # submission envelope, abort control and step sender plumbing;
+│                        #   DeferredFinish remains available for P/D handoff
+├── ledger.rs            # RequestLedger: admit/reject/push/finish/fail/retire,
+│                        #   prompt/completion tallies, one merged update per touched id
 ├── wiring.rs            # scheduler_pair, SchedulerHandle (submit/take_steps/load),
 │                        #   Engine { schedulers, info, lora }, LiveScheduler,
 │                        #   EngineInfo, LaunchedEngine { Handle | Stepped }
@@ -37,12 +32,10 @@ pegainfer-frontend/src/engine/
 Design decisions worth knowing before touching it:
 
 - **Step-batched wire.** One message per scheduler step, not one channel per request: the scheduler's natural output unit is the step batch, and per-request channels were tried and rejected (the scheduler for-loop over N channels was the bottleneck). The protocol stack demuxes.
-- **Flat `RequestUpdate`.** All facts a step produced for one request travel in one struct, so intra-request ordering is structure, not convention. This is what makes `defer_finish` safe: a P/D prefill executor can withhold a request's `Finished` until its KV saves are peer-visible and send it later from any thread — the deferred message carries the request's entire buffered update, so late delivery cannot reorder.
-- **Independent stop policy and cause.** `Request.stop_policy` keeps model EOS handling separate from explicit request stop IDs. A token-driven finish retains the triggering token in `RequestUpdate` and carries `Terminal::Finished.stop_cause`; decode paths include its real logprob when available, so the stepped bridge can report the actual explicit stop ID without reconstructing it. `ignore_eos` affects only model EOS. Legacy producers may leave the cause empty while they are migrated individually.
-- **Typestate lifecycle.** `QueuedRequest` (queued) and `ActiveRequest` (streaming) are owned tokens; admit/reject/retire/finish/fail consume them, so "terminal exactly once, nothing after it" cannot be miscoded — it does not compile. A handle dropped without a transition emits `Failed` from its `Drop`, which is also how a crashed scheduler answers every in-flight request: the driver drops the scheduler, the handles fall, the terminals ship.
-- **Emitter as single writer.** Schedulers never touch the channel; they call `StepEmitter` methods against their handles. The emitter stamps `ScheduledInfo` at admission, tallies token counts (terminal counts derive from the tally, never from model-side arithmetic), and `commit_step` publishes the whole step in one send.
-- **Pure polling driver.** `spawn_scheduler` owns the serve loop: drain submissions, `Scheduler::step`, publish load, commit. No idle/park distinction — the scheduler owns the GPU and spinning on it costs nothing anyone else could use; async KV I/O (prefetch, decode-overlap prefill) is naturally absorbed by polling. An idle iteration ends in a `spin_loop` hint (relaxes the core's issue slots, no latency cost — busy iterations never pause). The loop exits when the frontend drops the handle and the queue drains.
-- **Gemma 4 async prefill exception.** While asynchronous prefill is the only remaining work, Gemma 4 drains and joins that lane rather than hot-polling its completion; decode or queued work keeps the normal polling path.
+- **Flat `RequestUpdate`.** All facts a step produced for one request travel in one struct, so intra-request ordering is structure, not convention. The ledger merges every write for an id into that one record before the driver commits the step.
+- **Ledger lifecycle.** `RequestLedger` owns one account for every unanswered submission. A scheduler receives only `QueuedRequest { id, request }` and writes every lifecycle transition by id. The ledger rejects touches after closure, derives completion counts from `push_tokens`, and writes off every open account when an engine-fatal `step` error ends the driver.
+- **Ledger as single writer.** Schedulers never touch the step channel; they call ledger methods. `admit` stamps `ScheduledInfo` from the registered prompt length, `push_tokens` tallies completions, and `commit_step` publishes the merged statement once per driver iteration.
+- **Polling driver, scheduler-owned park.** `spawn_scheduler` owns the serve loop: drain submissions, `Scheduler::step`, publish metrics, commit. An idle iteration ends in a `spin_loop` hint. Gemma 4 has one deliberate park inside this policy: while async prefill is the only remaining work, its scheduler drains the lane and joins it rather than hot-polling the completion (a drain failure is engine-fatal); when decode or queued work exists it keeps polling the lane without blocking.
 - **Abort is a flag, not channel teardown.** `SchedulerHandle::submit` returns a `RequestControl`; the frontend flips its boolean abort flag and the scheduler retires the request silently on its next touch (no terminal — the frontend already dropped its state for that id).
 - **Channels:** the submit channel is crossbeam (sync consumer on the scheduler thread), steps are tokio mpsc (async consumer in the bridge); load is a shared cell read via `SchedulerHandle::load()` — pull-only by design, "notify me on load change" is deliberately unrepresentable (the driver busy-polls, so a subscription edge would fire per spin). All channels unbounded on purpose — admission control is the scheduler's job, expressed as `Rejected`, never as backpressure on submit.
 - **Control plane lives outside the contract.** `Scheduler` has no control method and the contract carries no control channel. A capability like LoRA is a private channel the model crate mints *before* `spawn_scheduler` — the scheduler closes over the receiver, the `LoraClient` sender surfaces as `Engine.lora: Option<LoraClient>`, and the `Option` *is* the capability (no `bool` flag, no registry until a second capability exists). The vocabulary (`LoraControl`, `LoraClient`) is still defined in the frontend crate because the frontend must speak it without holding model structs; only the wiring is the model's business.
@@ -98,7 +91,7 @@ All six lines are onboarded. Adding a model line = write `model_line.rs` in the 
 
 ## Protocol stacks
 
-**`vllm` (current default, fleet-proven).** Impersonates a vLLM EngineCore process over in-process ZMQ/msgpack because upstream `vllm-server` assumes the engine is a separate process. HTTP routes, OpenAI types, tokenizer, chat templates, Prometheus live in the external `vllm-server`/`vllm-metrics`/`vllm-text` crates. `SteppedEngineBridge` translates each `RequestUpdate` 1:1 into an EngineCore output (wall-clock timestamps are reconstructed from the contract's `Instant`s via a per-bridge unix anchor; a typed `StopCause` reports the actual request stop token, while the synthetic sentinel remains only as a compatibility fallback for producers that provide no cause).
+**`vllm` (current default, fleet-proven).** Impersonates a vLLM EngineCore process over in-process ZMQ/msgpack because upstream `vllm-server` assumes the engine is a separate process. HTTP routes, OpenAI types, tokenizer, chat templates, Prometheus live in the external `vllm-server`/`vllm-metrics`/`vllm-text` crates. `SteppedEngineBridge` translates each `RequestUpdate` 1:1 into an EngineCore output (wall-clock timestamps are reconstructed from the contract's `Instant`s via a per-bridge unix anchor; a `Finished{Stop}` appends the stop sentinel token, which is how usage keeps counting the suppressed EOS).
 
 **`dynamo` (planned second stack).** dynamo's `lib/llm` in-process path removes the wire protocol entirely (`EngineConfig::InProcessTokens` + `run_input`). The step contract was shaped so this stack can consume `StepOutputs` directly without impersonation overhead. Decision gate: prototype, A/B against the vllm stack, let TTFT/step-overhead numbers pick the default.
 
--- a/docs/subsystems/frontend/frontend-architecture.md
+++ b/docs/subsystems/frontend/frontend-architecture.md
@@ -1,6 +1,6 @@
 # Frontend architecture: pegainfer-frontend and the engine boundary
 
-**TL;DR:** `pegainfer-frontend` owns everything north of the model schedulers: the engine contract, the vLLM protocol stack, and the `ModelLine` dispatch trait. The contract now has two generations living side by side: the **step contract** (`StepOutputs` wire + `RequestLedger` lifecycle + a contract-owned polling driver — Qwen3, Gemma 4 and `pegainfer-sim` are migrated) and the **legacy handle contract** (`EngineHandle` + `TokenEvent` per-request events — glm52/qwen35/kimi-k2/deepseek-v2-lite still launch through it). **Next step: migrate glm52, then delete the legacy contract.**
+**TL;DR:** `pegainfer-frontend` owns everything north of the model schedulers: the engine contract, the vLLM protocol stack, and the `ModelLine` dispatch trait. The contract now has two generations living side by side: the **step contract** (`StepOutputs` wire + `RequestLedger` lifecycle + a contract-owned polling driver — Qwen3, Gemma 4, K3 and `pegainfer-sim` are migrated) and the **legacy handle contract** (`EngineHandle` + `TokenEvent` per-request events — glm52/qwen35/kimi-k2/deepseek-v2-lite still launch through it). **Next step: migrate glm52, then delete the legacy contract.**
 
 Last touched: 2026-09
 
@@ -16,13 +16,8 @@ pegainfer-frontend/src/engine/
 │                        #   Request { ..., stop_policy }, RequestUpdate { scheduled, tokens,
 │                        #   logprobs, cached_tokens, prompt_echo, kv_transfer, terminal },
 │                        #   Terminal { ..., stop_cause }
-├── request_lifecycle.rs # typestate handles: QueuedRequest ─admit→
-│                        #   ActiveRequest ─finish/fail/defer→ consumed; every
-│                        #   transition is by-move, a dropped handle emits Failed
-│                        #   (drop bomb); DeferredFinish; RequestControl
-├── emitter.rs           # StepEmitter: the single writer of the per-step buffer; stamps
-│                        #   timestamps, tallies prompt/completion counts, folds each
-│                        #   request's step into one RequestUpdate; commit_step sends once
+├── request_lifecycle.rs # submission envelope, abort control and step sender plumbing;
+│                        #   DeferredFinish remains available for P/D handoff
 ├── ledger.rs             # RequestLedger: admit/reject/push/finish/fail/retire,
 │                         #   prompt/completion tallies, one merged update per touched id
 ├── wiring.rs            # scheduler_pair, SchedulerHandle (submit/take_steps/load),
@@ -39,8 +34,7 @@ Design decisions worth knowing before touching it:
 - **Step-batched wire.** One message per scheduler step, not one channel per request: the scheduler's natural output unit is the step batch, and per-request channels were tried and rejected (the scheduler for-loop over N channels was the bottleneck). The protocol stack demuxes.
 - **Flat `RequestUpdate`.** All facts a step produced for one request travel in one struct, so intra-request ordering is structure, not convention. This is what makes `defer_finish` safe: a P/D prefill executor can withhold a request's `Finished` until its KV saves are peer-visible and send it later from any thread — the deferred message carries the request's entire buffered update, so late delivery cannot reorder.
 - **Independent stop policy and cause.** `Request.stop_policy` keeps model EOS handling separate from explicit request stop IDs. A token-driven finish retains the triggering token in `RequestUpdate` and carries `Terminal::Finished.stop_cause`; decode paths include its real logprob when available, so the stepped bridge can report the actual explicit stop ID without reconstructing it. `ignore_eos` affects only model EOS. Legacy producers may leave the cause empty while they are migrated individually.
-- **Typestate lifecycle.** `QueuedRequest` (queued) and `ActiveRequest` (streaming) are owned tokens; admit/reject/retire/finish/fail consume them, so "terminal exactly once, nothing after it" cannot be miscoded — it does not compile. A handle dropped without a transition emits `Failed` from its `Drop`, which is also how a crashed scheduler answers every in-flight request: the driver drops the scheduler, the handles fall, the terminals ship.
-- **Emitter as single writer.** Schedulers never touch the channel; they call `StepEmitter` methods against their handles. The emitter stamps `ScheduledInfo` at admission, tallies token counts (terminal counts derive from the tally, never from model-side arithmetic), and `commit_step` publishes the whole step in one send.
+- **Ledger lifecycle.** Schedulers carry plain `RequestId`s and mutate `RequestLedger`; `RequestEnvelope` answers submissions dropped before registration, while `DeferredFinish` preserves a complete buffered update when a terminal is delivered after a P/D handoff. `RequestControl` is the frontend abort flag. The ledger remains the single writer for admission, token accounting, terminal transitions, and step publication.
 - **Pure polling driver.** `spawn_scheduler` owns the serve loop: drain submissions, `Scheduler::step`, publish load, commit. No idle/park distinction — the scheduler owns the GPU and spinning on it costs nothing anyone else could use; async KV I/O (prefetch, decode-overlap prefill) is naturally absorbed by polling. An idle iteration ends in a `spin_loop` hint (relaxes the core's issue slots, no latency cost — busy iterations never pause). The loop exits when the frontend drops the handle and the queue drains.
 - **Gemma 4 async prefill exception.** While asynchronous prefill is the only remaining work, Gemma 4 drains and joins that lane rather than hot-polling its completion; decode or queued work keeps the normal polling path.
 - **Abort is a flag, not channel teardown.** `SchedulerHandle::submit` returns a `RequestControl`; the frontend flips its boolean abort flag and the scheduler retires the request silently on its next touch (no terminal — the frontend already dropped its state for that id).
--- a/pegainfer-frontend/src/engine/stop.rs
+++ b/pegainfer-frontend/src/engine/stop.rs
@@ -1,0 +1,104 @@
+/// How a request treats end-of-sequence tokens.
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
+pub enum EosPolicy {
+    /// Do not stop on model EOS tokens.
+    Ignore,
+    /// Use the model executor's configured EOS set.
+    #[default]
+    ModelDefault,
+    /// Stop only on this protocol-provided primary EOS token.
+    Token(u32),
+}
+
+/// Request-scoped token stopping policy.
+///
+/// EOS is kept separate from caller stop tokens because the vLLM protocol
+/// reports them differently: EOS has no 'stop_reason', while a request stop
+/// reports the actual matching token ID.
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
+pub struct StopPolicy {
+    pub eos: EosPolicy,
+    pub token_ids: Vec<u32>,
+}
+
+impl StopPolicy {
+    /// Classify a token using vLLM's priority: EOS first, then the request's
+    /// explicit stop-token set.
+    #[must_use]
+    pub fn classify(
+        &self,
+        token_id: u32,
+        is_model_eos: impl FnOnce(u32) -> bool,
+    ) -> Option<StopCause> {
+        let is_eos = match self.eos {
+            EosPolicy::Ignore => false,
+            EosPolicy::ModelDefault => is_model_eos(token_id),
+            EosPolicy::Token(eos_token_id) => token_id == eos_token_id,
+        };
+
+        if is_eos {
+            Some(StopCause::Eos(token_id))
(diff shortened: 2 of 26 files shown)


FeathBow (commented):
Thanks. Please first rebase this PR onto the current main and handle CI error :)

RicardoMin:
Hi @FeathBow, I have rebased this PR onto the latest `main` and addressed the CI error you mentioned. Could you please take another look when you have a chance? Thank you!

pegainfer-qwen3/src/executor.rs
```diff
@@ -3791,8 +3801,13 @@ impl LocalQwen3Lane {
                 .flat_map(|req| std::iter::repeat_n(&req.params, req.as_slice().len()))
                 .collect();
             let target_tokens = self.select_step_tokens(bufs.all_logits(), &params, sample_seed)?;
-            let request_results = build_verify_results(requests, &target_tokens)?;
-
+            let mut request_results = build_verify_results(requests, &target_tokens)?;
```
FeathBow, line 3881: A successful `try_execute_hedged_verify()` returns here before the new truncation block below. Inside the helper, A/B chains are compared by their untruncated `accepted_tokens.len()`, then the selected chain's KV pages and hidden rows are copied, and the untruncated result is passed to `record_verify_dflash_context()`. That appends `accepted_tokens.len()` hidden rows, updates acceptance totals, and ticks the auto-hedge controller from the same lengths.

The executor-side idempotent truncation in `executor/spec.rs` still limits the eventual returned span and committed KV length, but it cannot undo the worker's winner choice or those DFlash/controller state updates. For example, a raw-longer B chain that reaches an explicit stop before A can win here even though B becomes shorter once the request contract is applied.

Please pass `stop_policies` into `try_execute_hedged_verify()`, apply the originating request's policy to every A/B candidate before comparing lengths, and use only the truncated winner for copy/context/accounting. Add a `PEGAINFER_SPEC_HEDGE` gate with an explicit stop or EOS in the middle of a candidate span; the current hedge losslessness gate never combines hedging with a terminal policy.

pegainfer-frontend/src/vllm/wire.rs
```diff
@@ -76,7 +78,23 @@ pub(crate) fn convert_sampling(params: &EngineCoreSamplingParams) -> SamplingPar
...
+        // protocol's optional primary ID only tells us whether EOS is active;
+        // using it as a singleton would miss secondary model EOS IDs.
+        eos: params
+            .eos_token_id
+            .map_or(EosPolicy::Ignore, |_| EosPolicy::ModelDefault),
+        token_ids: params.stop_token_ids.clone(),
+    }
+}
+
+/// Reject request parameters the frontend cannot represent faithfully.
```
FeathBow, line 93: The pinned vLLM wire contract defines `min_tokens` as the minimum generation count before EOS or stop-token handling. This validator does not check it, `StopPolicy` carries no threshold, and the Qwen3 resolver classifies tokens starting with the first generated token. A request with `min_tokens=8` whose first token matches an explicit stop is therefore accepted and ends after one token instead of eight.

The broad parent PR #865 already had the minimal fail-early behavior, but it was lost when #978 was split out. Please restore the nonzero rejection and add it to the existing `unsupported_request_params` unit-test table. Both bridges call this shared validator before submission, and the stepped bridge already has generic unsupported-rejection coverage, so separate legacy/stepped `min_tokens` tests would mostly duplicate the same branch. Support can be enabled later when the sampler actually masks EOS/stop IDs through the threshold.

docs/subsystems/frontend/frontend-architecture.md
```diff
@@ -12,14 +12,19 @@ An engine is a set of schedulers, each a `Scheduler` implementation driven by th
...
-│                        #   prompt/completion tallies, one merged update per touched id
+├── step.rs              # the wire: RequestId, Request, StepOutputs { Vec<RequestUpdate> },
+│                        #   Request { ..., stop_policy }, RequestUpdate { scheduled, tokens,
+│                        #   logprobs, cached_tokens, prompt_echo, kv_transfer, terminal },
+│                        #   Terminal { ..., stop_cause }
+├── request_lifecycle.rs # typestate handles: QueuedRequest ─admit→
+│                        #   ActiveRequest ─finish/fail/defer→ consumed; every
+│                        #   transition is by-move, a dropped handle emits Failed
+│                        #   (drop bomb); DeferredFinish; RequestControl
+├── emitter.rs           # StepEmitter: the single writer of the per-step buffer; stamps
```
FeathBow: There is no `engine/emitter.rs`, `StepEmitter`, or lifecycle `ActiveRequest` at this head. More importantly, `request_lifecycle.rs:3-5` explicitly documents the opposite contract: schedulers carry plain `RequestId`s and mutate `RequestLedger`; no per-request typestate handle crosses the `Scheduler` trait. Terminal-on-drop belongs to `RequestEnvelope`/`DeferredFinish`, while the ledger remains the lifecycle authority for active requests.

Please remove the unrelated typestate/emitter design text and retain only this PR's accurate `StopPolicy`/`StopCause` updates, using the existing ledger-based layout and lifecycle description.

pegainfer-frontend/src/engine/stop.rs
```diff
@@ -0,0 +1,94 @@
...
+    ) -> Option<StopCause> {
+        let is_eos = match self.eos {
+            EosPolicy::Ignore => false,
+            EosPolicy::ModelDefault => is_model_eos(token_id),
+            EosPolicy::Token(eos_token_id) => token_id == eos_token_id,
+        };
+
+        if is_eos {
+            Some(StopCause::Eos(token_id))
+        } else if self.token_ids.contains(&token_id) {
```
FeathBow: This performs a linear scan for every classified emitted/accepted token. The PR accepts arbitrarily large stop sets and its own HTTP probe uses a complete vocabulary. In local microbenchmark, a miss in a 151,935-ID `Vec` costs 10.3–13.3 us, versus 26–27 ns for a same-size `BTreeSet`; the common one-ID `Vec` costs only 2.0–2.5 ns, so replacing every request with a tree would regress the normal case.

The same large vector is also deep-copied on every speculative step when the plan is built, when `StepCommand` is created, and when that command is cloned into the primary worker. At 151,935 `u32`s, each copy is about 594 KiB before allocator overhead.

Please normalize once at the wire boundary into a sorted, deduplicated shared slice (for example `Arc<[u32]>`), use binary search, and pass cheap shared clones through the plan/worker command. Verify the common zero/one-ID case alongside the full-vocabulary case; there is no need for a more elaborate collection abstraction unless those measurements require it.

pegainfer-frontend/src/engine/stop.rs
```diff
@@ -0,0 +1,94 @@
+/// How a request treats end-of-sequence tokens.
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
+pub enum EosPolicy {
+    /// Do not stop on model EOS tokens.
+    Ignore,
+    /// Use the model executor's configured EOS set.
+    #[default]
+    ModelDefault,
+    /// Stop only on this protocol-provided primary EOS token.
+    Token(u32),
```
FeathBow: No production path at this head constructs `EosPolicy::Token`: `convert_stop_policy()` maps an enabled EOS policy to `ModelDefault` so Qwen3 can honor the model's complete EOS set. The only other occurrence is the unit test that selects this otherwise unreachable variant. Keeping a public enum branch, classifier arm, and test for a discarded parent-PR design expands the contract without serving this PR's invariant.

Please remove `Token(u32)` and exercise the EOS-vs-explicit-stop priority test through the real `ModelDefault` path and its `is_model_eos` predicate. A future protocol that genuinely needs a singleton EOS policy can add it with a production consumer then.

pegainfer-qwen3/src/scheduler/test_support.rs
```diff
@@ -260,13 +297,83 @@ impl ModelExecutor for FakeExecutor {
...
+                    DraftRequestResult {
+                        request_id: req.request_id,
+                        token_ids,
+                    }
+                })
+                .collect(),
+        })
+    }
+
+    fn execute_speculative_verify(&mut self, plan: VerifyPlan<'_>) -> Result<VerifyResult> {
```
FeathBow: This fake draft/verify mode, its state, and its builder are used by exactly one new adapter test. That test supplies preselected accepted tokens, repeats the explicit-stop span result already covered in `scheduler/tests.rs:662-688`, and checks retirement/KV metrics through the pre-existing speculative-finish cleanup branch. It never enters `Qwen3Executor`, DFlash context recording, worker KV/hidden copying, or the hedge controller, so it cannot catch the P1 defect in this PR.

Please remove this fake speculative mode and its single `frontend_adapter` consumer. Spend that coverage budget on the requested `PEGAINFER_SPEC_HEDGE` + mid-span stop/EOS production-path gate, which exercises the invariant actually changed here. The prefill/decode adapter tests and bridge mapping tests can remain: they cover distinct project-owned branches and preserve the triggering token/logprob contract.

pegainfer-qwen3/src/scheduler/resolve.rs
```diff
@@ -13,6 +14,24 @@ use crate::executor::ModelExecutor;
...
+    executor: &impl ModelExecutor,
+    req: &ActiveRequestState,
+    token: u32,
+) -> Option<StopCause> {
+    req.stop_policy
+        .classify(token, |token_id| executor.is_stop_token(token_id))
+}
+
+fn pending_stop_cause(
+    executor: &impl ModelExecutor,
```
FeathBow: These helpers differ only in the request-state type; both read `req.stop_policy` and call the same classifier. Pass `&StopPolicy` to one small helper (or call `classify` directly at the three sites) instead of maintaining parallel wrappers for request shapes. This keeps the new policy logic in one place without introducing a trait solely to share one field.

pegainfer-qwen3/src/executor/spec.rs
```diff
@@ -191,3 +228,42 @@ impl Qwen3Executor {
...
+            accepted_tokens: vec![5, 7, 8, 9],
+        };
+
+        truncate_after_terminal(&mut result, &policy, &[99]);
+
+        assert_eq!(result.accepted_tokens, vec![5, 7]);
+        assert_eq!(result.matched_draft_tokens, 2);
+    }
+
+    #[test]
```
FeathBow: Neither test exercises a KV commit despite the first test's name; both call the same pure `Vec::truncate` helper and differ only in whether `StopPolicy::classify` recognizes an explicit stop or model EOS. Classification is already covered in `engine/stop.rs`.

Please keep one accurately named test for the executor-owned behavior — retaining the trigger and capping `matched_draft_tokens` — and remove the second classification variant. KV/hidden commit correctness belongs in the real hedge + terminal gate requested above, not in a helper test whose name implies effects it never reaches.

FeathBow (changes requested):
Thanks for separating the Qwen3 stop contract and for preserving the trigger token, logprob, completion count, EOS priority, and legacy fallback. 

First, a successful hedged DFlash verify returns before terminal truncation. It selects and copies a winner, advances DFlash hidden context, updates acceptance accounting, and ticks the hedge controller from the untruncated span. The later executor truncation protects the final returned/KV length, but cannot undo those worker-side decisions. Apply each request's stop policy to every A/B candidate before winner comparison and add a hedge + mid-span terminal gate.

Second, nonzero `min_tokens` is silently accepted even though the new policy starts EOS/explicit-stop classification at token one. Restore the fail-early rejection until sampler-side masking exists; one assertion in the shared validator test is sufficient because both bridges invoke that validator before submission.

Please also revert the frontend architecture text that describes nonexistent `ActiveRequest`/`StepEmitter` code. Finally, normalize and share large stop sets: the current `Vec::contains` is linear in a per-token path and the full vector is deep-copied three times per speculative step. A shared sorted slice with binary search removes the request-size-linear scan and bulk copies; verify both the common zero/one-ID and full-vocabulary shapes before choosing anything more elaborate.

Keep the patch narrow: remove the unconstructed `EosPolicy::Token` branch, collapse the two identical resolver wrappers, and replace the one-off fake speculative executor/test with the missing production hedge gate. The existing bridge mapping and prefill/decode logprob tests cover distinct local contracts and should remain.

RicardoMin:
## Summary

This update narrows the stop-contract change to Qwen3 while retaining the legacy frontend path for models that have not migrated yet.

The previous implementation could classify a terminal token after speculative candidate selection and KV-state preparation. When a stop token appeared in the middle of a verify span, the accepted suffix could therefore remain visible in the candidate state or be copied back before truncation.

This update applies terminal truncation before hedge winner selection and before speculative KV/state commit.

## Stop Contract

| Condition | `finish_reason` | Internal cause | Wire `stop_reason` |
| --- | --- | --- | --- |
| Model EOS | `stop` | `Eos(token_id)` | `null` |
| Explicit `stop_token_ids` match | `stop` | `Token(token_id)` | Actual token ID |
| Output limit reached | `length` | `None` | `null` |

The contract preserves the triggering token, its logprob, and completion-token accounting. `ignore_eos=true` disables only model EOS termination; explicit request stop tokens remain active.

## Review Fixes

- Classify and truncate every speculative candidate at its first terminal token.
- Select the hedge winner using the retained prefix length.
- Use the retained prefix for KV-page copy-back, hidden-state compaction, DFlash context, and counters.
- Keep an idempotent truncation check immediately before the final speculative commit.
- Fix the accepted-draft-token boundary so a terminal token inside the accepted draft prefix is counted correctly.
- Restore early rejection for non-zero `min_tokens`, since the current scheduler does not yet implement EOS/stop-token masking.
- Remove fake speculative scaffolding and duplicate helper tests.
- Keep legacy bridge behavior for un-migrated model lines.

Only Qwen3 consumes the typed stop policy in this change. The legacy frontend conversion path remains compatible with other models.

## Verification

| Test | Result |
| --- | --- |
| `cargo fmt --all -- --check` | Passed |
| `git diff --check` | Passed |
| `cargo test --release -p pegainfer-frontend --lib -- --nocapture --test-threads=1` | 74 passed, 0 failed |
| `cargo check --release -p pegainfer-qwen3 --lib` | Passed |
| `cargo check --release -p pegainfer-qwen3 --tests` | Passed |
| Strict DSpark hedge gate with Qwen3-4B and `dspark_qwen3_4b_block7` | Passed |
| Qwen3 HTTP request with `ignore_eos=true`, `max_tokens=8` | `finish_reason=length`, `completion_tokens=8` |
| Qwen3 HTTP request with explicit stop token coverage | `finish_reason=stop`, numeric `stop_reason`, trigger token preserved |

The change does not modify model mathematics, attention kernels, sampling kernels, or CUDA Graph shapes.

RicardoMin:
Hi @FeathBow, your requested changes have been implemented in the latest commits.

Please take another look when you get a chance. Thank you for the detailed review!

xiaguan (changes requested):
Thank you for the careful revision. The production ordering is much closer to the required contract, and the `min_tokens`, stop-set lookup, and documentation changes address several parts of the previous review. I still do not think the new hedge test protects the worker-side bug, and there is one invariant leak in the new public type.

1. **Please make the hedge regression fail against the previous implementation.**

   `dflash_hedged_midspan_stop_retains_trigger` only asserts the final emitted tokens and `Terminal`. Before the latest worker-side fix, `execute_speculative_verify_impl` already called `truncate_after_terminal` after `run_step` and before `RequestKv::apply_speculative`, so those assertions still pass even if the worker ranks A/B candidates, copies KV/hidden state, and records DFlash context from the untruncated result.

   The test also derives its stop ID from a baseline run on the same hedged engine. Because the stop policy does not affect sampling, the old worker reproduces the same candidates and winner; the later executor truncation then produces exactly the expected external prefix. The parent gate requires each child to execute a hedge span, but `total_wins > 0` is aggregated across all three children. The stop child is not required to have a B win, and it never demonstrates a case where terminal truncation changes the A/B ordering.

   Please add a real hedge case that observes the request-local raw and retained candidate lengths and the selected winner, with a fixture where truncation changes the winner decision. It should also verify that the retained winner is what feeds the KV/hidden context and committed/controller accounting. As a practical mutation check, this gate must fail when the new pre-selection truncation in `try_execute_hedged_verify` is reverted while the executor-side safety truncation remains.

2. **Please keep the sorted stop-set invariant inside `StopPolicy`.**

   `StopPolicy::classify` relies on `binary_search`, while both `eos` and `token_ids` are public. Any caller can therefore construct `StopPolicy { token_ids: Arc::from([7, 3]), ... }` and get a silently incorrect classification, bypassing the normalization promised by the type. The production callers only need `new`, `default`, and `classify`; please make the fields private and add a narrow read-only accessor only if a real caller needs one.

3. **Please reduce the low-value tests and refresh the comments to match the final flow.**

   `normalizes_stop_sets_across_common_sizes` repeats the one-ID behavior already covered above it and allocates an entire 151,936-ID set while primarily retesting `sort_unstable`, `dedup`, and `binary_search`. It does not prove that per-step clones share the allocation or that the production wire conversion preserves the contract. A focused dedup/membership test plus the meaningful wire/hedge gates is enough.

   Several comments now overstate or contradict the implementation: the hedge test claims it proves pre-winner selection and commit ordering although it only checks the final stream; the parent gate still says it runs “two” losslessness tests and “one child per lossless suite” after adding a third, non-losslessness child; and `execute_speculative_verify_impl` says the worker returns the mathematically accepted span even though the worker now applies the stop policy and the executor only rechecks the invariant. Please remove or update these while narrowing the tests.

The main implementation direction looks reasonable, but the current regression can stay green with the original worker-side bug restored, so I cannot approve this head yet. Thank you.

RicardoMin:
## Summary

This update addresses the remaining review feedback for the Qwen3 stop-contract migration.

## Review Fixes

- Apply the stop policy to every speculative hedge candidate before winner selection.
- Ensure a terminal suffix cannot affect the selected hedge winner, hidden-state compaction, DFlash context, KV commit, or accounting.
- Strengthen the production hedge gate to require a real `raw_winner=B -> retained_winner=A -> selected=A` transition.
- Verify that the retained length matches the context append and KV commit length.
- Keep `StopPolicy` fields private and preserve its normalized sorted-set invariant.
- Remove redundant helper coverage and align comments with the actual execution flow.
- Preserve the legacy contract for models that have not migrated to the typed stop policy.

## Verification

| Check | Result |
| --- | --- |
| Strict DSpark hedge gate (`hedged_ladder_passes_the_lossless_gates`) | Passed |
| Mid-span explicit-stop regression (`dflash_hedged_midspan_stop_retains_trigger`) | Passed |
| Mutation check (reverting pre-selection truncation in `try_execute_hedged_verify`) | Failed as expected (`retained_winner=B`, gate caught bug) |
| `cargo fmt --all -- --check` | Passed |
| `git diff --check` | Passed |

The strict hedge test now fails if terminal truncation is moved after winner selection, even when the final output is later corrected by the legacy safety truncation.
This PR remains scoped to Qwen3. Other model lines and their legacy scheduler contracts are unchanged.

RicardoMin:
Hi! @xiaguan ,The PR description has been updated with the full verification details. Please take another look when you have a moment. Thank you!

pegainfer-qwen3/tests/dflash_speculative_gate.rs
```diff
@@ -787,23 +895,99 @@ fn hedged_ladder_passes_the_lossless_gates() {
...
     }
-    assert!(
-        total_wins > 0,
-        "no hedge chain ever won across {total_rounds} hedged rounds"
-    );
-    assert!(
-        total_spans > total_wins,
-        "every hedge span won ({total_wins}/{total_spans}) — the discard path never executed"
-    );
+    assert!(total_rounds > 0 && total_spans > 0);
```
FeathBow, line 1003: The previous gate asserted total_wins > 0 (the page copy-back / hidden compaction branch executed at least once) and total_spans > total_wins (the discard branch executed too). This head parses the win count and drops it, replacing both with total_spans > 0. The stop child cannot supply that coverage: its qualifying trace is raw_winner=B with selected=A, which is precisely the case where copy-back does not run. So after this change no test shows the hedge copy-back branch ever executes. I recommend restoring both assertions over the two losslessness children (per child or aggregated as before) and leaving the stop child to its own worker-trace check.

pegainfer-qwen3/src/executor.rs
```diff
@@ -3632,9 +3636,93 @@ impl LocalQwen3Lane {
...
+        // winner; otherwise a discarded suffix can win the hedge and advance
+        // the wrong KV/hidden state and acceptance statistics.
+        let (results_a, results_b) = all_results.split_at_mut(requests.len());
+        anyhow::ensure!(
+            results_b.len() == hedge_spans.len(),
+            "hedge returned {} B results for {} hedge spans",
+            results_b.len(),
+            hedge_spans.len()
+        );
+        let trace = std::env::var_os("PEGAINFER_TEST_LOG").is_some();
```
FeathBow, line 3651: I recommend keeping raw A/B lengths (captured before truncation) plus selected_is_b, dropping the retained_* / *_best_* recomputation, and dropping the hedged field on VerifyResult and the hedged parameter on record_verify_dflash_context: if the commit log line is emitted for every verify round (not only hedged ones) the parent can pair detail and commit lines per request by order without the flag, and the context line is not needed.

FeathBow (commented):
Thanks. Two things I would still fix before merge. First, the parent ladder now parses the hedge win count and discards it: total_wins > 0 and total_spans > total_wins are gone, and the stop child's qualifying case is raw_winner=B with selected=A, which is exactly the case where copy-back does not run. So no test proves the copy-back branch executes any more; please restore both assertions for the two losslessness children. Second, the trace that makes the new gate mutation-sensitive is larger than the two facts it needs (raw B win flipped to A with retained < raw, and selected_len == commit_len): retained_winner equals selected by construction and context_len equals selected_len by construction, and the hedged flag on VerifyResult and record_verify_dflash_context exists only to pair log lines. btw please remove outdated docs and descriptions.

FeathBow:
@codex review

RicardoMin:
Sorry, I do not have my computer on hand at the moment. I will resolve the CL problem tomorrow.

pegainfer-frontend/src/vllm/wire.rs
```diff
@@ -76,14 +78,36 @@ pub(crate) fn convert_sampling(params: &EngineCoreSamplingParams) -> SamplingPar
...
 
-/// Reject request parameters the engine would otherwise silently ignore.
+pub(crate) fn convert_stop_policy(params: &EngineCoreSamplingParams) -> StopPolicy {
+    StopPolicy::new(
+        // Qwen3 owns the complete model EOS set in generation_config. The
+        // protocol's optional primary ID only tells us whether EOS is active;
+        // using it as a singleton would miss secondary model EOS IDs.
+        params
+            .eos_token_id
+            .map_or(EosPolicy::Ignore, |_| EosPolicy::ModelDefault),
```
FeathBow, line 88: This conversion discards the protocol's primary EOS ID and applies EOS priority to the model's entire EOS set. The pinned vLLM lowering already merges secondary EOS IDs into stop_token_ids; its stop classifier omits stop_reason only for the primary EOS and reports the actual ID for the other stop-token matches.

For example, with primary EOS 99 and secondary EOS 77, generating 77 with EOS enabled should report stop_reason=77. Here classify() returns StopCause::Eos(77), so the bridge reports null instead. The same loss occurs when the caller explicitly includes 77 in the stop set.

Please retain enough information to distinguish the primary EOS from the other stop IDs when reporting the cause. The adjacent comment about missing secondary EOS with a primary-only policy also needs adjustment: those secondary IDs already arrive in the wire stop set.

pegainfer-qwen3/tests/dflash_speculative_gate.rs
```diff
@@ -787,23 +897,110 @@ fn hedged_ladder_passes_the_lossless_gates() {
...
+                let selected_len = value("selected_len=").and_then(|v| v.parse().ok());
+                if !request.is_some_and(|id| stop_requests.contains(id)) {
+                    continue;
+                }
+                let context_len = request
+                    .and_then(|id| context_by_request.get_mut(id))
+                    .and_then(VecDeque::pop_front);
+                let commit_len = request
+                    .and_then(|id| commit_by_request.get_mut(id))
+                    .and_then(VecDeque::pop_front);
```
FeathBow, line 977: Each hedge-detail line pops one entry from the request's commit queue, but commit lines are emitted for every verify round while detail lines are emitted only for rounds that actually produce a hedge. If a request has an ordinary verify round before its first hedged round, the first detail line is compared with that earlier commit.

After the missing-field issue is resolved, this can still cause a false failure or a coincidental pass when the unrelated lengths match. Please make the pairing account for rounds without a hedge so the assertion compares the worker result and commit from the same round.

FeathBow (changes requested):
Thank you for addressing the earlier worker-ordering, stop-set ownership, and win/discard coverage feedback. The production path now applies terminal truncation before hedge selection and context recording.

I found four remaining issues. The strict hedge gate still requires an appended= field that is no longer emitted, so it cannot pass when executed. The wire policy also treats secondary model EOS IDs as primary EOS, losing the stop reason required by the pinned vLLM contract. In addition, the HTTP probe accepts several invalid outcomes, and the detail/commit pairing can compare different verify rounds.

Please address these issues and refresh the validation evidence for the resulting head. Handle redundant compatibility and test/comment descriptions.

RicardoMin:
@codex review

RicardoMin:
## Summary

This head resolves all four findings from FeathBow's Sep 8 review and refreshes
the validation evidence end-to-end on a fresh 32 GB instance (Qwen3-4B +
`deepseek-ai/dspark_qwen3_4b_block7`).

## Review Fixes

- **Gate fields now match production output.** The strict gate no longer
  requires the removed `appended=` field. It parses the production
  `Qwen3 DFlash hedge detail round=/request=/raw_a=/raw_b_lens=/selected=/selected_len=`
  line and pairs it with the `Qwen3 DFlash context context_len=` and
  `Qwen3 DFlash commit accepted_len=` lines, asserting
  `selected_len == context_len == commit_len`.
- **Same-round pairing.** Hedge-detail, context, and commit records are keyed
  by `(round, request)`, so an ordinary verify round before the first hedged
  round can no longer be compared against a different round's commit.
- **Primary vs secondary EOS.** `convert_stop_policy` keeps the protocol's
  primary `eos_token_id` as `EosPolicy::Token(primary)` instead of
  `ModelDefault`. Secondary model EOS IDs — which the pinned vLLM lowering
  already merges into `stop_token_ids` — now classify as
  `StopCause::Token(id)` and report the actual `stop_reason`; only the primary
  EOS yields `StopCause::Eos` with a null wire `stop_reason`. Covered by
  `convert_stop_policy_keeps_eos_and_explicit_stops_independent` and
  `primary_eos_keeps_secondary_model_eos_as_token_stop`.
- **Copy-back / discard coverage restored.** The parent ladder re-asserts
  `total_wins > 0` and `total_spans > total_wins` over the two losslessness
  children, and the stop child still requires a worker-side
  `raw_winner=B -> selected=A -> selected_len < raw_b_max` case.
- **Stale descriptions cleaned.** The nonexistent `ActiveRequest`/`StepEmitter`
  architecture text was dropped, and production comments now describe the
  actual pre-selection truncation flow.

## Verification

Environment: RTX 4080 SUPER 32 GB, driver 580.142 / CUDA 12.8, Rust
nightly-2026-07-10; target `Qwen/Qwen3-4B`, drafter
`deepseek-ai/dspark_qwen3_4b_block7` (block_size=7, markov_rank=256).

| Check | Result |
| --- | --- |
| Strict DSpark hedge gate `hedged_ladder_passes_the_lossless_gates` (`PEGAINFER_REQUIRE_HEDGE_GATE=1`, `PEGAINFER_SPEC_HEDGE=8`, positions `0,1,2`) | Passed (36.4 s; greedy + heterogeneous losslessness + mid-span stop children) |
| Mutation check: revert pre-selection truncation in `try_execute_hedged_verify` (executor-side safety truncation kept) | Failed as expected — `hedge/commit mismatch round=28: selected 2 vs commit 1` |
| `cargo test --release -p pegainfer-frontend --lib` | 80 passed, 0 failed |
| `cargo fmt --all -- --check` | Passed |
| `git diff --check` | Passed |

This PR remains scoped to Qwen3. Other model lines and their legacy scheduler
contracts are unchanged.

xiaguan:
Thanks for the updates! There are still a couple of gaps in the validation:

- In the hedge test, `continue` skips any result whose last token isn't the stop token, including a result that emitted more tokens after the stop. I appended an extra token to one stopped result locally, and the full gate still passed. Could you check the stop position, finish reason and count before filtering candidates, and validate the no-stop outcomes too?
- The attached HTTP probe can also report a false pass. I ran the unmodified script against a deliberately invalid local HTTP service: a string `stop_reason`, eight completion tokens with a full-vocabulary stop set, `[null]` logprobs, and streamed text after the finish event. It still passed all eight checks and exited 0, even with a model-name mismatch. Could you make these checks reject malformed responses and verify the actual trigger ID, its logprob, token count and stream termination? Please commit the corrected probe so the reported HTTP results are reproducible. This checks the probe's assertions; it doesn't mean the real server produced those bad responses.

There is also some duplication we can remove: merge the two `StepCollector` collection loops, consolidate the overlapping stop-policy cases in `stop.rs` and `wire.rs`, and carry the policy in `VerifyStepItem` instead of maintaining a parallel array. The goal here is fewer wrappers and duplicated responsibilities; there's no need to split files just to meet a line count.

RicardoMin:
## Summary

This update completes the final review round for the Qwen3 typed stop contract.
It closes the hedge-gate filter hole the reviewer reproduced, replaces the
attached HTTP probe with a strict, repository-committed version, and removes
the remaining duplication around request stop policies.

## Review fixes

1. **The hedge gate no longer filters candidates out.**
   `dflash_hedged_midspan_stop_retains_trigger` used to `continue` past any
   result whose last token was not the stop ID, so a run that emitted extra
   tokens after a mid-span stop, and every no-stop run, went unvalidated. It
   now validates every candidate: if the stop token appears it must be the
   final emitted token (position, `finish_reason`, and `completion_tokens` are
   all asserted); if it never appears, the run must terminate cleanly at the
   length limit. Appending a token after the stop now fails the gate
   (`stop N was followed by 1 more token(s)`).

2. **The HTTP probe is strict and reproducible.**
   The previous attached script accepted a string `stop_reason`, eight
   completion tokens under a full-vocabulary stop set, `[null]` logprobs,
   streamed text after the finish event, and a model-name mismatch. The new
   `scripts/qwen3_stop_contract_probe.py` rejects all of them and verifies the
   real trigger ID, its logprob, the token count, and stream termination. A
   built-in `--self-check` replays those five malformed shapes (plus a valid
   control) against a local mock server, so the reviewer's regression is
   reproducible with no GPU.

3. **Duplication removed.**
   `StopPolicy` is now carried in `VerifyStepItem` instead of a parallel array
   threaded through `VerifyPlan` and `StepCommand`; the two `StepCollector`
   collection loops are merged; and the overlapping stop-policy test cases are
   consolidated so `stop.rs` owns classification semantics while `wire.rs`
   only tests the wire-to-policy conversion.

## What the probe verifies

Against each OpenAI-compatible server it drives 7 `/v1/completions` shapes and
one `/v1/models` probe, then scores 9 checks:

| Check | Passes only when |
| --- | --- |
| `model_present` | `/v1/models` lists the requested model name |
| `baseline_control` | `ignore_eos`, no stop set → `length` with exactly `max_tokens` tokens and no `stop_reason` |
| `explicit_stop_ignore_eos` | `stop`, integer `stop_reason` from the requested set, one token under a full-vocabulary set |
| `explicit_stop_eos_enabled` | `stop`; `stop_reason` is a legal integer or absent (model EOS may win) |
| `stop_set_order_invariant` | ascending and reversed stop sets yield the same ID and token count |
| `trigger_logprob_preserved` | the last emitted token carries a non-null finite logprob |
| `stream_reports_typed_stop` | stream ends with `[DONE]`, typed `stop_reason`, no content after the finish event, valid trigger logprob |
| `mixed_controls_pass_3_of_3` | three concurrent ignore-EOS requests all end at the length limit |
| `mixed_explicit_stops_pass_3_of_3` | three concurrent explicit-stop requests all end with a typed stop |

`--self-check` spins up a local mock server and asserts that a valid service
passes while five malformed responses (string `stop_reason`, extra tokens under
a full-vocabulary set, `[null]` logprobs, post-finish stream content,
model-name mismatch) are each rejected. `--require-legacy-gap` additionally
asserts the expected adapted-passes/legacy-fails A/B outcome as an exit code.

## Running it

Self-check (no GPU, stdlib only):

```bash
python3 scripts/qwen3_stop_contract_probe.py --self-check
```

Start the adapted Qwen3 server:

```bash
cargo run --release -p pegainfer-server -- \
  --model-path "$QWEN3_MODEL" \
  --served-model-name qwen3-adapted \
  --port 18081
```

Start the legacy Qwen3.5 server (its Triton AOT kernels need a Python with
Triton at build time):

```bash
export PEGAINFER_TRITON_PYTHON=/path/to/python-with-triton
cargo run --release -p pegainfer-server --features qwen35 -- \
  --model-path "$QWEN35_MODEL" \
  --served-model-name qwen35-legacy \
  --port 18082
```

Then run the A/B probe:

```bash
python3 scripts/qwen3_stop_contract_probe.py \
  --qwen3-url http://127.0.0.1:18081 --qwen3-model qwen3-adapted \
  --qwen35-url http://127.0.0.1:18082 --qwen35-model qwen35-legacy \
  --require-legacy-gap --out stop-contract-ab.json
```

## Validation

- `cargo test --release -p pegainfer-frontend --lib`: 80 passed
- `cargo test --release -p pegainfer-qwen3 --lib`: 104 passed
- `cargo clippy --release --locked -p pegainfer-frontend -p pegainfer-qwen3 --all-targets -- -D warnings`: passed
- Strict DSpark hedge gate (`PEGAINFER_REQUIRE_HEDGE_GATE=1`, Qwen3-4B +
  `dspark_qwen3_4b_block7`): passed
- Mutation checks: reverting worker-side truncation, the executor-side safety
  truncation, or appending a token after a stop each make the gate fail
- Probe self-check: 6/6; live A/B (Qwen3-4B vs Qwen3.5-0.8B): adapted 9/9,
  legacy 2/9 with the typed-stop checks failing exactly as expected

xiaguan:
Thanks for the updates! The hedge test now rejects the injected token after a stop, and the policy-array and collection-loop cleanup looks good.

I rechecked `9b91ad31` and found three cases where the new HTTP probe still reports **9/9 PASS**:

- An extra token in the same SSE frame as the finish metadata is skipped by `request_stream`, so a full-vocabulary stop response with two tokens is counted as one.
- With a single explicit stop ID, a missing logprob on the final stop token inherits the previous token's finite logprob and passes the trigger check.
- A response containing token ID `12095` but reporting `stop_reason=17` passes when both IDs are in the stop set. Membership alone doesn't verify the actual trigger.

These were deliberately constructed local HTTP responses, not errors observed from the inference server. Could you request the returned token IDs and check the exact trigger, its corresponding logprob, and the complete token sequence? The parser should also account for or reject content in a terminal frame instead of silently skipping it.

One other point: removing **only the executor-side second truncation loop**, with worker truncation intact, still passed the real hedge gate locally. That differs from the reported mutation result. Could you clarify what that second pass needs to protect and update the validation note? Both worker paths already normalize the span before recording context; if a guard is needed at commit, an explicit invariant check would make a broken worker result visible instead of silently truncating it again.

RicardoMin:
Related to #865

# Qwen3: preserve explicit stop-token causes in the stepped contract

## Why this is a separate PR

This is the scoped follow-up requested by the maintainers during review of
[#865](https://github.com/pegainfer-project/pegainfer/pull/865).  They asked
that the broad stop-contract change be split so that the shared boundary,
frontend bridges, and one model can be reviewed and validated independently.
This PR therefore extracts the Qwen3 migration from that work; Qwen3.5 and the
other model schedulers remain on their existing contract for now.

## Summary

The previous vLLM dependency update fixed several frontend compatibility issues,
but the stepped Qwen3 path still collapsed two independent controls into the
single legacy `ignore_eos` flag.  That made an explicit `stop_token_ids` request
indistinguishable from a model-EOS request and forced the bridge to guess a
synthetic stop token after the scheduler had already discarded the real one.

This PR gives Qwen3 a typed stop contract.  It preserves the sampled trigger
token and its logprob, carries the concrete `StopCause` through the scheduler
and stepped bridge, and keeps EOS handling independent from request-provided
stop IDs.

## What was wrong

The old contract exposed only `FinishReason::Stop` or `FinishReason::Length`.
When a request stopped, the bridge could not tell whether the model emitted EOS
or an explicit request stop token.  It therefore reconstructed a sentinel (EOS
first, otherwise the first configured stop ID).  That reconstruction can report
the wrong token, loses the token's logprob, and is incorrect for a speculative
span where the first terminal token is followed by additional accepted tokens.

The old boolean also could not express the valid combination "ignore model EOS,
but still stop on these explicit request token IDs".

## Contract change

| Event | Legacy behavior | Qwen3 stepped behavior |
| --- | --- | --- |
| Model EOS | `FinishReason::Stop`; trigger may be suppressed or reconstructed | `FinishReason::Stop` + `StopCause::Eos(id)`; token is retained internally; wire `stop_reason` is absent |
| Explicit `stop_token_ids` match | Often treated as ordinary output or replaced by a guessed sentinel | `FinishReason::Stop` + `StopCause::Token(id)`; actual ID is reported as wire `stop_reason` |
| `ignore_eos=true` | Could inadvertently disable explicit stops | Disables only model EOS; explicit request stops remain active |
| Length limit | `FinishReason::Length` | `FinishReason::Length` + no stop cause; final sampled token is retained |
| Speculative span | Trigger/suffix ownership was ambiguous | Commit the prefix through the first trigger and discard only the suffix after it |

EOS has precedence when the same ID is both the active EOS token and an explicit
request stop.  Completion-token accounting is incremented once, including the
trigger token.

## Scope and compatibility

Following the maintainer's scope request on #865, this PR intentionally
migrates **Qwen3 only**.  The shared request/step types
accept an optional typed cause, while the existing legacy event path remains
available for models that have not been audited.  The legacy bridge keeps its
synthetic-sentinel fallback only when an old producer supplies no typed cause.
Therefore Qwen3.5 and other model schedulers are not changed in this PR and do
not need to adopt the new resolver contract yet.  If the maintainers agree with
the semantics, the remaining model lines can be migrated one at a time with
their own lifecycle tests.

## Implementation

- Added `StopPolicy`, `EosPolicy`, and `StopCause` at the frontend engine
  boundary.
- Converted wire EOS and explicit stop fields without collapsing them into one
  boolean for the stepped path.
- Propagated the policy through Qwen3 request state, ledger updates, prefill,
  ordinary decode, and speculative verification.
- Emitted the trigger token before terminal metadata and retained its logprob.
- Mapped `StopCause::Token(id)` to the vLLM-compatible wire `stop_reason`.
- Left the legacy bridge fallback and un-migrated model implementations intact.
- Updated two existing test fixtures (K3 and simulator) only to supply the new
  shared default field; no legacy model runtime behavior is changed.

## Automated verification (previous validation)

These results are from earlier validation. The current rerun below covers HTTP A/B.

| Check | Result |
| --- | --- |
| `cargo test --release -p pegainfer-frontend --lib` | 80 passed, 0 failed |
| `cargo test --release -p pegainfer-qwen3 --lib` | 104 passed, 0 failed |
| Qwen3 request-stop focused tests | 4 passed, 0 failed |
| Qwen3 speculative-stop focused tests | 2 passed, 0 failed |
| `pegainfer-sim` integration target `frontend_e2e` | 18 passed, 0 failed |
| `cargo check --release -p pegainfer-qwen35 --features qwen35` | Passed (control build only) |
| `cargo build --release --locked -p pegainfer-server --features qwen35 --bin pegainfer` | Passed |
| `cargo fmt --all -- --check` | Passed |
| `git diff --check` | Passed |

## HTTP A/B verification

Re-ran the unmodified in-repository `scripts/qwen3_stop_contract_probe.py` on
2026-09-18 at 10:17 UTC against the source for PR head
`3e1b392ba8d17868411c293f3ec0832034e3d3f5`, using real Qwen3-4B and
Qwen3.5-0.8B weights on the rented SM89 GPU. **Exit status: 0.**

The previous 8/8 and 2/8 totals omitted `model_present`. The committed probe
checks nine conditions: **Qwen3 passes 9/9; the healthy Qwen3.5 legacy control
passes 3/9**, with six expected `tokens_after_stop` failures.
`legacy_gap_observed=True`.

The full-vocabulary stop set makes the first sampled token an explicit stop.
Results below are `finish_reason / stop_reason / completion_tokens`:

| Target | Control (`ignore_eos=true`) | Explicit stop + EOS ignored | Explicit stop + EOS enabled |
| --- | --- | --- | --- |
| Qwen3-4B (adapted) | `length / null / 8` | `stop / 12095 / 1` | `stop / 12095 / 1` |
| Qwen3.5-0.8B (legacy control) | `length / null / 8` | `length / null / 8` | `length / null / 8` |

Qwen3's JSON and SSE logprob cases both returned token IDs `[12095]`, the exact
`stop_reason=12095`, one completion token, and sampled logprob `-0.35472107`.
The streaming case also passed complete-sequence and termination checks.

Qwen3.5 is intentionally unchanged by this Qwen3-scoped PR. The probe requires
healthy endpoints and the semantic legacy stop gap; an unavailable endpoint or
wrong model cannot satisfy `--require-legacy-gap`. This result covers the
probe's cases, not every legacy model or workload.

<details>
<summary>Source and binary provenance</summary>

All 1,029 source-manifest entries, including the four submodule revisions,
match this PR head; submodules are clean. The previously validated release
binary was reused after checking its SHA-256, rather than rebuilt for this
HTTP-only rerun.

- Probe SHA-256: `3dd7403426bd491d85e4259a3ba6f02b214c3d131504e35be6324d47a3bf74a3`
- Server binary SHA-256: `47b63dc8bd59107cfa3733b1050fc5733fab736f061d5efbde3824c5cad028d1`

</details>

## Reproduction

The validated release binary was built with:

```bash
cargo build --release --locked -p pegainfer-server --features qwen35 --bin pegainfer
```

Qwen3.5 requires its feature-gated Triton build environment. Start these in
separate terminals, waiting for Qwen3 to become ready before starting Qwen3.5.
This order matters on the shared GPU because startup computes KV capacity from
available memory. The model paths below are the ones used on the test instance.

```bash
./target/release/pegainfer \
  --model-path /root/autodl-tmp/models/Qwen3-4B \
  --served-model-name qwen3-adapted --port 18081 \
  --gpu-memory-utilization 0.35
```

```bash
./target/release/pegainfer \
  --model-path /root/autodl-tmp/models/Qwen3.5-0.8B \
  --served-model-name qwen35-legacy --port 18082 --max-batch 8
```

Exact probe command used for this rerun (Python standard library only):

```bash
python3 -B scripts/qwen3_stop_contract_probe.py \
  --qwen3-eos-token-id 151645 \
  --qwen35-eos-token-id 248046 \
  --require-legacy-gap \
  --out /root/autodl-tmp/pr978-head-3e1b392b-validation-20260918/http-ab.json
```

The output directory must exist. The EOS values are the primary IDs resolved
by the serving tokenizers. URLs, served names, vocabulary sizes and the
eight-token budget use the committed script's defaults.

The script's own output, copied verbatim:

```text

qwen3: http://127.0.0.1:18081 (qwen3-adapted)
  [PASS] model_present
  [PASS] baseline_control
  [PASS] explicit_stop_ignore_eos
  [PASS] explicit_stop_eos_enabled
  [PASS] stop_set_order_invariant
  [PASS] trigger_logprob_preserved
  [PASS] stream_reports_typed_stop
  [PASS] mixed_controls_pass_3_of_3
  [PASS] mixed_explicit_stops_pass_3_of_3
healthy=True, new_contract_passed=True

qwen35: http://127.0.0.1:18082 (qwen35-legacy)
  [PASS] model_present
  [PASS] baseline_control
  [FAIL] explicit_stop_ignore_eos: tokens_after_stop
  [FAIL] explicit_stop_eos_enabled: tokens_after_stop
  [FAIL] stop_set_order_invariant: tokens_after_stop
  [FAIL] trigger_logprob_preserved: tokens_after_stop
  [FAIL] stream_reports_typed_stop: tokens_after_stop
  [PASS] mixed_controls_pass_3_of_3
  [FAIL] mixed_explicit_stops_pass_3_of_3: tokens_after_stop
healthy=True, new_contract_passed=False

legacy_gap_observed=True
wrote /root/autodl-tmp/pr978-head-3e1b392b-validation-20260918/http-ab.json
```

The process exited **0**. The `[FAIL]` lines are the expected unchanged legacy
behavior; `--require-legacy-gap` requires that contrast while Qwen3 must pass
all nine checks and both endpoints must remain healthy.

## Follow-up

As requested during review of #865, this PR deliberately stops at the Qwen3
migration boundary.  After the
maintainers confirm that the independent EOS/request-stop semantics are wanted,
the same policy propagation and resolver audit can be applied to Qwen3.5 and the
other legacy model lines in separate, model-scoped changes.

RicardoMin:
Hi @xiaguan ,
I've addressed the previous feedback and updated the PR accordingly. Could you please take another look when you have a moment?
Thanks!

FeathBow:
@codex review please

FeathBow (approved):
Thanks! Only one thing blocking from my side, the HTTP table reports 8/8 while the committed probe gates on nine checks, so could you please re-run the in-repo script against this head and paste its own output.

RicardoMin:
Thanks for catching the stale table. It omitted `model_present`; the committed probe has nine checks.

I reran the unmodified in-repo `scripts/qwen3_stop_contract_probe.py` on 2026-09-18 against the source for head `3e1b392ba8d17868411c293f3ec0832034e3d3f5`, using real Qwen3-4B and Qwen3.5-0.8B weights. I verified all 1,029 source/submodule manifest entries against this head and reused the previously validated release binary after checking its SHA-256. No source changes were needed.

Command:

```bash
python3 -B scripts/qwen3_stop_contract_probe.py \
  --qwen3-eos-token-id 151645 \
  --qwen35-eos-token-id 248046 \
  --require-legacy-gap \
  --out /root/autodl-tmp/pr978-head-3e1b392b-validation-20260918/http-ab.json
```

The script's own output, verbatim:

```text

qwen3: http://127.0.0.1:18081 (qwen3-adapted)
  [PASS] model_present
  [PASS] baseline_control
  [PASS] explicit_stop_ignore_eos
  [PASS] explicit_stop_eos_enabled
  [PASS] stop_set_order_invariant
  [PASS] trigger_logprob_preserved
  [PASS] stream_reports_typed_stop
  [PASS] mixed_controls_pass_3_of_3
  [PASS] mixed_explicit_stops_pass_3_of_3
healthy=True, new_contract_passed=True

qwen35: http://127.0.0.1:18082 (qwen35-legacy)
  [PASS] model_present
  [PASS] baseline_control
  [FAIL] explicit_stop_ignore_eos: tokens_after_stop
  [FAIL] explicit_stop_eos_enabled: tokens_after_stop
  [FAIL] stop_set_order_invariant: tokens_after_stop
  [FAIL] trigger_logprob_preserved: tokens_after_stop
  [FAIL] stream_reports_typed_stop: tokens_after_stop
  [PASS] mixed_controls_pass_3_of_3
  [FAIL] mixed_explicit_stops_pass_3_of_3: tokens_after_stop
healthy=True, new_contract_passed=False

legacy_gap_observed=True
wrote /root/autodl-tmp/pr978-head-3e1b392b-validation-20260918/http-ab.json
```

Exit status: **0**. Qwen3 passes **9/9**; the healthy, unchanged Qwen3.5 control passes **3/9**, with the six expected `tokens_after_stop` failures. `--require-legacy-gap` explicitly requires this contrast. The correct HTTP totals are therefore **9/9 and 3/9**, including `model_present`.

RicardoMin:
Hi! @FeathBow, I've rerun the test and pasted the results below. Could you please take another look when you have a chance?

FeathBow:
Thanks for your contributions! No blockers in the workflow from my side.

FeathBow (approved):

PR #981, opened 2026-08-28
feat(qwen35): launch only on the step contract

xiaguan:
## Summary
- Qwen3.5 `ModelLine::launch` returns `LaunchedEngine::Stepped` only; `Qwen35Scheduler` implements submit/step/metrics. This crate no longer has a dual `EngineHandle` path.
- Tests rewritten onto `Engine` / `EngineHarness`; crate-internal scheduler tests talk the contract only.
- GPU: `PEGAINFER_TEST_MODEL_PATH=/data/models/Qwen3.5-4B` — `e2e_scheduler` + `sampling_behavior` + `chunked_prefill`: 4 passed, 1 ignored (TP2, one GPU). e2e load aligned to 8 slots (HEAD already scheduled at 8; load was still `MAX_BATCH=64` and OOM'd 16GB).
- Lib tests 90 passed / 7 ignored; clippy `-D warnings --all-targets` passed.
- CUDA/cuBLAS bind on first `step` of the driver thread; overlap wait stays inside `step`.
- Frontend legacy modules are not deleted (glm52/kimi/dsv2/gemma4 still `Handle`).

## Test plan
- [ ] `cargo test --release -p pegainfer-qwen35 --features qwen35 --lib`
- [ ] `cargo clippy --release -p pegainfer-qwen35 --features qwen35 --all-targets -- -D warnings`
- [ ] GPU gates with `PEGAINFER_TEST_MODEL_PATH=/data/models/Qwen3.5-4B`: `e2e_scheduler`, `sampling_behavior`, `chunked_prefill`
- [ ] `serving_tp2` when 2 GPUs are available (ignored on one GPU)

Made with [Cursor](https://cursor.com)


RicardoMin:
Hi @xiaguan, while planning the Qwen3.5 stop-policy work for #1062, I found that the current legacy interfaces would need some additional changes to carry the request’s stop policy and return the actual triggering token, its logprob, and the stop cause together.
I then noticed that this PR is migrating Qwen3.5 to the step contract, which already provides the structures needed for that work. Building on this migration would let us avoid adding temporary plumbing to the legacy interfaces, while keeping the stop-policy implementation in a separate follow-up PR.
I noticed this PR hasn’t had recent updates. Do you think you’ll have time to bring it up to date and complete it for merging? If you’re busy with other work at the moment, I’d be happy to help update and validate it. Please let me know what would be most useful.

PR #983, opened 2026-08-28
refactor(qwen3): move the report harness out of the library API

### LINKED ISSUE #944: Move the Qwen3 benchmark harness out of the library public API
## Problem

`pegainfer-qwen3/src/lib.rs` unconditionally exports `pub mod kernel_bench`, making a report-only harness part of the production library API.

At `69c4e149`, `src/kernel_bench.rs` is 1,505 lines with 55 functions and roughly 68 public items spanning decode attention, batched and single prefill attention, dense operators, L2 cache clearing, device bandwidth, and cuBLASLt tuning.

The consumers are the two `kernel-report` feature-gated binaries:

- `qwen3_kernel_report`
- `qwen3_model_report`

Production Qwen3 users do not need this API. A core CSR builder is also re-exported through `kernel_bench` even though the model report can import it from `pegainfer_core` directly.

## Direction

Make the report harness bin-owned and private. Remove the library-level public surface instead of preserving it through a compatibility wrapper.

A shared private report-support module can be included by both binaries. Split it by cohesive benchmark domain where that reduces the amount of code a maintainer must understand; avoid thin files that only forward names.

## Acceptance criteria

- `pegainfer-qwen3` no longer exposes `pub mod kernel_bench`.
- The normal library build does not compile report-only benchmark harness code.
- `qwen3_kernel_report` and `qwen3_model_report` still compile under `kernel-report`.
- The model report imports shared production primitives such as `build_split_kv_csr` from their owning crate rather than through a report re-export.
- Report-only types and helpers are private to the binary targets.
- The existing report CLI flags and output schemas are unchanged.
- Qwen3 all-target check and Clippy pass with `kernel-report` enabled.
- No performance claim is required; this issue changes ownership and public surface only.

Do not add a new public report-support crate unless a third independent consumer establishes that boundary.

## Contributor guardrails

- The PR description must show the public surface before and after, list the two real consumers, and explain where each moved domain is now owned.
- Do not leave a deprecated/public compatibility module, add a public report-support crate, or replace the module with re-export wrappers.
- Do not duplicate the 1,505-line harness into both binaries. Shared code stays in one private bin-owned module.
- Do not add source-grep, string-count, or module-existence tests. Compiling the two real binaries and the production library is the relevant integration evidence.
- Do not change report schemas, CLI behavior, benchmark cases, numeric policy, or kernel selection. Such changes need separate motivation and measurement.
- Include exact `cargo check` and Clippy commands/results for the production library and both `kernel-report` binaries. Unrelated cleanup is out of scope.

onenewcode:
## Description

Fixes #944.

`pegainfer-qwen3` exposed `pub mod kernel_bench` (1112 lines) as public API. Every normal library and server build compiled report-only benchmark code and shipped its types in the crate's public surface.

The harness moves to `pegainfer-qwen3/src/bin/report_support/`, split across five modules, and is `#[path]`-included by the two `kernel-report` binaries. It is now compiled per binary and never as part of the library.

- `qwen3_kernel_report` takes the whole tree. `qwen3_model_report` needs only `common::L2CacheClear` and includes that module alone, with the `#[allow(dead_code)]` on its own `mod` declaration rather than inside the shared module — so the other four modules stay dead-code-checked by the kernel report's build. Removing the allow and building `qwen3_kernel_report` alone yields 0 dead-code warnings, confirming the allow covers only the model report's narrower use.
- The model report imports production primitives from their owning crates: `build_split_kv_csr` from `pegainfer_core::ops`, `split_qkv_into` from `pegainfer_kernels::ops`, instead of reaching through a harness re-export.
- `SplitKvConfig::{new, label}` widen to `pub`. `parse` was already public and calls `new`, so a private constructor only pushed external callers onto the unchecked struct-literal path; `label` is `parse`'s inverse, and restating its format in the harness would silently break the round-trip contract that `label_parse_round_trips` guards.
- `SPLIT_KV_CHUNK_TOKENS` / `SPLIT_KV_TUNED_MAX_CHUNKS` are re-exported through `runtime` so the harness keeps deriving the Tuned split width from the runtime constants exactly as it did before the move, rather than restating `64x64` where a later retune would silently desync the report's `split_tmp_*` sizing.

Report schemas, CLI flags, benchmark cases, numeric policy, and kernel selection are unchanged; both binaries' `--help` output is identical.

**Incidental fix.** `qwen3_model_report` did not compile under `--features kernel-report` at `main`: it called `ops::split_qkv_into`, which `pegainfer-core` does not re-export. This was invisible because no CI job enables `kernel-report` — the two binaries carry `required-features = ["kernel-report"]`, and `qwen3-cuda-clippy` does not pass it (the feature pulls `pegainfer-cupti`, excluded at `ci.yml:300` for missing CUPTI headers in the CI toolkit). Worth a follow-up: this PR does not create that gap, but it does move ~1500 more lines behind it.

**Docs.** `docs/models/qwen3/roadmap.md` cleanup ledger strikes the `kernel_bench.rs` file-size entry. `docs/subsystems/kernels/kernel-op-reports.md` TL;DR records the new home and notes that paths in its Execution Log are historical — those are dated past-tense step records (Step 1 still uses the pre-flattening `crates/` layout), so rewriting them to today's paths would falsify the record rather than fix it.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue) — the `qwen3_model_report` compile break described above
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) — the crate's public API loses `pub mod kernel_bench`. No runtime behavior changes and no in-workspace consumer breaks; flagged because #944 mandates the removal with no compatibility shim, so any out-of-tree user of that path is affected.
- [x] Documentation update

## Checklist

Run locally on RTX 5090 (sm_120), driver 610.57.04:

| Gate | Result |
| --- | --- |
| `cargo check --release -p pegainfer-qwen3 --lib` | pass, 0 warnings |
| `cargo check --release -p pegainfer-qwen3 --all-targets --features kernel-report` | pass |
| `cargo clippy --release -p pegainfer-qwen3 --all-targets --features kernel-report -- -D warnings` | pass |
| `cargo test --release -p pegainfer-qwen3 --lib` | 88 passed, 0 failed |
| `cargo build --release -p pegainfer-qwen3 --features kernel-report --bin qwen3_kernel_report --bin qwen3_model_report` | both link |
| `cargo fmt --check` | pass |


FeathBow:
Please rebase onto main first, and address the CI issues :)

FeathBow:
LGTM. No blocking implementation defect found. The harness is bin-owned, its benchmark behaviour is preserved, and the normal library no longer includes `kernel_bench`.

FeathBow (approved):

Issue #1001, opened 2026-08-30
tracking: split #946 (qwen35 TP sharding + batched decode + CUDA graph + eval tooling) into acceptance-bounded PRs


Ma1oneZhang:
## Background

PR #946 (`feat(qwen35): shard GDR state per TP rank + batched eager/graph decode under TP`, +2920/−316, 27 files) stacks four independently verifiable changes on one branch. Per reviewer request ([comment](https://github.com/pegainfer-project/pegainfer/pull/946#issuecomment-5390279330)), it needs to land as acceptance-bounded PRs. This issue tracks the split. The two open codex review comments on #946 are absorbed by the relevant splits below.

## Splits

- [ ] **2b — TP correctness / memory**: sharded linear-attention / GDR state per TP rank (weight-loader stitch/shard, per-rank stitched qkv/conv1d + row/col shards, state/buffers at local sizes, all-reduce after local linear out_proj, TP-local group-6 decode reroute). Acceptance: 27B TP2 fits on 2×48 GB (Phase-1 replicated state OOMed); 27B/9B HF logits + e2e gates pass on the rebased head.
- [ ] **Batched eager TP decode**: one batched `run_decode_batch` over all decode rows (step-scoped pointer table, single batched rank-0 select, per-row fan-out) replacing per-request bs=1 ticks. Acceptance: throughput A/B on its own (#946 measured 16-concurrency aggregate 24.9 → 292.3 tok/s); gates unchanged.
- [ ] **TP CUDA Graph**: startup pre-capture sweep with warmup collectives per bucket message size, slot-indexed per-rank recurrent/conv state with scheduler-owned slot_idx + D2D compaction on retire, replay-only serving, 600 s watchdog. Gate: `enable_cuda_graph && local_decode_group_is_compiled(tp)`; 27B TP2 (group-6) stays batched eager. Acceptance: graph A/B (#946 measured 9B TP2 steady 706 → 767 tok/s, TPOT −8.8%); HF gates eager+graph PASS. Also absorbs the codex P2 on graph-slot reserve sizing (see below).
- [ ] **Eval tooling + benchmark docs**: `scripts/eval_mc.py`, `scripts/eval_rerun_truncated.py`, `docs/benchmarks/qwen35-27b-tp2-knowledge-eval.md`, tp-implementation records — after an exact-head runtime rerun of the benchmark numbers.

## Open codex review comments absorbed by the split

[...]

PR #946, opened 2026-08-20
feat(qwen35): shard GDR state per TP rank + batched eager/graph decode under TP (+27B benchmark validation)

Ma1oneZhang:
## Summary

Rebased onto current main (#870 landed its own Phase 1/2a). This PR carries the deltas #870 lacks, plus 27B benchmark validation:

- **Phase 2b — sharded linear-attention/GDR state per TP rank** (`2530dfc2`): rank-local slices end to end (weight_loader stitch/shard loaders, per-rank stitched qkv/conv1d + row/col shards, state/buffers at local sizes, all-reduce after local linear `out_proj`, TP-local group-6 decode reroute). Acceptance: 27B TP2 now fits on 2×48 GB cards (Phase-1 replicated state OOMed).
- **perf — batched eager TP decode** (`b3aa5e11`): #870's `execute_decode_rows` ran one bs=1 forward per request per tick; this adds `run_decode_batch` — one `batch_decode_eager_logits` over all decode rows with a step-scoped pointer table, one batched rank-0 `select_batch`, per-row fan-out. Measured: 16-concurrency aggregate **24.9 → 292.3 tok/s**.
- **P2c — CUDA Graph under TP** (`0570313f`): startup pre-capture sweep ported from qwen3 (warmup collectives per bucket message size, Capture/Launch per bucket, Finalize assert, 600 s watchdog), slot-indexed per-rank recurrent/conv state with scheduler-owned `slot_idx` + D2D compaction on retire, serve-time replay-only. Gate: `enable_cuda_graph && local_decode_group_is_compiled(tp)` — 4B/9B TP2 capture; 27B TP2 (group-6, TP-invariant) stays batched eager until group-6 batch-decode kernels exist. 9B TP2 serving bench: steady output **706 → 767 tok/s (+8.7%)**, TPOT −8.8%.
- **docs + benchmark record** (`b376772a`, `f56ce511`): rebase + P2c landing records in `tp-implementation.md`; 27B knowledge-benchmark parity snapshot in `docs/benchmarks/qwen35-27b-tp2-knowledge-eval.md`; runner scripts `scripts/eval_mc.py` + `scripts/eval_rerun_truncated.py`.

## Validation (2× RTX 4090, --features qwen35, on this branch)

- `cargo check`/`fmt` clean; lib suite **105/105**
[...]

PR #1003, opened 2026-08-30
feat(qwen35): shard linear-attention/GDR state per TP rank (split 1/4 of #946)

Ma1oneZhang:
## Description

Split 1/4 of #946, tracked in #1001. Followed by #1004 → #1005; merge order is this PR first.

Shards the Qwen3.5 linear-attention (GDR) recurrent state and per-request TP state per rank: `LocalGeometry` gains rank-local linear-attention head/dim accessors (with indivisibility validation on `linear_num_key_heads` / `linear_num_value_heads`), and `RecurrentState` / `GdrChunkwiseScratch35` size themselves from the rank-local geometry instead of global dims.

## Type of Change

## Evidence

- `cargo check -p pegainfer-qwen35 --features qwen35 --all-targets`: clean (sm_89, 2×RTX-4090 build host).
- `cargo test -p pegainfer-qwen35 --features qwen35 --lib`: 106/106 pass (7 GPU-gated tests skipped: no model weights on this host).
- `cargo fmt --all -- --check`: clean.

GPU-gated acceptance from the original #946 (27B TP2 HF/E2E + throughput A/B, measured 2026-08-20 on the pre-rebase branch) still needs a rerun on this rebased stack and is tracked in #1001; no numbers are re-claimed here.

## Checklist


Ma1oneZhang:
@codex please review

xiaguan:
Thanks for splitting this out. Two items are still needed before this PR is independently mergeable:

1. Please rerun the exact-head acceptance promised in #1001: 27B TP2 HF short/long gates, TP2 E2E including cleanup/re-admission, and the load/HBM evidence. The current checks cover compilation and CPU tests, while the PR notes that all GPU gates were skipped; the pre-rebase results do not validate `fcdeb5a4`.
2. Please update `docs/models/qwen35/tp-implementation.md` in this PR. It still says P2B sharding is the next follow-up and that linear-attention/GDR state remains replicated. Since #1003 must merge before the downstream documentation PR, the implementation record needs to remain accurate at this merge boundary.

Ma1oneZhang:
Exact-head acceptance rerun on `fcdeb5a4`, 2xRTX 4090 48GB (sm_89), model `Qwen/Qwen3.5-27B` @ `fc05daec18b0a78c049392ed2e771dde82bdf654` (fixture-pinned revision):

- `pega_logprobs_match_hf_golden_within_qwen35_tolerance_tp2` -- **PASS**. Sequential eager: mean 0.0210 / p99 0.0749 / max 0.1240; batched eager leg (drop -> re-prefill slot cycles): mean 0.0201 / max 0.0803. Within fixture tolerances (mean 0.06, p99/margin 0.20).
- `pega_logprobs_match_hf_long_golden_within_qwen35_tolerance_tp2` -- **PASS**. Prompts 4097/8192: mean 0.0177 / p99 0.0660 / max 0.0660.
- `test_e2e_qwen35_scheduler_tp2` -- **PASS** (81s): context rejection, logprobs parity, sequential/concurrent, consumer-drop survival.
- `qwen35_tp2_serves_openai_completions_over_http` -- **PASS** (28s).

Load/HBM: peak per-rank 35,988 / 36,822 MiB of 49,140 MiB -- 27B TP2 fits on 2x48GB with sharded GDR state; each test process fully freed HBM between runs.

Host notes: no system NCCL on this box; used pip wheel `nvidia-nccl-cu12` 2.31.2 with a `libnccl.so -> libnccl.so.2` symlink (cudarc search list lacks `libnccl.so.2`) via `LD_LIBRARY_PATH`.

The `tp-implementation.md` doc update is next; will push it to this branch.

xiaguan:
Thanks — the exact-code-head TP2 acceptance and the documentation update resolve the previous requests.

One blocker remains in `load_tensor_1d_f32_shard`: it casts `TensorView::data()` directly to `&[f32]` without validating the dtype or pointer alignment. A valid safetensors file can place an F32 tensor after an odd-sized tensor, making this pointer unaligned; constructing the slice is then UB, while another dtype with a byte length divisible by four is silently reinterpreted as F32. Please share a typed, alignment-safe F32 helper between the full and shard loaders: validate F32/1D, borrow aligned data, and decode unaligned bytes with `f32::from_le_bytes`.

pegainfer-qwen35/src/batch_decode.rs
```diff
@@ -268,13 +271,15 @@ impl Qwen35Model {
...
         let mut positions = Vec::with_capacity(bs);
+        let mut start_positions = Vec::with_capacity(bs);
         for (i, kv) in kv_states.iter_mut().enumerate() {
             let pos = kv.seq_len();
             self.ensure_rope_cache_covers(pos + 1)?;
             kv.ensure_capacity(pos + 1)?;
             kv.advance(1);
             recurrent_states[i].seq_len += 1;
             positions.push(pos as i32);
+            start_positions.push(pos);
```
FeathBow: `start_positions` is allocated and filled before the GQA predicate is checked. It is needed for the 27B group-6 paged-prefill fallback, but for models whose group has a compiled decode kernel it is immediately discarded. Those compiled-group paths therefore perform one avoidable host allocation and fill per rank per token.

The local predicate also appears equivalent to the existing config predicate: both full-attention head counts are divided by the same TP width, so `(q / tp) / (kv / tp) == q / kv` for every validated geometry. As a result, `LocalGeometry::local_decode_group_is_compiled()` returns the same result as `Config35::decode_group_is_compiled()` for valid configurations.

Would it be possible to create or derive `start_positions` only inside the uncompiled branch and share the one-token paged-plan construction with the existing hybrid fallback? This would keep the compiled route simpler and avoid maintaining two similar builders. I have not measured a throughput regression here.

pegainfer-qwen35/src/config/tp.rs
```diff
@@ -126,12 +131,37 @@ impl LocalGeometry {
...
+        // Phase 2b shards linear attention/GDR heads per rank; fail closed on
+        // indivisible head counts rather than falling back to replication.
+        if !config.linear_num_key_heads.is_multiple_of(tp.world_size()) {
+            return Err(ConfigError::TpIndivisible {
+                field: "linear_num_key_heads",
+                value: config.linear_num_key_heads,
+                world_size: tp.world_size(),
+            });
+        }
+        if !config
```
FeathBow: `LocalGeometry` receives a validated `Config35`, and `Config35::try_from` already guarantees `linear_num_value_heads % linear_num_key_heads == 0`. Once the preceding guard proves `linear_num_key_heads % world_size == 0`, value-head divisibility follows algebraically; this branch cannot fail for a production config.

The new test reaches this branch by mutating `linear_num_value_heads` from 32 to 31 after validation, which constructs a state outside the `Config35` invariant rather than exercising an input boundary. Could we keep the key-head TP guard, exercise it with an internally consistent case such as key/value = 17/34, and omit this derived branch, its test, and the corresponding documentation claim? That would leave one source of truth for the relationship between these head counts.

pegainfer-qwen35/src/weights.rs
```diff
@@ -912,4 +1039,153 @@ mod tests {
...
+        let len: usize = shape.iter().product();
+        let data: Vec<u8> = (0..len).flat_map(|i| (i as f32).to_le_bytes()).collect();
+        let view =
+            safetensors::tensor::TensorView::new(safetensors::Dtype::F32, shape.to_vec(), &data)
+                .unwrap();
+        safetensors::serialize([(name.to_string(), view)], None).unwrap()
+    }
+
+    #[test]
+    fn linear_attention_tp2_slices_match_synthetic_checkpoint() {
```
FeathBow: This test never calls `load_tensor_2d_row_stitch`, `load_tensor_1d_stitch`, either model loader wrapper, H2D, or a GDR kernel. It serializes/deserializes an F32 safetensors blob even though the production stitch loaders consume BF16, then manually slices CPU vectors with the same `segments` tuples it is asserting. The QKV `expected` value is built from those same tuples, while the conv and value-head assertions duplicate the two exact segment tests immediately above.

Could we remove `safetensors_fixture_f32` and this test, and rely on the direct QKV/conv segment tests for the non-obvious local layout plus the real HF gate for the production loader/kernel path? The adjacent TP1 segment test also appears unnecessary because its own comment notes that the TP1 production branch does not call this helper. This would avoid implying production checkpoint coverage where the test currently verifies dependency parsing and repeated slice arithmetic.

pegainfer-qwen35/src/config/tp.rs
```diff
@@ -83,6 +83,11 @@ pub(crate) struct LocalGeometry {
     local_full_attn_q_dim: usize,
     local_full_attn_kv_dim: usize,
     local_full_attn_gated_q_dim: usize,
+    local_linear_num_key_heads: usize,
+    local_linear_num_value_heads: usize,
+    local_linear_q_dim: usize,
```
FeathBow: No production code reads this field; it is stored only so the new arithmetic tests can inspect it. The constructor already needs a local `q_dim` to derive `local_linear_qkv_dim`, so do that calculation locally and omit the field. This avoids carrying test-only state in the runtime geometry contract.

pegainfer-qwen35/src/recurrent_state.rs
```diff
@@ -1,8 +1,11 @@
 //! Recurrent state for Qwen3.5 linear attention layers.
 //!
 //! Each linear attention layer maintains:
-//! - Recurrent state: [num_value_heads, key_head_dim, value_head_dim] f32, V contiguous ([H,K,V])
-//! - Conv state: [qkv_dim × (conv_kernel_dim - 1)] bf16
+//! - Recurrent state: [local_value_heads, key_head_dim, value_head_dim] f32, V contiguous ([H,K,V])
+//! - Conv state: [local_qkv_dim × (conv_kernel_dim - 1)] bf16
+//!
+//! Under TP both are rank-local: Phase 2b shards value heads (and fused qkv
```
FeathBow: The local shape and "never all-reduced" rule are useful context, but the same Phase-2b explanation now appears at the module, both fields, size helper, allocator callers, prefill, decode, loader sites, and worker capacity calculation. Could we keep one module/type-level ownership invariant plus the non-obvious fused-QKV stitching and post-out-proj reduction comments, while trimming phase labels and comments that restate the following expression or method name? That would preserve the durable design context with less repetition in production code.

FeathBow (commented):
Thanks! The core P2B design is coherent, some smaill changes for the acceptance boundary. The remaining notes are non-blocking cleanup suggestions intended to keep the patch focused.

FeathBow (commented):
seems all resolved, please rebase origin main :)

pegainfer-qwen35/src/batch_decode.rs
```diff
@@ -450,31 +473,14 @@ impl Qwen35Model {
...
-        .with_context(|| {
-            format!(
-                "hybrid decode build PrefillPagedPlan bs={bs}, pages={}, heads={}/{}, head_dim={}",
-                page_indices.iter().map(Vec::len).sum::<usize>(),
-                self.config.num_attention_heads,
-                self.config.num_key_value_heads,
-                self.config.head_dim
-            )
-        })?;
+            "hybrid decode",
```
FeathBow, line 482: `one_token_paged_plan` now has two callers: the eager route passes `geometry.local_num_attention_heads()` / `local_num_key_value_heads()`, while the hybrid route passes `self.config.num_attention_heads` / `num_key_value_heads`. The hybrid route is only reachable from the graph path, and TP refuses graphs, so the values are equal today. I recommend using the geometry accessors in both call sites so the helper has one contract and the equality does not depend on a reader remembering the graph/TP exclusion.

pegainfer-qwen35/src/config/tp.rs
```diff
@@ -308,11 +356,51 @@ mod tests {
...
+            ConfigError::TpIndivisible {
+                field: "linear_num_key_heads",
+                value: 17,
+                world_size: 2,
+            }
+        );
+    }
+
+    #[test]
+    fn computes_tp2_linear_attention_local_dimensions() {
```
FeathBow: None of them can fail without the formula changing, so they detect edits rather than protect a boundary. I recommend removing them and keeping `requires_linear_attention_key_head_divisibility` (a real input boundary) and `linear_qkv_shard_segments_stitch_head_local_slices` (the non-obvious per-segment tiling)

FeathBow (commented):
nothing below blocks merge. remaining non-blocking cleanup.

PR #1004, opened 2026-08-30
perf(qwen35): batch eager decode rows under TP (split 2/4 of #946)

Ma1oneZhang:
## Description

Split 2/4 of #946, tracked in #1001. **Stacked on #1003** (contains its commit); followed by #1005. Merge order: PR1 → this PR → PR3.

Batches the eager TP decode path: one batched decode step across ready rows per TP rank instead of per-request steps, including the paged full-attention fallback for TP-local GQA groups without a compiled batch-decode kernel (27B group-6 case).

## Type of Change

## Evidence

- `cargo check -p pegainfer-qwen35 --features qwen35 --all-targets`: clean.
- `cargo test -p pegainfer-qwen35 --features qwen35 --lib`: 106/106 pass (7 GPU-gated skipped).
- `cargo fmt --all -- --check`: clean.

Throughput A/B numbers cited in #946 come from its 2026-08-20 pre-rebase measurement; rerun on this stack is tracked in #1001.

## Checklist


Ma1oneZhang:
@codex please review

FeathBow:
@codex review please

pegainfer-qwen35/src/tp_executor.rs
```diff
@@ -1341,6 +1341,133 @@ impl TpWorkerState {
...
+    /// one batched sampling pass over the per-row sampling params.
+    ///
+    /// Seeded rows keep their former per-row semantics: they pass step 0 and
+    /// `select_batch` isolates each seeded row into its own single-row philox
+    /// call keyed on (request seed, step), so seeded output stays independent
+    /// of batch composition. Unseeded rows decorrelate inside the batched
+    /// call through the per-step command seed, same as the single-GPU batched
+    /// path. Returns one result row per request in command order on rank 0,
+    /// empty elsewhere.
+    fn run_decode_batch(
```
FeathBow, line 1365: Every TP2 number so far (the #1003 acceptance) went through the bs=1 per-request loop this PR removes. The batched leg of pega_logprobs_match_hf_golden_within_qwen35_tolerance_tp2 issues one N-row execute_decode per step, so on this head it is exactly the multi-row batched forward gate: GEMM and attention shapes change with bs, and HF tolerance under real batches is what proves the step. I recommend posting, at the rebased head: the TP2 short and long HF gates (both legs), test_e2e_qwen35_scheduler_tp2, and the three tp2_* worker tests (unified step, drop-all capacity, re-admission). The 24.9 -> 292.3 tok/s reference is the #946 pre-rebase figure; since the title is perf, I recommend either a same-context bs=1-loop vs batched A/B on this head or dropping the figure from the commit and leaving the A/B to #1001 as the body already says.

pegainfer-qwen35/src/tp_executor.rs
```diff
@@ -1341,6 +1341,133 @@ impl TpWorkerState {
...
+        let mut recurrent_refs: Vec<&mut RecurrentState> = Vec::with_capacity(bs);
+        for (kv, recurrent) in kv_slots.into_iter().zip(recurrent_slots) {
+            kv_refs.push(kv.expect("decode row state resolved above"));
+            recurrent_refs.push(recurrent.expect("decode row state resolved above"));
+        }
+
+        // Step-scoped GDR pointer tables over the full decode batch. They are
+        // rebuilt every step, so swap_remove retirement between steps can
+        // never leave a stale table behind.
+        let linear_pointer_tables = LinearStatePointerTables::from_recurrent_refs(
```
FeathBow: Maybe Perf improve, the removed per-request table was built once per request, and the graph path builds its table once over fixed slots. The pointer values do have to be re-uploaded each step because swap_remove moves states, but the allocation does not. I recommend holding one LinearStatePointerTables of capacity max_batch in TpWorkerState and adding a refill(ctx, refs, bs) that memcpy_htod's into the existing slices; validate_for already accepts capacity >= bs.

pegainfer-qwen35/src/tp_executor.rs
```diff
@@ -1341,6 +1341,133 @@ impl TpWorkerState {
...
+    ) -> Result<Vec<DecodeRequestResult>> {
+        let bs = requests.len();
+        if bs == 0 {
+            return Ok(Vec::new());
+        }
+
+        // Resolve the worker state slot of every row in command order.
+        // Decode request ids are unique within one command
+        // (validate_decode_requests), so each slot is borrowed at most once.
+        let mut row_of_state: Vec<Option<usize>> = vec![None; self.requests.len()];
```
FeathBow, line 1378: The resolution goes id -> state index -> row_of_state -> two Option vectors -> unwrap into refs. Since ids are unique per command, one iter_mut().enumerate() pass over self.requests that pushes (row, &mut state.kv, &mut state.recurrent) for hit slots, followed by sort_unstable_by_key(row) and a split into the two ref vectors, gives the same disjoint borrows with no Option layers, no expect, and no debug_assert. The phase check and the missing-state error can stay in the first loop that maps ids to rows.

PR #1006, opened 2026-08-30
docs(qwen35): TP2 implementation record + knowledge-eval tooling (split 4/4 of #946)

Ma1oneZhang:
## Description

Split 4/4 of #946, tracked in #1001. Independent of splits 1–3 (docs + bench tooling only; no runtime dependency).

- Documents the Qwen3.5 TP implementation record and the 27B TP2 knowledge-eval setup.
- Fixes `scripts/eval_mc.py` MMLU-Redux extraction (absorbs the codex review comment from #946): answers are extracted via the shared `sg_extract_labels(text, 'ABCD')` marker logic instead of a first-capital-letter regex, matching the other benchmarks.
- Flags the 27B TP2 MMLU-Redux 94.09 figure in the doc as produced by the pre-fix extractor: do not cite until rerun with the fixed extractor.

## Type of Change

- [x] Documentation update

## Evidence

- Extraction fix validated on 8 hand-written answer-shape samples (marker-first, bare letter, parenthesised, mixed case).
- Docs-only otherwise; no runtime code touched.
- Benchmark figures in the doc are the 2026-08-20 pre-rebase measurements from #946 and are marked rerun-pending where the extractor fix changes semantics (#1001).

## Checklist


Ma1oneZhang:
@codex please review

FeathBow:
@codex review please

scripts/eval_rerun_truncated.py
```diff
@@ -0,0 +1,123 @@
...
+"""
+import argparse
+import asyncio
+import json
+import sys
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).parent))
+from eval_mc import (first_capital, mmlupro_extract, MMLUREDUX_RE,
+                     sg_extract_labels, sg_extract_content, STOP_MAP)
```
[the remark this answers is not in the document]
Ma1oneZhang: Fixed in 3a1c528c: the rerun script now imports the shared marker-aware extractor and routes MMLU-Redux through `sg_extract_labels(text, 'ABCD')`, matching eval_mc.py; the dead `MMLUREDUX_RE` import is gone. `py_compile` + import-surface check pass.

scripts/eval_rerun_truncated.py
```diff
@@ -0,0 +1,123 @@
...
+    if bench == 'ceval':
+        return first_capital(text)
+    if bench == 'mmlu_pro':
+        return mmlupro_extract(text).lower()
+    if bench == 'mmlu_redux':
+        m = MMLUREDUX_RE.search(text)
+        return m.group(1) if m else ''
+    pred = sg_extract_labels(text)
+    if pred is None and item.get('options'):
+        content = sg_extract_content(text, item['options'])
```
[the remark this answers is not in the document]
Ma1oneZhang: Fixed in 3a1c528c on the producer side: eval_mc.py now persists `options` in every SuperGPQA sample record, so this content fallback fires on reruns. Sample files written before that commit still lack the field and keep letter-only extraction.

docs/index.md
```diff
@@ -234,6 +234,7 @@ Organized by domain (model line / subsystem / playbook / lesson) instead of by l
...
 | `benchmarks/mixed-load-itl.md` | Qwen3-4B + Qwen3.5 mixed-load ITL (#244, #375): chunking-off sweeps via `bench_serving mixed`. Both freeze active decode for the full prefill. Qwen3 p99 blows up with prompt/QPS; the old Qwen3.5 “p99-immune” table is a **measurement artifact** (primary: hardcoded `max_batch=4` slot starvation — see #470 / `models/qwen35/mixed-load-itl-470.md`; secondary: short `bg_output_len`). Prefix reuse defeats it on Qwen3. |
 | `benchmarks/accuracy-eval-results.md` | Phase 1 GSM8K: Qwen3-4B PASS (pegainfer 85.37% vs HF 85.82%, delta -0.45 pp). Qwen3.5-4B historical FAIL recovered by #250 (strict 79.38%, flexible 79.30% vs HF 79.45%). |
+| `benchmarks/qwen35-27b-tp2-knowledge-eval.md` | Qwen3.5-27B TP2 (2× RTX 4090, batched eager decode) knowledge benchmarks vs official: MMLU-Redux 94.09 vs 93.2 (full 5330), C-Eval 88.11 vs 90.5 (full 1346, thinking-cap truncation rerun-merged) — both in cross-harness band; MMLU-Pro/SuperGPQA partial smokes only. |
```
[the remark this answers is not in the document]
Ma1oneZhang: Fixed in 270ebc81: the index row now states the 94.09 was scored with the pre-fix first-capital extractor and is rerun-pending / not citable; C-Eval keeps its in-band framing.

docs/models/qwen35/tp-design.md
```diff
@@ -1,6 +1,6 @@
 # Qwen3.5 Tensor Parallelism Design
 
-> **TL;DR:** Qwen3.5 TP Phase 2 is two separately delivered correctness milestones: P2a adds eager `RunUnifiedStep` with a shared ordered `RequestId` plan while retaining Phase 1 replicated GDR; P2b shards the head-indexed linear-attention/GDR surface and adds only the hidden all-reduce after local `out_proj`.
+> **TL;DR:** Qwen3.5 tensor parallelism should reuse Qwen3's controller/worker TP runtime and stay degree-parametric. Phases 1, 2a, and 2b are implemented (see `tp-implementation.md` for the landing record, including the rebase onto #870): eager dense TP, TP mixed-step unified execution, and sharded linear-attention/GDR state. Remaining design work: TP CUDA Graph capture.
```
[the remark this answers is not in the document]
Ma1oneZhang: Fixed at the tip (tp-design.md TL;DR): Phases 1/2a/2b/2c implemented, P2c gated on the compiled decode GQA group (4B/9B TP2 capture; 27B group-6 stays eager); remaining work named as group-6 batch-decode kernels and TP perf gates.

Ma1oneZhang:
@codex review

FeathBow:
@codex review please

scripts/eval_mc.py
```diff
@@ -0,0 +1,397 @@
...
+            if m:
+                hit = m.group(1)
+                if hit in esc:
+                    return options_content[esc.index(hit)]
+                return hit
+    return None
+
+
+# ---------------------------------------------------------------- runner
+STOP_MAP = {'mmlu_pro': ['Question:']}
```
FeathBow: Could we remove `Question:` from the generation-time stop strings for this thinking-model evaluation? The current vLLM Rust dependency matches stop strings against decoded text before separating reasoning from final content. If the model repeats `Question:` during reasoning, generation ends before it produces a final answer, and the sample is scored as incorrect.

The rerun script uses the same stop string, so increasing the token budget does not address this termination condition. The existing `scripts/eval_gsm8k_thinking.py` already documents this issue and avoids using `Question:` as a generation-time stop.

scripts/eval_mc.py
```diff
@@ -0,0 +1,397 @@
...
+                msg = data['choices'][0]['message']
+                usage = data.get('usage') or {}
+                return {
+                    'content': msg.get('content') or '',
+                    'reasoning': msg.get('reasoning') or '',
+                    'prompt_tokens': usage.get('prompt_tokens') or 0,
+                    'completion_tokens': usage.get('completion_tokens') or 0,
+                    'finish_reason': data['choices'][0].get('finish_reason') or '',
+                }
+            body = f'HTTP {r.status_code} {r.text[:200]!r}'
```
FeathBow, line 257: Could we make the evaluation exit unsuccessfully when requests remain unresolved after retries? `run_completion` currently converts an exhausted request into empty content and a reasoning field marked `__ERROR__`. `evaluate` counts that sample as an incorrect answer, while `main()` returns `None`, producing exit code 0.

An unavailable service, an incorrect model name, or invalid request parameters can therefore produce a normal-looking low-accuracy report. Recording `api_errors` is useful, but callers relying on the exit status can still treat an incomplete run as successful.

Please retain the failure diagnostics and clearly mark the evaluation as incomplete, so service failures do not become evidence about model accuracy.

scripts/eval_rerun_truncated.py
```diff
@@ -0,0 +1,124 @@
...
+                pred = None
+    return pred or ''
+
+
+async def run(base_url, model, rows, bench, max_tokens, concurrency, timeout):
+    sem = asyncio.Semaphore(concurrency)
+    async with httpx.AsyncClient(timeout=httpx.Timeout(timeout)) as client:
+        async def one(row):
+            async with sem:
+                payload = {'model': model, 'max_tokens': max_tokens,
```
FeathBow: Could the rerun script inherit and validate the configuration from the original summary? The initial evaluator accepts custom `--model` and `--temperature` values, but the rerun hardcodes `temperature=0.0` and uses an independent model default without checking the original run.

After an initial evaluation with a different model or nonzero temperature, following the documented rerun command can merge outputs from different configurations. The summary still retains the original `model` and `temperature`, so the merged score is attributed to a configuration that did not produce all of its samples.

Please carry the original configuration into the rerun and ensure that any configuration changes are reflected accurately in the resulting report.

scripts/eval_mc.py
```diff
@@ -0,0 +1,397 @@
...
+                t0 = time.time()
+                out = await run_completion(client, args.base_url, args.model,
+                                           prompt, args.max_tokens, args.temperature, name)
+                return i, out, round(time.time() - t0, 2)
+
+        t0 = time.time()
+        results = {}
+        tasks = [one(i, p) for i, p in enumerate(prompts)]
+        done = 0
+        for fut in asyncio.as_completed(tasks):
```
FeathBow: Could we persist each completed sample as it arrives? Results currently remain in the in-memory `results` dictionary until every request finishes; only then are the sample file and summary written.

scripts/eval_rerun_truncated.py
```diff
@@ -0,0 +1,135 @@
...
+                            row['completion_tokens'] = usage.get('completion_tokens') or 0
+                            row['finish_reason'] = data['choices'][0].get('finish_reason') or ''
+                            return row
+                        err = f'HTTP {r.status_code} {r.text[:200]!r}'
+                    except Exception as e:  # noqa: BLE001
+                        err = f'{type(e).__name__}: {e!s}'
+                    await asyncio.sleep(min(2 ** attempt, 30))
+                raise RuntimeError(f"rerun failed: {err}")
+
+        return await asyncio.gather(*[one(r) for r in rows])
```
FeathBow: Could we persist each completed rerun before waiting for the whole batch? If 47 of 48 requests succeed and the last one exhausts its retries, gather raises before main reaches either output write, so all 47 successful responses are lost. An interrupted run has the same problem.

These reruns use a larger thinking budget, making that lost work particularly expensive. The initial runner already writes and flushes each completed sample; reusing that approach here would preserve progress while still reporting the failed request explicitly.

pegainfer-qwen35/src/lib.rs
```diff
@@ -304,6 +304,31 @@ mod tests {
         assert_eq!(Qwen35SchedulerPolicy::default(), Qwen35SchedulerPolicy::Off);
     }
 
+    #[test]
```
FeathBow: Could we drop these three missing-path smoke tests and their supporting comments/imports? They exercise the same startup chain and only assert that the error does not contain the old “eager execution only” wording. A reworded TP rejection would pass, and a CUDA initialization error can also satisfy the assertion because device initialization happens before reading the model config.

Consequently, these tests do not establish that startup reached model loading or that the graph/eager gate behaves correctly. The existing TP2 serving test already starts through this entry point with enable_cuda_graph=true and exercises real requests, so these additional tests do not appear to provide distinct behavioral coverage.

FeathBow:
@codex review

FeathBow:
@codex review Again please？

FeathBow:
Thanks @codex please check again

Ma1oneZhang:
@codex  please check again

FeathBow:
@codex check again please?

FeathBow:
@codex check again now.

scripts/eval_mc.py
```diff
@@ -0,0 +1,486 @@
...
+        print(f'[{name}] checkpoint {partial_path.name} does not match this '
+              'run configuration; starting fresh', file=sys.stderr, flush=True)
+        partial_path.unlink()
+    if partial_path.exists():
+        for line in partial_path.read_text(encoding='utf-8').splitlines():
+            try:
+                prev = json.loads(line)
+            except json.JSONDecodeError:
+                continue  # torn tail write of the killed run
+            i = prev.get('idx')
```
FeathBow, line 348: Could we rescore recovered responses using the current extractor and golds before counting them? The resume check compares generation settings and prompts, but then trusts the saved gold, pred, and correct fields. If an extractor is fixed or a dataset corrects an answer without changing the prompt, rerunning the same command can still report the old accuracy.

The generated text can be reused; routing it through the shared scoring step with the current sample would avoid regenerating expensive answers while keeping the score current.

scripts/eval_rerun_truncated.py
```diff
@@ -0,0 +1,217 @@
...
+    return pred or ''
+
+
+async def run(base_url, model, temperature, rows, bench, max_tokens, concurrency, timeout,
+              partial_path):
+    sem = asyncio.Semaphore(concurrency)
+    # Recover rows from a killed previous attempt before touching the
+    # checkpoint: the partial JSONL is appended to, never truncated.
+    out_rows = {}
+    if partial_path.exists():
```
FeathBow, line 49: Could we repair the incomplete trailing record before opening this checkpoint for append? Skipping a JSONDecodeError leaves the original bytes in place. If a killed write left a partial JSON object without a newline, the next completed record is appended directly to that fragment, so the next resume discards that successful response too.

Both runners have this sequence. Restoring the record boundary before appending would preserve the new completions; malformed complete lines should also be distinguished from an interrupted tail rather than silently skipped.

scripts/eval_mc.py
```diff
@@ -0,0 +1,486 @@
...
+    wall_min = (time.time() - t0) / 60.0
+    tot_completion = sum(r.get('completion_tokens', 0) for r in records)
+    (out_path / f'{name}_samples.json').write_text(json.dumps(records, ensure_ascii=False, indent=1))
+    summary = {'benchmark': name, 'model': args.model, 'n': len(records),
+               'acc': round(acc * 100, 2), 'api_errors': fails,
+               'truncated_thinking': trunc,
+               'max_tokens': args.max_tokens, 'temperature': args.temperature,
+               'concurrency': args.concurrency,
+               'sample': args.sample or None, 'seed': args.seed,
+               'wall_min': round(wall_min, 2),
```
FeathBow, line 414: Could agg_toks_per_s use only tokens generated during the interval measured by t0? records already contains recovered responses, so tot_completion includes historical tokens, while wall_min only covers this invocation. Resuming 999 completed samples and generating one more therefore divides all 1,000 samples' tokens by the time for one; a fully cached run reports throughput without making any requests.

Keeping the cumulative token count is useful, but throughput needs matching token and time intervals, or should be omitted when that timing is unavailable.

scripts/eval_rerun_truncated.py
```diff
@@ -0,0 +1,217 @@
...
+                    if STOP_MAP.get(bench):
+                        payload['stop'] = STOP_MAP[bench]
+                    for attempt in range(6):
+                        try:
+                            r = await client.post(f'{base_url}/chat/completions', json=payload)
+                            if r.status_code == 200:
+                                data = r.json()
+                                msg = data['choices'][0]['message']
+                                usage = data.get('usage') or {}
+                                row = dict(row)
```
FeathBow, line 82: Could we update pred and correct before persisting a successful rerun? This copies the original empty-answer record and replaces its output, but retains the old score fields. A sample whose gold is B can now answer B while its partial JSONL record still says pred="" and correct=false.

The final merged file is rescored later, but an interrupted run leaves this inconsistent partial artifact. Scoring before persistence, or explicitly storing raw responses without stale score fields, would keep the checkpoint unambiguous.

PR #1013, opened 2026-08-30
perf(gemma4): an admission coalesce door protects the decode tail

### LINKED ISSUE #1011: gemma4: a live decode batch pays one admission event per arrival
## Description

A live decode stream's tail gap prices the number of admission events, not their size: every arrival that rides the batch costs the stream one mixed-step stall. Under closed-loop load the arrivals land one at a time right after completions, so a c8 batch takes five separate stalls where a batching engine takes two, and its ITL p99 sits at 65–70 ms against the other engine's 14. Nothing in the intake lets arrivals wait for each other.

## Proposed Plan

- An opt-in door in the engine loop, `PEGAINFER_ADMIT_COALESCE_MS=N` (1 ≤ N ≤ 2000; unset, `0` or `off` admits on sight): arrivals that would ride a live batch wait out the window together and share one mixed admission; an idle engine never waits.
- Two releases inside the window: a full mixed cohort (the queue would fill every free slot up to the gathered-prompt ceiling) opens the door early, and a shallow batch — live rows plus arrivals under half the slots — skips it, since with few streams the tail sits under the percentile on its own and waiting only taxes first-token time.
- Default off. The closed-loop c8 tail is bistable under the door (the completion waves either synchronize into cohorts or they do not) and the open-loop measurement that decides the serving default shows the closed-loop tail to be a bench artefact; the door ships as a tail-latency profile for closed-loop-shaped traffic, with that data in the PR.

## Acceptance Criteria

- Unset, the pinned-composition prompts answer with the same bytes as `main`, and the engine gates pass; with the door set, the lifecycle and roster gates pass and the door's refusals (out-of-range values) are exercised.
- Serving A/B, door on against off on one binary at c4 / c8 / c16, two cooled rounds each, reported as mean [range] with the bistable c8 rounds shown as they are, plus the first-token cost the door charges.
- Clippy at deny-warnings and the library tests stay green.

FeathBow:
## Description

Closes #1011

`PEGAINFER_ADMIT_COALESCE_MS=N` holds arrivals that would ride a live decode batch for up to N ms, so one window's arrivals share a single mixed admission instead of invading the batch once each — a live stream's tail gap prices the number of admission events, not their size. A full mixed cohort releases early, a shallow batch (live rows plus arrivals under half the slots) skips the door, and an idle engine admits on sight. Unset, `0` or `off` serves byte-identically to before; the door is off by default.

It stays off by default on the numbers below: under closed-loop load the c8 tail under the door is bistable — the completion waves either synchronize into cohorts and the tail drops to the other engine's band, or they do not and the door only charges first-token time — c4 gains nothing, and c16 gains. Open-loop Poisson arrivals, the measurement that decides the serving default, show the closed-loop tail to be a bench artefact of synchronized fixed-length cohorts; the door is a profile for traffic shaped like that.

## Test Env

Single GPU (`sm_89`, 48 GiB, x86_64)

## Verification

- `cargo fmt --all -- --check` and `git diff --check` against the base pass; the Gemma 4 release server builds; all-target Gemma Clippy passes with `-D warnings`; release lib tests pass (43, the door's value parser among them).
- Unset: the 26B checkpoint answers four greedy public `/v1/completions` requests (31-, 35-, 33- and 1421-token prompts, 64 forced tokens) with bytes identical to `main`'s binary, md5 `219de1e3c09c`; the idle-refill, full-roster, roster-edge, mixed-step and gathered-walk gates pass.
- `PEGAINFER_ADMIT_COALESCE_MS=5000` and `=abc` refuse to start with the parser's message; with the door at 300 ms the shared-lane and gathered lifecycles, the roster edge and the full-roster gate pass.
- Serving A/B on one binary, door off against `PEGAINFER_ADMIT_COALESCE_MS=300`, `vllm bench serve` random 1024 in / 128 out at 24 / 48 / 64 prompts, two cooled alternating rounds, mean [range]:

  | | c4 | c8 | c16 |
  | --- | --- | --- | --- |
  | output tok/s, off | 288.2 [286.4–290.0] | 438.3 [435.2–441.3] | 600.5 [598.3–602.7] |
  | output tok/s, door | 287.8 [286.9–288.7] | 413.1 [412.1–414.0] | 597.9 [597.2–598.6] |
  | median TPOT ms, off | 12.05 [11.97–12.12] | 15.38 [15.15–15.60] | 23.55 [23.39–23.71] |
  | median TPOT ms, door | 12.08 [12.06–12.10] | 15.36 [15.28–15.44] | 21.54 [21.50–21.57] |
  | p99 ITL ms, off | 89.0 [88.6–89.4] | 94.2 [94.2–94.3] | 97.8 [97.6–98.1] |
  | p99 ITL ms, door | 89.5 [88.6–90.4] | 94.6 [93.8–95.5] | 98.7 [98.6–98.8] |
  | median TTFT ms, off | 249.4 [248.2–250.6] | 382.4 [382.3–382.4] | 393.1 [391.1–395.1] |
  | median TTFT ms, door | 249.0 [248.9–249.1] | 401.5 [401.1–401.9] | 680.8 [679.8–681.7] |

  c4 is flat — the shallow-batch skip keeps the door out. Both c8 rounds landed on the coin's tax side: the completion waves did not synchronize into cohorts, so the door charged 5.7% throughput and 19 ms of first-token time for no tail change; the campaign round that caught the cohort attractor (ITL p99 falling from the high band to the other engine's) does so with roughly even odds, which is exactly why the door is not the default. c16 trades first-token time (+288 ms median) for the decode tail's pace (median TPOT −8.5%) at flat throughput.


FeathBow:
the Issue and PR now say a window's arrivals land as one capacity-bounded back-to-back intake burst, name the 512-row gather budget explicitly, and state that with 1024-token prompts every admission is its own mixed step. No architectural change is attempted. The title becomes `perf(gemma4): an admission coalesce door batches admission interruptions`, and the description claims exactly what the table shows — c16 trades median TPOT for first-token time at flat throughput, c4 is neutral, c8 pays, p99 ITL is flat to slightly worse everywhere — so the door stays off by default. I also ran the workload where the gather can actually merge, 256 in and 128 out at c16 over two cooled alternating rounds on one binary, and the door loses there too, so no tail-protection claim survives anywhere in the body. 

The timeout language now follows your three cases: a full cohort releases before N, the timeout release lands no earlier than N at the first intake turn after the window elapses, and there is no hard bound on how much later, since the engine drains its submission channel before each intake. The old door-plus-lane lifecycle line is removed as well — that combination now refuses at load, before a device opens — and the verification describes the gates as they run today, with the serving-knob guard clearing the door everywhere and the dedicated intake gate as the one place it is enabled. That gate passes at the exact head through the maintainer runner: a closed wait schedules nothing while decode advances the incumbent, the cohort release drains one burst with one Scheduled event per waiter, the timeout release lands at the injected deadline, and a full roster admits nothing until a slot frees.

PR #1015, opened 2026-08-31
refactor(gemma4): one forward path carries the numeric gates

### LINKED ISSUE #1014: gemma4: the probe forward and the serving path both carry the numeric gates
## Description

The numeric gates were designed against a standalone probe forward; the serving slices moved them onto the serving path itself, and the probe forward and its layer oracle are gone. Two leftovers of the old arrangement remain. Nothing checks the one path against itself — the probe comparison used to answer for incremental-versus-whole-prompt consistency, and nothing replaced it. And `GemmaServe` still carries a test-only switch that freezes the sliding window's front release for one gate, `eviction_is_footprint_only`, which compares release-on against release-off — an axis that only exists because of the switch.

## Proposed Plan

- An incremental-versus-recompute gate on the serving path: every prompt position's logits through the decode arena match a whole-prompt recompute of the same path, and decode steps fed the recompute's own greedy picks match it too, so one divergence cannot cascade.
- The release switch and the eviction gate retire together: `window_left` masks out-of-window keys in-kernel, the waypoint gates cross the window against HF with the resident-page arithmetic asserted, and both arms of the new gate release their fronts as production does.

## Acceptance Criteria

- The new gate passes through the runner, whose membership check owns the swap; the waypoint and greedy HF gates pass unchanged with the same fixtures and tolerances.
- The release build is byte-identical to `main` (the retired switch was test-only): the pinned-composition prompts answer with the same bytes.
- Clippy at deny-warnings and the library tests stay green.

FeathBow:
## Description

Closes #1014

The serving slices already retired the probe forward: the HF fixture gates drive the serving path teacher forced, with the same fixtures, tolerances and distributional floors. What remained of the old arrangement goes here. An incremental-versus-recompute gate answers for the one path's self-consistency where the probe comparison used to: every prompt position's logits through the decode arena match a whole-prompt recompute, and four decode steps fed the recompute's own picks match it too. The release-freezing test switch retires with the eviction gate it served — `window_left` masks out-of-window keys in-kernel, the waypoint gates cross the window against HF with the resident-page arithmetic asserted, and both arms of the new gate release their fronts as production does.

## Test Env

Single GPU (`sm_89`, 48 GiB, x86_64)

## Verification

- `cargo fmt --all -- --check` and `git diff --check` against `main` pass; the Gemma 4 release server builds; all-target Gemma Clippy passes with `-D warnings`; release lib tests pass.
- `incremental_serving_matches_recompute` passes 1/1 through the runner (whose membership check owns the manifest swap), on the golden fixture's short prompt the 2.0 ceiling was calibrated on: per-position max |dlogit| runs 0 to 0.79 against the 2.0 ceiling, with one near-tie top-1 flip admitted under the rule that both rows must rank the two candidates within the ceiling of each other; the four decode steps stay under it too. The 12B waypoint and greedy HF gates pass unchanged with the same fixtures and tolerances.
- The 26B checkpoint answers four greedy public `/v1/completions` requests (31-, 35-, 33- and 1421-token prompts, 64 forced tokens) with bytes identical to `main`'s binary, md5 `219de1e3c09c` (everything removed was `#[cfg(test)]`).

### DIFF
--- a/pegainfer-gemma4/src/kv.rs
+++ b/pegainfer-gemma4/src/kv.rs
@@ -132,9 +132,8 @@ impl SlidingLocalKv {
     }
 
     /// Move the frontier without releasing anything: the overlapped
-    /// prefill defers its release to join time, and the eviction A/B pins
-    /// the footprint. Everything else goes through
-    /// [`Self::advance_and_release`].
+    /// prefill defers its release to join time. Everything else goes
+    /// through [`Self::advance_and_release`].
     pub(crate) fn advance(&mut self, count: usize) {
         self.frontier += count;
     }
--- a/pegainfer-gemma4/src/serve.rs
+++ b/pegainfer-gemma4/src/serve.rs
@@ -667,8 +667,6 @@ pub(crate) struct GemmaServe {
     sliding_window: usize,
     global_split_factor: usize,
     final_logit_softcapping: f32,
-    #[cfg(test)]
-    release_enabled: bool,
     sliding_cos: DeviceVec,
     sliding_sin: DeviceVec,
     global_cos: DeviceVec,
@@ -797,8 +795,6 @@ impl GemmaServe {
             sliding_window,
             global_split_factor,
             final_logit_softcapping,
-            #[cfg(test)]
-            release_enabled: true,
             sliding_cos,
             sliding_sin,
             global_cos,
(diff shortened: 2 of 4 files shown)


FeathBow:
The strict gate has its exact-head runner output. At the rebased head the reworked gate already ran with the strict assert and passed under the default bf16 profile: thirteen compared rows, zero argmax flips, per-position max |dlogit| 0 to 1.5 against the 2.0 ceiling. Under `PEGAINFER_GATE_STORAGE=fp8` the same gate is red by design — the e4m3 pool's first divergent decision lands at position 2 and the strict assert stops it — which is the color that profile exists to show, beside its red agreement and waypoint gates. The earlier "one admitted near-tie flip" line came from a calibration on the pre-rebase base and does not reproduce on this one, which is exactly what let the contract tighten to strict with no separate margin constant.

The body and Issue text move with it: the short trajectory itself never crosses the window and says so, the crossing stays pinned by the waypoint, mixed-window, overlap and ragged gates, the no-coverage claim contracts to what the gate uniquely adds (full-vocabulary drift at every position of the nine-token prompt plus four non-cascading continuation steps), and the near-tie description is gone.

PR #1034, opened 2026-09-07
feat(qwen35): keep the auto prefill budget running under stream overlap

CAICAIIs:
## Summary

Make the `auto` scheduler policy's finishing-window deferral overlap-aware: with `--decode-overlap stream`, the window no longer defers the whole prefill budget, because the overlapped chunk already runs off the decode step — the deferral only delayed prefill.

## Background

#1033 allowed `auto` + `stream`. The follow-up measurement showed the combination's one open-loop cost — QPS16 TTFT `1828 vs 1484 ms` (stream alone) — comes almost entirely from this window: with many short decoders, some active request is within the 4-token finish window on most steps, and each such step runs decode-only while the prefill waits.

## Change

- `choose_prefill_budget` takes the overlap mode and skips the finishing-window full deferral when stream overlap is enabled; the capped FIFO-front budget (`min(remaining, --max-prefill-tokens)`) applies instead.
- Serial (`off` overlap) semantics are unchanged; new unit test `adaptive_prefill_budget_keeps_prefill_running_under_stream_overlap` pins the overlap branch.
- Docs updated (`adaptive-scheduler-policy.md`, `unified-prefill-overlap.md`).

## Evidence (1x A100-40GB, upstream/main `fb16fc15` + this change, vLLM 0.27.0 baseline)

| QPS16 1024/128 | TPOT ms | ITL p99 ms | TTFT ms | out tok/s |
| #1033 combo (window kept) | 20.81 | 41.71 | 1828 | 873 |
| **combo, overlap-aware (this PR)** | 23.77 | **41.65** | **1264** | **992** |
| vLLM 0.27.0 | 23.60 | 93.43 | 218 | 1404 |

c8/c16 unchanged within run noise (ITL p99 `34.5 / 36.4 ms`). The redundant QPS16 TPOT win (already parity with vLLM) is traded for 31% TTFT and 14% throughput at an identical tail. The remaining TTFT/throughput gap to vLLM is total-throughput-bound (kernel-side batch decode slope), not schedulable.

Correctness: release lib tests 98 passed (including the new overlap-branch unit test), `hf_golden_gate` TP1 2/2 and TP2 (`--ignored`) 2/2, `e2e_scheduler` passed.

## Claim boundary

Both knobs remain opt-in (`off` defaults); `stream` keeps the `--max-batch <= 32` cap and single-GPU-only restriction. Single-run numbers on one GPU — not a parity or production-readiness claim.

### DIFF
--- a/pegainfer-qwen35/src/scheduler/plan.rs
+++ b/pegainfer-qwen35/src/scheduler/plan.rs
@@ -67,6 +67,7 @@ pub(super) fn choose_prefill_budget(
     base_budget: usize,
     active: &[ActiveDecodeState],
     prefilling: &[PrefillQueueState],
+    decode_overlap: bool,
 ) -> usize {
     assert!(
         base_budget > 0,
@@ -76,9 +77,13 @@ pub(super) fn choose_prefill_budget(
         return base_budget;
     }
 
-    if active
-        .iter()
-        .any(|req| req.remaining_tokens() <= DECODE_FINISH_WINDOW_TOKENS)
+    // With stream overlap the finishing request's last tokens keep ticking on
+    // the decode stream while the chunk runs aside, so the full prefill
+    // deferral would only delay prefill without buying decode latency.
+    if !decode_overlap
+        && active
+            .iter()
+            .any(|req| req.remaining_tokens() <= DECODE_FINISH_WINDOW_TOKENS)
     {
         return 0;
     }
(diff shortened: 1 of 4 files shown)


pegainfer-qwen35/src/scheduler/plan.rs
```diff
@@ -386,10 +415,45 @@ mod tests {
...
+                Qwen35SchedulerPolicy::Auto,
+                1024,
+                &active,
+                &prefilling,
+                true
+            ),
+            1024,
+            "with stream overlap the finishing window keeps prefill on the prefill stream instead of deferring it"
+        );
         assert!(
```
FeathBow, line 457: no blocking but the first assertion correctly exercises the changed branch: a decoder has four tokens left, overlap is enabled, and the budget remains 1024 instead of returning zero. The final assertion supplies a separate empty queue to build_next_plan(), so it does not use that budget or establish the overlap behaviour, its “zero prefill budget” explanation contradicts this case's expected result.

FeathBow (approved):

PR #1035, opened 2026-09-07
refactor(qwen35): split TP worker runtime into tp_executor/worker.rs

CAICAIIs:
## What

Split the TP worker runtime out of the `tp_executor.rs` God module into `tp_executor/worker.rs`, the same seam pattern as the landed `scheduler/{backend,tp}.rs` splits (#967, #968). Pure move; behavior unchanged. Rebased over #1004/#1005/#1033/#1038 — the moved content includes the batched eager decode, decode-graph precapture, and slot-compaction machinery as worker-concern code.

**Stacked PR 1/3** — the follow-ups (#1036 responses split, #1037 scheduler steps split) sit on top of this branch; merge in order.

## Structure

`tp_executor.rs` keeps command/reply types, the `Qwen35TpExecutor` orchestration (including the precapture sweep and retired-slot tracking), plan validators, and response validation. `tp_executor/worker.rs` now owns the per-rank runtime: `TpWorker` spawn/drop lifecycle, `TpStartupGate`, the NCCL startup watchdog, `TpWorkerState`/`TpWorkerPrepared` and the per-rank command loop (`run`/`respond`/`execute_*`/`run_decode_batch{,_graph}`/`precapture_*`), the slot bookkeeping helpers, the shared decode-row view/sampling helpers, recurrent-capacity math, and the worker `CublasThreadGuard` binding. Cross-concern items reach each other via the same `use super::*;` style the landed scheduler splits use; moved items gain `pub(super)` only where the entry or its tests name them.

## Defense table (all inherited — pure move)

| Defense | Failure mode | Successor |
| --- | --- | --- |
| NCCL startup watchdog (60s abort) | startup hang blocks the process silently | inherited: `worker.rs` `spawn_nccl_startup_watchdog` |
| Bounded worker shutdown (30s → abort) | worker thread never exits | inherited: `worker.rs` `TpWorker::join_bounded` |
| Worker panic → runtime poison | one rank dies, peers block in collective | inherited: `worker.rs` spawn closure catch_unwind |
| Recurrent-capacity floor (`max_batch > 0` ensure) | silent zero-capacity executor | inherited: `worker.rs` `TpWorkerPrepared::new` |
| cuBLAS/CUDA thread rebinding per worker thread | wrong-context kernel launches | inherited: `worker.rs` `bind_worker_thread`/`CublasThreadGuard` |
| Graph-state drop before NCCL comm teardown | teardown hang on recorded collectives | inherited: `worker.rs` `TpWorkerState` field-order comment |

[...]

PR #1036, opened 2026-09-07
refactor(qwen35): split TP response contract into tp_executor/responses.rs

CAICAIIs:
## What

Split the TP response contract out of `tp_executor.rs` into `tp_executor/responses.rs`, following #worker-split. Behavior unchanged.

**Stacked PR 2/3** — based on #1035 (worker split); open against `main` because stacked cross-fork bases are not supported, so its Files-changed view includes #1035 until that merges. **Merge #1035 first.** Rebuilt over #1004/#1005/#1033/#1038: the precapture sweep keeps `validate_ack_responses` on the production path, so it stays visible rather than test-gated.

## Structure

`responses.rs` (394 lines) owns the reply envelope (`TpWorkerReply`/`TpWorkerResponse`), bounded response collection with the 300s step timeout, the poison-on-failure contract, and per-command response validation. `tp_executor.rs` keeps the executor orchestration, commands, plan validators, and tests that don't exercise the response contract. The three response-contract tests (`runtime_response_failure_poisons_executor`, `runtime_response_collection_fails_fast_when_peer_never_responds`, `disconnected_runtime_response_poisons_executor`) move with the contract they pin.

## Behavior-preserving consolidations (functional style)

- `validate_exact_rank_responses`: the push loop folds into an `into_iter().map(..).collect::<Result<Vec<_>>>()` pipeline over the same checks (count, rank range, duplicate ranks, completeness), same error strings.
- The three rank-0 payload validators (`validate_prefill/decode/unified_responses`) collapse into one `validate_primary_responses` helper over payload extractors, keeping the exact prior error messages (operation + payload noun cover all three wording variants).

## Defense table (all inherited — pure move)

| Defense | Failure mode | Successor |
| --- | --- | --- |
| 300s step timeout → abort | a rank never answers, scheduler hangs forever | inherited: `responses.rs` `recv_runtime_response` |
| Fail-fast collection on first failed rank | peers blocked in a collective keep waiting | inherited: `responses.rs` `collect_runtime_responses` |
| Channel-disconnect → poison (not silent Ok) | replies lost after worker death | inherited: `responses.rs` `recv_runtime_response` |
| Exact-rank set validation (count/range/dup/completeness) | misrouted or replayed replies | inherited: `responses.rs` `validate_exact_rank_responses` |
| Rank-0 primary payload + non-primary ack contract | secondary rank returning artifacts | inherited: `responses.rs` `validate_primary_responses` |
| Poison preserves first failure reason | root cause overwritten by cascades | inherited, pinned by the moved tests |

## Validation

- `cargo fmt --check` clean; `cargo clippy -p pegainfer-qwen35 --features qwen35 --all-targets -- -D warnings` clean.
- `cargo test --lib` green (moved response tests included); `tp2_*` GPU tests pass on 2×A100; `e2e_scheduler` green.


FeathBow:
@codex check this please

PR #1046, opened 2026-09-13
perf(qwen35): align the logits GEMM width so the output projection stops landing on an align-1 sm_75 kernel

CAICAIIs:
## Summary

The once-per-step output-projection GEMM (logits over the selection vocab) was landing on `cutlass_75_tensorop_bf16_s1688gemm_bf16_128x64_tn_align1` — an sm_75-era kernel at alignment 1 — costing **1.67 ms/decode-step, ~12% of c16 TPOT**. Root cause: Qwen3.5-4B's tokenizer-decodable vocab is **248077 — odd** — so both the GEMM M dimension and the logits output leading dimension defeat cublasLt's vectorized ampere kernels.

`bound_selection_vocab` now rounds the selection width up to the 128-token tile multiple (248077 → 248192), clamped by the checkpoint's 248,320 rows so the GEMM still reads inside the mapped weight. The extra rows are real trained embeddings, and sampling over them matches HF, which computes logits over the full checkpoint vocab. Downstream buffers and the sampler key off the same config width, so nothing else moves. Found via nsys kernel attribution; the full doc ships in the second commit (`docs/models/qwen35/decode-kernel-attribution.md`).

## Measured (1x A100-40GB, upstream/main `139d925e` + this change, vLLM 0.27.0 baseline)

| cell | before TPOT / ITL p99 / TTFT ms | after TPOT / ITL p99 / TTFT ms | vLLM 0.27 TPOT ms |
| --- | --- | --- | --- |
| 1024/256 c8 | 11.98 / 34.5 / 394 | **11.23 / ~34 / ~392** | 9.88 |
| 1024/256 c16 | 14.26 / 36.4 / 691 | **13.51 / ~36 / ~682** | 11.00 |
| QPS16 1024/128 | 23.77 / 41.7 / 1264 | **22.68 / 40.6 / 1157** | 23.60 |

Measured on top of the merged `auto`+`stream` posture (#1033/#1034): c16 TPOT −5.3%, c8 −6.3%, QPS16 −4.6% (back below vLLM), throughputs +4–6%, TTFT/ITL unchanged or better.

## Review follow-ups

Three commits sit on top of the perf commit:

1. **Padded rows must not be selectable** (Codex P1). Tile-aligning the selection width made ids `248077..248192` eligible for greedy/sampled selection. Each logits GEMM is now followed by the existing `SuppressIds` kernel forcing those rows to `-inf`, so selection scans exactly the pre-fix decodable vocab. The suppress list is uploaded once at load and runs inside the decode CUDA Graph (no per-step host cost).
2. **Padded rows must not move the routing decision** (FeathBow). `select_batch` read the routing width from `logits.hidden_dim`, so `effectively_greedy`'s `top_p <= 1/vocab` threshold moved from `1/248077` to `1/248192` — a request with `top_p` between those values would drop off the deterministic argmax path onto the rejection sampler, which can pick either member of a bf16-tied top. `SampleScratch` now carries the arena width *and* the semantic selection width separately (`new` keeps `selection_width == vocab`; qwen35 builds through `with_selection_width(selection_vocab, decodable_vocab)`), so a padded arena routes exactly the rows an unpadded one would.
3. **The mask cannot be skipped at a logits site** (FeathBow). The aligned GEMM and the pad-row mask were two statements repeated at three sites, so a fourth logits site could run the GEMM and skip the mask — and the failure is silent, because the pad rows are trained embeddings with plausible logits. `Qwen35Model::output_logits_into` now owns both statements and the three sites call it, and `output_projection()` is private (the GEMM-tuning helper is its only remaining caller), so the GEMM is not reachable without the mask. The ops re-exports the old split needed are gone with it.
4. A style pass: comments cut to the constraint that cannot be recovered from the code, a test-only getter and a construction-guard test for a programmer-error assertion removed, and the two invariants recorded in the attribution doc rather than only in the diff.

## Testing

Verified on 1x A100-SXM4-40GB (`dgx`, CUDA 11.5, sm_80), branch at `fe48b666`, model `Qwen3.5-4B`, built with `--features qwen35` and `PEGAINFER_TRITON_PYTHON` on a Triton 3.7.1 venv.

- lib tests for the touched packages (`pegainfer-sample`, `pegainfer-qwen35`) **102 passed / 0 failed / 7 ignored**, including the two `bound_selection_vocab` alignment/clamp unit tests.
- **`pad_columns_are_suppressed_on_the_model_logits_paths`** — new, and the one test that covers the mask invariant on the model itself rather than on a synthetic arena. It drives the real checkpoint through both logits paths (`batch_prefill_logits` and `batch_decode_graph`) and requires every column in `decodable_vocab..selection_vocab` (248077..248192) to come out `-inf` while the decodable prefix does not. **Mutation-checked**: commenting the mask out of `output_logits_into` fails it with `prefill: pad id 248077 survived selection (want -inf)`, so it is a test with teeth rather than a tautology.
- `pegainfer-sample` integration suite **11 passed / 0 failed** (was 9), including the new `padded_arena_width_does_not_suppress_the_argmax_routing`: a 512-column arena with 256 decodable and `top_p = 1/256` exactly on the boundary, where the arena-width scratch routes to the sampler and can return the bf16-tied peer, and the decodable-width scratch keeps the deterministic argmax for all 64 seeds.
- `hf_golden_gate` **2/2 passed** (TP1; the TP2 graph variant is the pre-existing host-environmental NCCL precapture issue noted above). It replays fixed teacher-forced sequences and compares logprobs against HF fixtures — it never selects a token, and the row logsumexp it scores is mask-invariant (`exp(-inf) = 0`), so it is evidence that the aligned width leaves prefill/decode numerics alone, not evidence about the mask. That distinction is now written down in the attribution doc instead of being implied by a blanket "output equivalence" claim.
- `e2e_scheduler` **2/2 passed**.
- **Production A/B, same session on the same GPU** (base `139d925` vs this branch, real server on GPU 7, `vllm bench serve` random 1024/256, greedy, seed 42, zero failed requests in every cell):

  | cell | base TPOT / TTFT / ITL ms | branch TPOT / TTFT / ITL ms |
  | --- | --- | --- |
  | c16 | 14.90 / 738.7 / 14.84 | **14.09 / 718.7 / 14.04** |
  | c8 | 12.23 / 379.1 / 12.19 | **11.57 / 399.4 / 11.52** |

  Both cells improve ~5.4%, so the speed-up reproduces against a same-session baseline rather than only across sessions, and the review follow-ups do not give any of it back. Re-measured after the `output_logits_into` refactor (at `05ffa360`; the only later commit makes `output_projection` private): c16 `14.12 ms` / TTFT 708.5, c8 `11.48 ms` / TTFT 377.2, 0 failed.

- **HTTP invariant probe** against the branch server, which keys the observed contract rather than raw text equality (greedy decode on this host is not bit-reproducible, see below): across 4 prompts with `logprobs=1`, all 96 emitted token strings are real tokenizer tokens with zero empty/undecodable strings, so no id at or above the decodable `248077` ever reaches the wire; `temperature=1, top_p=4.03e-6` (the case FeathBow raised) returns **byte-identical text to greedy** on every prompt; and `top_p=0.9` differs from greedy on 3 of 4 prompts, confirming the wide-nucleus sampling path is still live rather than everything being pinned to argmax.
- Widths confirmed against the checkpoint: decodable `248077`, weight rows `248320`, aligned selection `248192`, so the suppressed region is exactly `248077..248192` (115 rows) and the GEMM stays inside the mapped weight.
- Kernel check on a clean c16 trace: the align-1 kernel no longer appears; the replacement ampere GEMM family drops the step cost by ~0.7–1.3 ms.

### What was NOT verifiable, and why

- **Raw output differentials between binaries are not usable as evidence here.** Running the *unmodified* base binary twice under identical parallel load produced different greedy output (4 differing rows), and it also differed run-to-run at concurrency 1. Greedy decode on this setup is not bit-reproducible, so a base-vs-branch text diff measures runtime noise, not this change. The deterministic gates above (`hf_golden_gate`, the sampler tests) are the sound evidence for output behaviour.
- This round's new test and the refactor were verified with `--test-threads=1`; the flake below is why. The full `cargo test --workspace --lib` could not run: the workspace build pulls the `moe` feature, which needs NCCL >= 2.30.4, and every NCCL on this host is 2.29.7. The touched-package lib tests were run instead.
- **The sampler integration suite must run single-threaded on this host; in parallel it flakes, on the base commit too.** Default (parallel) execution failed 1 of 4 runs on this branch and 1 of 6 runs on unmodified `139d925`, always as `batch sampling kernel failed with error 1` from whichever test happened to overlap; with `--test-threads=1` the suite passed 4/4 on both. Earlier in this PR that flake was misattributed to `min_p_row_takes_the_sampler_path_and_filters` specifically — it moves between tests, which is what pointed at cross-test GPU contention rather than any one test. Pre-existing, not caused by this PR, and not fixed here. Every serial run quoted above is `--test-threads=1`.

## Claim boundary

Single-run cells on one GPU; the GDN/full-attn decode kernel gaps (2.2×/2.4× per layer-step, 6× at bs1) documented in the attribution doc remain open and are the next optimization targets. The perf table's "before/after" cells predate the review follow-ups; the re-run above is within run noise of the original fix.


pegainfer-qwen35/src/config/model.rs
```diff
@@ -169,6 +173,7 @@ impl Config35 {
         // up to the tile-aligned multiple — the checkpoint's remaining rows are
         // real trained embeddings and stay inside the mapped weight.
         let aligned = effective_vocab.next_multiple_of(128);
```
FeathBow, line 177: Could we keep using the decodable vocabulary size when deciding whether a request takes the argmax path? This changes `selection_vocab` from 248077 to 248192, and `select_batch` passes that padded width from `logits.hidden_dim` into `effectively_greedy`.

For example, with `temperature=1`, `top_k=-1`, and `top_p=4.03e-6`, the old `1 / vocab` threshold routes the request to argmax, while the new threshold sends it to the rejection sampler. That can change which token wins when the largest BF16 logits are tied. The existing `tiny_top_p_routes_to_argmax_even_under_bf16_ties` test covers why this distinction matters.

The `-inf` mask keeps padded IDs out of the result, but it doesn't preserve this routing decision. Could we use `decodable_vocab` for that decision while keeping the aligned width for the GEMM and buffer layout?

pegainfer-qwen35/src/weights.rs
```diff
@@ -334,6 +352,16 @@ impl Qwen35Model {
         self.lm_head.as_ref().unwrap_or(&self.embed_tokens)
     }
 
+    /// Force the tile-alignment pad rows of a logits buffer to -inf.
```
FeathBow: Right now the aligned GEMM and the pad-row mask are two separate statements repeated at three sites (prefill.rs:120-128, batch_decode.rs:697-705, batch_decode.rs:776-784). A fourth logits site can call the GEMM and skip the mask, and nothing in the suite would catch it. The config unit tests only assert config values, the new sampler test builds its own arena, and hf_golden_gate compares HF top-K, which is only sensitive to the mask if HF's stored top-K lands inside 248077..248192.

That matters more than usual here because this commit's own doc says the failure is silent: the pad rows are trained embeddings with plausible logits, so a missed mask surfaces as an undecodable id on the wire and fed back into later decode steps, not as a crash.

Would you consider folding the two statements into one method so the invariant cannot be skipped?

FeathBow (commented):
Somethings I'd like to settle before merging. The suppression invariant is a call-site convention today and nothing in the suite would catch a fourth logits site that skips it, which matters because this PR's own doc says the failure is silent, btw some parts annotations outdated.

FeathBow:
@codex review please

FeathBow (approved):
non-blocking but docs outdated.

PR #1052, opened 2026-09-16
feat(gemma4): generate the hd512 global prefill kernel and serve it behind a knob

### LINKED ISSUE #1051: gemma4: the global hd512 prefill runs at a fifth of the Hopper roofline
## Description

Gemma 4's ten global layers attend at head_dim 512. The vendored FlashInfer FA2 path serves that shape by time-sharing K and V in one shared-memory buffer, V loaded only after Q·K^T, with a 64-key CTA tile — its own comment says so, to fit head_dim 512 on the 99 KB SKUs (SM86/89/120/121). There is no architecture condition on that choice, so on a Hopper card with 227 KB of shared memory the kernel pays twice for a budget that is not its own.

Measured on one GH200 at the chunked prefill's real shapes (one 8192-row chunk against everything before it, 32 query heads over 4 KV heads, bf16, causal) with the roofline taken in the same process (686 TFLOP/s sustained bf16 GEMM): FlashAttention 4 holds 44–49% of that roofline across the range, the kernel we serve 19–28%, 1.46× to 2.22× slower than FA4 over the same page table. At 31B the global attention is the largest single account of a long prefill, and this one kernel accounts for essentially the whole of the long-prompt gap against vLLM.

## Proposed Plan

- A TileLang-generated hd512 varlen paged prefill kernel for sm_90, generated at build time next to the crate that serves it, through the same `PEGAINFER_GEMMA4_TILELANG_PYTHON` door the routed lane already uses; the AOT section of the kernels build takes a family table so a second family costs a row, not a branch.
- The global pool paged at the attention key block, so a key tile is one copy.
- A runtime knob, `PEGAINFER_GLOBAL_ATTN=tilelang`, selecting the generated kernel for the global family's prompt rows; unset or `off` keeps the incumbent kernel and byte-identical serving.
- A paired harness timing both kernels on one pool and one plan, and an oracle holding the generated kernel against the incumbent on the checkpoint.

## Acceptance Criteria

- The generated kernel is bit-identical to its JIT twin on every AOT case, including hostile padding past the plan, and within 0.002 of an fp32 reference.
- The oracle on the checkpoint holds the generated prefill against the incumbent within its line; the split state's oracle numbers do not move.
- A same-host, same-binary A/B (the knob is the only difference) shows the prefill account moving and nothing else: TPOT and peak memory unchanged.

FeathBow:
## Description

Closes #1051

**The shape was never the problem; the budget was.** FlashInfer's FA2 path serves head_dim 512 by time-sharing K and V in one shared-memory buffer with a 64-key tile, a layout its own comment ties to the 99 KB SKUs. On a GH200 that leaves the kernel at 19–28% of the measured roofline where FlashAttention 4 holds 44–49%, and at 31B the ten global layers' prefill is the largest account of a long prompt.

**A generated kernel, built where the line lives.** `pegainfer-gemma4/kernels/` carries a TileLang definition of the varlen paged prefill at head_dim 512 and a generator that emits the CUDA, a launcher table and a host stub at build time; the kernels crate's AOT section now takes a family table, so this family sits beside the routed lane's rather than duplicating its build logic. A build without TileLang carries a stub that refuses to serve, and says so.

**The global pool is paged at the key block.** The global family's page size moves to 64 so one key tile is one TMA copy; the sliding pool is untouched in this change.

**One knob, three lines it can fall on.** `PEGAINFER_GLOBAL_ATTN=tilelang` routes the global family's prompt rows through the generated kernel; unset or `off` keeps the incumbent and byte-identical serving. The walk inside the kernel is bounded by the CTA count the host computed for the plan, so a CTA past the plan does nothing rather than reading past it.

**Measured, not assumed.** A paired harness (`hd512_prefill_bench`) times both kernels on one pool and one plan; the serving oracle holds the generated kernel against the incumbent on the checkpoint.

### Test Env

- One GH200, CUDA 12.6 toolchain for the kernels crate, TileLang 0.1.12 for the generated family; Gemma 4 31B bf16 for the serving A/B and the 31B oracles, the pinned 12B checkpoint and fixtures for the maintained gate suite.

### Verification

Bound to the tip of this range; every run on the box above.

- AOT gate: every prefill case bit-identical between the AOT and JIT builds, |Δ| ≤ 0.002 against an fp32 reference, hostile padding past the plan left unwritten, the refusal rows all refused.
- Paired kernel harness at the 31B global shape: 32K chunk 78.1 → 32.6 ms (2.40×), 163K chunk 423.6 → 180.8 ms (2.34×).
- Serving oracle on the checkpoint: the generated prefill within |Δ logit| 0.5 of the incumbent at 12B and 0.625 at 31B against a line of 2.0, the incumbent bit-stable run to run; the maintained gate suite green on the 12B fixtures with the split state's oracle numbers unchanged.
- Same-binary A/B on 31B, the knob the only difference, single request, four prompt lengths, four rounds, 12 kept requests per cell, round spread ≤ 0.7%, every paired interval narrower than a point and off zero: TTFT −9.2 .. −8.5% at 10,602 tokens, −20.4 .. −19.8% at 40,002, −28.6 .. −27.8% at 81,653, −36.7 .. −36.6% at 163,336; E2EL −2.1 .. −2.0% to −33.2 .. −33.0% over the same cells. TPOT −0.3 .. +0.1% (decode does not touch this kernel) and peak device memory identical at 82,655 MiB on both arms, which is what a prefill-only change should look like. Against the 12B-era prediction from the kernel ratios the saving is 26.3 s of a predicted 29.0 s at 163K; the remainder sits in the short-kv chunks where the old kernel was least behind.
- Three-arm fair A/B against vLLM on the same card (vLLM's own bench client, both engines at the 165,888 ceiling vLLM negotiated, 8192-token chunk from vLLM's log, KV bf16 on both sides, prefix caching off): the generated kernel moves the TTFT crossover to about 40K tokens and the E2EL crossover to about 81.6K; at 163,336 TTFT −12.0 .. −11.7% and E2EL −7.9 .. −7.6% against vLLM, with TPOT untouched and still behind (+26.8 .. +27.3%), which the follow-up changes take up.
- `cargo fmt --check`, `clippy -D warnings` on the kernels and gemma4 crates including tests and examples, lib tests green.


FeathBow:
@codex review again plx

FeathBow:
@codex review again

FeathBow:
@codex review

PR #1054, opened 2026-09-17
perf(qwen35): carry the GDN decode state slice in registers across the kv_mem reduction

CAICAIIs:
## Summary

The gated delta rule decode kernel did its two passes (decay, then rank-1 update) with a global round-trip between them: pass 1 stored the decayed state slice back to global memory, the kv_mem reduction consumed it, and pass 2 reloaded it to apply delta·k and store again. The key slice is only 32 floats per thread — it lives in registers across the reduction instead, so global memory sees one read and one write per state element.

## Measured (1x A100-40GB, vLLM 0.27.0 baseline)

| cell | before TPOT ms | after TPOT ms | delta | vLLM 0.27 TPOT ms |
| 1024/256 c16 | 13.51 | **12.98** | −3.9% | 11.00 |
| bs1 1024/256 | 9.77 | **9.36** | −4.2%* | 8.32 |

- c16 "before" = upstream/main + #1046 (single-variable A/B against this change).
- bs1 "before" = upstream/main at `139d925` without #1046: the cell bundles this change with #1046's alignment effect at bs1.
- c16 output throughput: 936 → 1000 tok/s.

The kernel's per-layer-step time was 97.0 µs vs vLLM's FLA `fused_recurrent_gated_delta_rule_packed` 43.6 µs; the register-carried state removes the extra global round-trip that was the difference from FLA's bandwidth-floor implementation. Found via nsys kernel attribution (a fuller doc covering the decode kernel gaps is being prepared separately).

## Testing

- The rewrite is arithmetically identical — same f32 ops in the same order, state values carried in registers instead of a global round-trip.
- `hf_golden_gate` TP1 2/2; lib tests 102 passed.

## Claim boundary

Single-run cells on one GPU; draft pending review. The kernel still has no sm_80 tuning pass; the remaining GDN gap to FLA and the c16 GEMM family deficit (+2.8 ms/step) stay open in the attribution doc's queue.


FeathBow:
@codex review

FeathBow (approved):

PR #1066, opened 2026-09-19
feat(build): route nvcc through an optional launcher

luicarus:
## What this does

`RUSTC_WRAPPER=sccache` only intercepts `rustc`. `pegainfer-kernels/build.rs` invokes `nvcc` directly, so the CUDA translation units under `pegainfer-kernels/csrc` bypass sccache and are rebuilt from cold on every CUDA lane. This is why a lane can report a high overall sccache hit rate while still spending several minutes inside Cargo.

sccache supports `nvcc` by decomposing the invocation through `nvcc --dryrun` and caching the underlying `cicc`, `ptxas`, `cudafe++`, and host-compiler calls. For that to work, sccache has to be the process launching `nvcc`.

Previously, `CudaToolkit` exposed `nvcc` as a `PathBuf`, and the build script constructed `Command`s from that path directly in four places. This PR replaces that with a command description that can carry an optional launcher prefix:

```rust
pub struct NvccCommand {
    pub launcher: Vec<OsString>, // empty by default
    pub program: PathBuf,
}
```

All four invocation paths — the arch probe, gencode probe, kernel-lab shared link, and translation-unit worker pool — now go through `NvccCommand::command()`. This applies the launcher consistently to every `nvcc` invocation instead of requiring individual call sites to opt in:

```text
PEGAINFER_NVCC_LAUNCHER=sccache  ->  sccache <nvcc> -c kernel.cu ...
unset                            ->  <nvcc> -c kernel.cu ...
```

`CudaToolkit::discover` also emits `rerun-if-env-changed` for `PEGAINFER_NVCC_LAUNCHER`, so changing the launcher invalidates the build-script result as expected.

## Scope

This PR has two parts:

1. add the launcher seam in `pegainfer-build` / `pegainfer-kernels`;
2. set `PEGAINFER_NVCC_LAUNCHER=sccache` on the CUDA CI job so the path is exercised in CI.

The variable remains unset by default, including on developer machines, so existing `nvcc` invocation behavior is unchanged unless a launcher is explicitly configured.

### Unit tests, no GPU required

`pegainfer-build` already has a `TempTree` fake-toolkit helper and runs in the CPU unit-test job, so the discovery and command-construction paths can be tested without CUDA hardware.

Local verification:

```text
cargo test -p pegainfer-build --release --locked --lib
running 15 tests
test result: ok. 15 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

Four of those are new and go end to end over a fake toolkit tree:

| Test | Covers |
| --- | --- |
| `discovered_toolkit_invokes_its_own_nvcc_without_a_launcher` | discovery resolves `bin/nvcc` and invokes it directly when no launcher is configured |
| `discovered_toolkit_places_the_launcher_before_its_nvcc` | discovery with a launcher produces `["sccache", "<tree>/bin/nvcc"]` |
| `launcher_prefix_survives_arguments_appended_by_the_build_script` | build-script arguments such as `-c kernel.cu -o kernel_cuda.o` are appended after the launcher and `nvcc` path |
| `empty_launcher_env_is_treated_as_unset` | an exported empty value does not become an empty executable name |

Because the launcher environment variable is process-wide, these tests serialize access with a mutex and restore the previous value afterward.

### A bug caught by the tests

The first implementation of `command()` used `Command::new(self.program)` on the no-launcher path and then appended `self.program` again, which would have changed the existing behavior into:

```text
nvcc nvcc -c ...
```

The new tests caught this immediately:

```text
left:  ["/usr/local/cuda/bin/nvcc", "/usr/local/cuda/bin/nvcc"]
right: ["/usr/local/cuda/bin/nvcc"]
```

The implementation now branches on `launcher.split_first()`, preserving the existing direct-`nvcc` path when no launcher is configured.

## What is not yet measured

The purpose of this PR is to measure the actual CUDA cache behavior in CI. The expectation is that the shared translation units stop being rebuilt from cold, but two things need to be verified from the PR runs:

- **CUDA-specific cache hits.** The overall sccache rate is dominated by Rust compilation and can look healthy even when CUDA compilation misses. If CUDA hits remain low, the first thing to inspect is path stability, particularly `OUT_DIR`, because sccache's `nvcc` implementation normalizes paths used by `cicc`.
- **Decomposition overhead.** sccache invokes `nvcc --dryrun` before consulting the cache. If CUDA objects rarely hit, that extra work could offset part of the benefit.

Once this PR has completed CI runs, I'll post CUDA hit/miss data and the relevant per-step timings here rather than claiming an improvement up front.

## Relation to #1048

This is a follow-up to #1048, which removed the redundant CUDA compile jobs and consolidated the CUDA workflow into one row per model line.

The review on #1048 identified direct `nvcc` compilation as the remaining repeated work on the CUDA critical path and suggested putting sccache in front of those invocations. This PR implements that launcher seam.

It is based on `main` after #1048, so enabling the launcher in CI requires only one additional `env:` entry on the consolidated CUDA job.

Tracked under #975. The launcher approach was suggested by @FeathBow during review of #1048.


luicarus:
Second run is green, and the launcher is behaving as intended.

Before this change, direct `nvcc` invocations never passed through sccache, so the stats had no CUDA cache category at all. The first launcher-enabled run therefore gives us a mostly cold baseline for this path; the second run shows what happens once those CUDA artifacts are available to the cache.

### CUDA cache, by job

| Job | First run | Second run |
| --- | ---: | ---: |
| `Qwen3 CUDA Clippy` — CUDA hit rate | 0.00% | **95.00%** |
| `Gemma 4 CUDA Clippy` — CUDA hit rate | 55.56% | **100.00%** |
| `Qwen3.5 CUDA Clippy` — CUDA hit rate | 5.00% | **100.00%** |

`CUBIN` and `PTX` show the same pattern. For example, Qwen3.5 moves from 20% / 15% to 100% / 95%.

### Observed work and timing

| | First run | Second run | Change |
| --- | ---: | ---: | ---: |
| CUDA `Compilations` (3 jobs) | 265 | **122** | **−143 (−54%)** |
| CUDA `Cache misses` (3 jobs) | 51 | **1** | **−50** |
| Summed CUDA job runtime | 865s | **696s** | **−169s** |
| Workflow wall time | 5m52s | **4m23s** | **−89s** |

Per job, the CUDA compilation counts move from 94 → 43, 75 → 49, and 96 → 30.

The cache effect itself is clear from the hit/miss data: after one launcher-enabled run, all three CUDA lanes are almost entirely hitting cached CUDA work.

### Timing signal

The timing change is also consistent with that cache effect, although two runs are not enough to assign a precise steady-state speedup.

The CPU jobs, which do not use the new `nvcc` launcher path, became slower between the same two runs:

- `CPU Clippy`: 111s → 137s
- `Simulated frontend E2E`: 156s → 183s
- `CPU unit tests (pegainfer-sim)`: 112s → 142s

Meanwhile, the summed CUDA job runtime fell by 169s.

That makes a global "the second run just got faster runners" explanation unlikely, but I would still treat the 89s wall-time reduction as an observed result rather than a stable estimate from only two samples.

### Runs

- First launcher-enabled run: [35431322883](https://github.com/pegainfer-project/pegainfer/actions/runs/35431322883) — failed only on `Rust formatting`; the CUDA jobs completed, so their cache/timing data is still usable
- Second run: [35483044289](https://github.com/pegainfer-project/pegainfer/actions/runs/35483044289) — 13/13 green

### Caveats

- Two runs are a small sample. The CUDA cache behavior is consistent across all three lanes, but I would not assign a precise steady-state wall-time saving from these two points alone.
- `Cache write errors` are still present: 3 per CUDA job on the second run, down from 25 / 7 / 53 on the first. That is the separate cache-write issue already listed as a follow-up in #975; this PR does not change cache-backend behavior.
- The first run already has the launcher enabled, so it is a cold-cache baseline for the new path rather than a true pre-launcher baseline. The pre-launcher measurements in #975 come from a different commit and are not directly interchangeable with these two runs.

FeathBow (commented):
The seam works and CI shows it. On run 35483044289 the three CUDA lanes hit 19/20, 27/27 and 20/20 on CUDA compilations; on main's run 35502584091 the stats have no CUDA category at all. The overall rate barely moves, 96.70% to 96.79%.

The second one is still open, and I think it decides how much this actually bought. sccache runs `nvcc --dryrun` before every invocation, and the same lane reports 228 non-cacheable calls. The clippy step is still 149s with Rust at 96% hits and CUDA fully cached, and nobody has broken that number down. Could you measure the decomposition cost you flagged — a lane with the launcher unset against one with it set, at a warm cache, or just the per-step timings — so the net is a number rather than an inference?

The other half of the picture belongs in the description. Setup is now 42% of that lane: 110s of 263s, split 52 checkout, 40 CUDA toolkit, 9 toolchain, 5 apt, paid three times over. #975 asks to fuse duplicate gates and reduce repeated setup; #1048 did the first, this does the nvcc work that came out of its review, and the repeated setup is untouched.

Code, four narrowings:

- `launcher` can be `Option<OsString>`. `into_iter().collect()` makes it 0 or 1 elements always, so `split_first`, `args(rest)` and the `join` in `display` serve an unreachable case. `RUSTC_WRAPPER` also takes a single program without splitting.
- Wire `display()` up rather than delete it. Two diagnostics in build.rs say "nvcc" while a launcher is running `sccache nvcc`.
- Move the env read into a `launcher_from_env()` that `discover()` calls once, and let `new` take the launcher. That takes the four `unsafe` env writes in the tests down to one.
- Then drop the three fake-tree tests that duplicate the directly constructed ones, and `launcher_is_read_from_the_environment` outright — it sets no variable, and `argv.len() == launcher.len() + 1` restates what `command()` does.

luicarus:
I ran the direct-`nvcc` vs warm-sccache comparison you asked for and updated the PR description with the broader timing picture, including the remaining per-runner setup cost.

For the A/B, both runs use the same implementation commit and workflow except for the launcher:

- direct `nvcc`: [35824811409](https://github.com/pegainfer-project/pegainfer/actions/runs/35824811409)
- warm sccache: [35828158053](https://github.com/pegainfer-project/pegainfer/actions/runs/35828158053)

| Job | Direct `nvcc` | Warm sccache | Observed difference |
| --- | ---: | ---: | ---: |
| `Gemma 4 CUDA Clippy` | 218s | 104s | −114s |
| `Qwen3 CUDA Clippy` | 175s | 125s | −50s |
| `Qwen3.5 CUDA Clippy` | 200s | 109s | −91s |
| **Summed Clippy step** | **593s** | **338s** | **−255s (−43%)** |

The launcher-enabled side includes the `nvcc --dryrun` decomposition overhead, so this is the end-to-end comparison rather than an inferred cache saving.

There is still hosted-runner variance, so I would not treat 255s as a precise steady-state number. The stronger signal is that all three CUDA lanes move in the same direction and the warm run reports 90–100% CUDA cache hits.

I also broke down the representative 149s `Qwen3 CUDA Clippy` step: about 40.6s is dependency fetch, 60.0s is the `build.rs`/CUDA window, and 48.5s is the remaining Rust build. The launcher can only affect work inside that 60s window.

I've updated the description to include these measurements and the other half of the picture you called out: roughly 110s / 263s (~42%) of the representative lane is still repeated setup, which this PR leaves untouched.

xiaguan (changes requested):
Thanks for this, and for the careful A/B. The −43% on the CUDA lanes is a real win and we want it in.

One request before merging: the seam could be much smaller. The core is just "prefix nvcc with an optional launcher", so a new `NvccCommand` type with four methods and ~100 lines of argv tests feels heavier than the problem. Could you simplify it to roughly:

- keep `CudaToolkit::nvcc: PathBuf` as before;
- add one `fn nvcc_command(&self) -> Command` on `CudaToolkit` that reads `PEGAINFER_NVCC_LAUNCHER` (empty = unset) and builds the command;
- route the four call sites through it (that part is already right);
- keep a single small test if you like, and trim the doc comments to a line or two.

The CI change and `rerun-if-env-changed` stay as they are. Happy to merge once it's slimmed down 🙏

luicarus:
Thanks, this was a good call. I slimmed the seam down substantially.

One small deviation from the sketch: I kept `nvcc_command` as a local helper taking the nvcc path rather than making it a `CudaToolkit` method.

`nvcc_accepts_gencode` sits several helpers deep and is reached through the `glm52_*_arch_args`, `k3_*_arch_args`, and `tilelang_gencode` paths. Those helpers otherwise only need the nvcc path, so making this a `&self` method would require threading `&CudaToolkit` through a number of signatures solely to reach the eventual spawn site.

With the local helper, those signatures remain unchanged and the implementation change is limited to the four actual launch sites:

```text
Command::new(nvcc)  ->  nvcc_command(nvcc)
```

`CudaToolkit::nvcc` is a `PathBuf` again, the CI env entry and `rerun-if-env-changed` behavior are unchanged, and the `NvccCommand` type plus the larger argv-test surface are gone. The net diff against `main` is now **+18/−4 across three files**.

I also updated the PR description: the A/B results are now under `Measured impact`, and the repeated setup cost you pointed out has its own section.

The slimmed-down version is green and the cache behavior is unchanged: CUDA hit rates remain at 90–100%, with fewer actual compilations than the previous warm run (27 vs. 44).

That particular run was slower in wall time, but the cache counters do not indicate a cache regression: `Average compiler` increased from 1.077s to 12.375s while the number of actual compilations fell. The same run also shows unrelated timing variance (`Install build dependencies` took 24s vs. ~4–5s previously), so I would not infer a performance regression from this single hosted-runner sample.

I don't have permission to re-run the workflow independently, but the functional/cache result of the simplified implementation is green and consistent with the earlier runs.

FeathBow (commented):
I recommend moving the `rerun-if-env-changed=PEGAINFER_NVCC_LAUNCHER` line into pegainfer-kernels/build.rs next to the read. It belonged in `discover()` while the read lived there. Now pegainfer-cupti reruns its build script on a launcher change without ever running nvcc, and pegainfer-build drops out of the diff once it moves.

Optional: if the launcher can't be spawned, every message still says nvcc. The build does fail, so it's only the wording. Including the spawn error in the worker's "Failed to run nvcc for X" panic would be enough.

Issue #1068, opened 2026-09-20
hardening: config and weight loaders trust fields and shapes they never verify


sysuTobo:
## Description

The model configuration and weight loaders accept input they never check, so a checkpoint that disagrees with our assumptions loads successfully and then produces plausible-looking garbage. Three instances of the same failure class:

- **Unknown config fields vanish.** `TextConfig` / `RawConfig` / `RopeParameters` (`pegainfer-qwen35/src/config/model.rs`) have no `#[serde(deny_unknown_fields)]`, and `RopeParameters` deliberately reads only two keys. A new generation can rename or repurpose any ignored key — `rope_type`, `mrope_section`, `partial_rotary_factor` at the wrong level, a new `output_gate_type` with real effect — and validation still passes. The other model lines parse configs the same way. This is not hypothetical: Qwen3.8-27B's `config.json` carries `text_config.output_gate_type`, which our `TextConfig` does not declare and silently discards. That field happened to be inert upstream (verified across transformers 5.2.0 / 5.8.0 / 5.17.0), but nothing in this repository would have told us that — had it meant something, we would have served a silently wrong model.
- **1D tensors are checked for rank and dtype, never for length.** `load_tensor_1d` / `load_tensor_1d_f32` (`pegainfer-core/src/weight_loader.rs`) validate dtype and `shape().len() == 1`, then upload whatever length arrived. `input_layernorm`, `post_attention_layernorm`, `q_norm` / `k_norm` (head_dim), `dt_bias` / `A_log` (value heads) and the flattened `conv1d` all have an extent the config already determines, and none of them is compared to it. A short vector is an out-of-bounds read inside the kernel that consumes it.
- **2D projections sized themselves from the file.** On the single-GPU path `row_shard_if_needed` / `col_shard_if_needed` / `gated_q_proj` / `linear_in_proj_qkv` fell back to a `tensor_2d` that takes the tensor's own `(rows, cols)` as truth, so only `lm_head` was verified against the config; under TP the shard-range checks caught it, on one card nothing did. This slice is fixed on `feat/qwen38-27b`, which threads the config-derived extents through those four paths plus `embed_tokens` with a unit test pinning the expected numbers to shapes read off the real 27B checkpoint. The two bullets above stay open.

## Proposed Plan

[...]

PR #1069, opened 2026-09-20
feat(qwen35): serve Qwen3.8-27B on the existing qwen35 line

### LINKED ISSUE #1067: qwen35: the Qwen3.8-27B checkpoint does not load, and nothing could check it if it did
## Description

Qwen3.8-27B is the next checkpoint on the Qwen3.5 architecture. Diffed field-by-field and tensor-by-tensor against `Qwen/Qwen3.5-27B` (revision `1d4bf0f2ff6012fd82039f2fa52739d0dd7c60c0`), it is the same model: identical `model_type: qwen3_5` and `architectures`, identical geometry (16 full + 48 linear layers, hidden 5120, head_dim 256, 24 Q / 4 KV, 48 V / 16 QK @ 128, vocab 248,320, ctx 262,144), zero shape deltas across all 1,199 tensors, and only inert config-field differences. It should ride the existing `qwen35` line and its probe-based dispatch. Three things stop that today, plus a fourth that stops it from fitting once loaded:

- **The loader rejects its GDN scalars.** Qwen3.8 stores `linear_attn.A_log` [48] and `linear_attn.norm.weight` [128] as bf16 for all 48 linear layers where Qwen3.5 stores f32 — a transformers 5.x save-time cast policy change, exactly the 16,896-byte `total_size` gap between the two checkpoints. `tensor_f32_cow` demands f32, so the load dies with `expected dtype F32, got BF16`. The reference implementation upcasts the same tensors at the point of use.
- **The fixture table has no key for it.** `fixture_size_name` keys logits goldens on `(hidden_size, num_hidden_layers)`, and both 27B checkpoints are `(5120, 64)` — a Qwen3.8 oracle would alias the Qwen3.5 one with no way to name it apart.
- **Its chat template has no parity gate.** The template adds `reasoning_effort` / `preserve_thinking` kwargs over the Qwen3.5 family, and no test holds our renders against the checkpoint tokenizer's.
- **TP ignores the caller's capacity, and at 27B that decides whether the model fits.** `Qwen35TpExecutor::from_runtime_with_capacity` builds every rank through the constructor that reserves the `MAX_BATCH` default of 64 decode slots; the recurrent-state pool is sized at load time, so `--max-batch 8` still reserves 64 slots' worth — 9,396 MB at this geometry — and TP2 fails with `insufficient device memory` on two 46 GB cards that could serve the requested capacity. Budget measured from the safetensors headers: text tower 53.79 GB bf16 (about 26.9 GB per rank at TP2), full-attention KV 64 KiB per token.

One field deserves a warning label: `text_config.output_gate_type: "swish"` is new, while this line's attention output gate is hardcoded sigmoid. It is inert metadata, on three grounds: the field has zero readers in transformers' `qwen3_5` code at 5.2.0, 5.8.0 and 5.17.0 (its only consumer is the unrelated `qwen4_exp`, which accepts only `{"sigmoid","silu"}`); swish is silu, which both HF's `Qwen3_5RMSNormGated` and our `rms_norm_gated` kernel already implement; and a two-hypothesis probe — sigmoid vs silu patched into HF at the module's single gate call site, same inputs both ways — puts sigmoid at NLL 1.822 with coherent generations against silu's 14.268 token soup, with a Qwen3.5-0.8B control confirming the instrument measures what it claims. Do not implement "swish".

## Proposed Plan

- Widen bf16 to f32 for exactly the two GDN scalar vectors — an exact cast, so both storages reach the kernels as the same f32 values — and leave `tensor_f32_cow` strict for every other consumer.
- Key the fixture table by `(generation, size)`, probing the generation from `text_config.output_gate_type`; the existing `config_sha256` + `model_revision` asserts stay as the real mispairing guard. Commit short and long Qwen3.8-27B fixtures dumped at the pinned revision, and make the dumper refuse a non-finite reference: transformers' eager GDN fallback returns all-NaN logprobs on this geometry without flash-linear-attention, which today reads as a passing gate.
- Thread the caller's `--max-batch` into the TP per-rank build, matching the single-GPU path.
- Add a chat-template parity gate — plain, multi-turn, unicode, three reasoning efforts, thinking disabled, disabled-with-effort, preserve_thinking — reusing the gemma4 parity machinery, with the checkpoint and template digests recorded in the fixture.
- Assert config-derived extents on the unsharded load path while touching it; one slice of #1068.

## Acceptance Criteria

- Qwen3.8-27B loads and serves at TP2 at the requested capacity; the short and long logits gates hold within the line's existing tolerances (no new per-size constants), sequential and batched variants against the same oracle.
- The chat parity gate matches the HF render byte-for-byte on all committed cases.
- Qwen3.5 behaviour is unchanged: same fixtures, same probe outcome, same defaults, strict dtype checks elsewhere untouched.
- Follow-ups stay explicit: the 333 `model.visual.*` tensors stay unloaded (as on Qwen3.5), the 15 `mtp.*` tensors stay unused pending native MTP drafting, and group-6 GQA keeps the eager decode path.

sysuTobo:
## Summary

Qwen3.8-27B rides the existing `qwen35` line — no new model crate, no new cargo feature, no kernel change. The checkpoint is architecturally identical to Qwen3.5-27B (same `model_type: qwen3_5`, zero shape deltas across all 1,199 tensors), so the probe-based dispatch already claims it; four concrete gaps stopped it from loading, fitting, and being verifiable. Each fix commit closes one, and the test commits hold the result against Hugging Face references.

Fixes #1067. The shape-assertion commit also lands one slice of #1068; that issue stays open for the config unknown-field policy and the 1D length checks.

## Invariants being changed

- A Qwen3.5-class checkpoint whose GDN scalars are stored as bf16 must load and must reach the kernels as the same f32 values the reference implementation uses (bf16 → f32 is exact; the strict f32 path stays strict for every other consumer).
- The unsharded load path must reject a checkpoint whose tensors disagree with its `config.json` at the loader, not inside a config-sized GEMM.
- The TP executor must size its recurrent-state pool at the caller's requested decode capacity, not at the 64-slot default.
- Logits goldens must be keyed so a same-geometry generation cannot alias another checkpoint's oracle.

## Commits

- `chore`: ignore the agent's local `.qoder` state
- `fix(qwen35)`: read the GDN scalars from either f32 or bf16 storage
- `fix(qwen35)`: assert config-derived shapes on the unsharded load path
- `fix(qwen35)`: honour the caller's decode capacity in the TP executor
- `test(qwen35)`: key logits goldens by generation as well as geometry — committed Qwen3.8-27B short/long fixtures, dumper refuses non-finite references
- `test(frontend)`: gate Qwen3.8 chat renders against the HF reference — parity machinery shared with gemma4 via `tests/common`
- `test(frontend)`: use the repo-relative model path in the chat dumper example
- `docs(qwen35)`: record serving Qwen3.8-27B — design record, measured budget, follow-ups

## Evidence

2× L20 46 GB (sm_89), TP2, `Qwen/Qwen3.8-27B` @ `1d4bf0f2ff6012fd82039f2fa52739d0dd7c60c0`:

- `hf_golden_gate` short: **2 passed** — sequential eager (108 positions) mean **0.0238** / p99 **0.0862** / max 0.1593; batched eager mean 0.0231 / p99 0.0823; no argmax violation. Tolerances are the line's existing 4B calibration (0.06 / 0.20) — no new constants.
- `hf_golden_gate` long: **1 passed** — 4,097 + 8,192-token prompts, mean 0.0226 / p99 0.0860 / max 0.0882.
- `qwen38_chat_template_parity`: **10/10** renders byte-identical to the checkpoint tokenizer's (plain, multi-turn, unicode, `reasoning_effort` xhigh|medium|low, `enable_thinking=false`, disabled-with-effort, `preserve_thinking`).
- `pegainfer-core --lib` +2 widening tests; `pegainfer-qwen35 --lib` 103 passed; `cargo clippy --all-targets -D warnings` and `cargo fmt --all --check` clean on the three touched crates.
- Qwen3.5 unchanged: same fixtures, same probe outcome, group-6 GQA stays on the eager decode path.

Design record with the full checkpoint facts, the `output_gate_type: "swish"` probe (it is inert — do not implement it), and the two oracle traps: `docs/models/qwen35/support-qwen38.md` (added in this PR).

## Out of scope

- The 333 `model.visual.*` tensors stay unloaded (as on Qwen3.5); the 15 `mtp.*` tensors stay unused — native MTP drafting is a follow-up.
- Group-6 batched-decode kernels (the CUDA-graph variant for 24Q/4KV) stay follow-up; decode runs eager.
- The rest of #1068: config unknown-field policy, 1D length checks.

## Test plan

- [x] TP2 logits gates, short + long, against the committed fixtures
- [x] Chat template parity, 10 cases
- [x] `cargo test --release --workspace --lib` on touched crates, clippy `-D warnings`, `cargo fmt --all --check`
- [ ] Reviewer re-run: `PEGAINFER_TEST_MODEL_PATH=models/Qwen3.8-27B PEGAINFER_TEST_MODEL_REVISION=1d4bf0f2ff6012fd82039f2fa52739d0dd7c60c0 cargo test --release -p pegainfer-qwen35 --features qwen35 --test hf_golden_gate` (the revision env is mandatory on hand-downloaded checkpoints — the gate otherwise skips silently)


xiaguan (changes requested):
Direction and evidence look good. A few changes patch call sites instead of fixing the layer below — please tighten:

1. **Shape checks belong in the loader.** Instead of `tensor_2d` + `check_matrix` at four call sites (plus a separate `ensure!` for embed), make `load_tensor_2d` take the expected `(rows, cols)` and check the header before upload. The 1D loaders from #1068 can follow the same shape.
2. **Drop the parallel widen API.** bf16→f32 is lossless, so strictness buys nothing there. Let `tensor_f32_cow` accept 1D bf16 and remove `*_widened` / `gdn_scalar_f32`.
3. **Fixture lookup by `config_sha256`.** Detecting the generation from `output_gate_type` (the field this PR proves inert) plus the size table and `COMMITTED_FIXTURES` list is three things to keep in sync. Fixtures already record `config_sha256`, so match on that.
4. One generic chat-golden dumper with per-model case tables, not a copy of the gemma4 one.
5. Remove the `MAX_BATCH`-defaulting constructor so capacity is always explicit.
6. Trim comments and the support doc to the essentials. Drop the unrelated `CLAUDE.md` edits (keep the model-table row) and `.qoder`.

docs/models/qwen35/support-qwen38.md
```diff
@@ -0,0 +1,257 @@
...
+  figure predates this checkpoint; it is ~26.9 GB/rank, not 17.5).
+- GQA group is `24/4 = 6`, outside `SUPPORTED_GQA_GROUP_SIZES`, so 27B decode
+  stays on the batched eager path — independent of the generation.
+- Full-attention KV is 16 layers × 2 × 4 heads × 256 × 2 B = **64 KiB/token**; a
+  single 262,144-token sequence reserves 16 GiB before any GDN recurrent state.
+- The 15 `mtp.*` tensors (both generations ship `mtp_num_hidden_layers: 1`) are
+  loaded by nobody. Native-MTP speculation is the obvious free lunch on this line;
+  precedent exists in `glm52` (`GLM52_MTP_DRAFTS`) and qwen3 (DFlash/DSpark).
+
+## Verification
```
FeathBow: After the rebase, could this table say which head the numbers were taken on, and add single-GPU Qwen3.5 `hf_golden_gate` rows (say 2B tied and 4B untied)? The shape checks only run on the single-GPU path and every run here is TP2, so they haven't met a real checkpoint yet. Once they move into `load_tensor_2d` they run on every load, and those rows are also what "Qwen3.5 behaviour is unchanged" can point to.

pegainfer-core/src/weight_loader.rs
```diff
@@ -436,6 +436,30 @@ fn tensor_f32_cow<'d>(
     }
 }
 
+/// 1D f32 payload that additionally accepts a bf16-stored vector and widens it.
+///
+/// Qwen3.8 ships the gated-DeltaNet scalars (`linear_attn.A_log`,
```
FeathBow: Small one, and it mostly goes away when this folds into `tensor_f32_cow`. The "Qwen3.8 stores these as bf16" note is repeated in `gdn_scalar_f32` and both field docs in `layers.rs`. One place is enough.

pegainfer-core/src/weight_loader.rs
```diff
@@ -1093,6 +1154,46 @@ mod tests {
         assert!(tensor_f32_cow(&f32_2d_view, "w").is_err());
     }
 
+    #[test]
+    fn tensor_f32_widen_cow_widens_bf16_without_losing_the_stored_value() {
```
FeathBow: I'd drop both of these. The first mostly checks that bf16 to f32 is exact, which is the `half` crate's behaviour. Once the widening folds into `tensor_f32_cow`, the existing `tensor_f32_cow_rejects_wrong_dtype_and_rank` needs its bf16 assertion flipped anyway, and the Qwen3.8 golden gate already loads the real bf16 scalars, so adjusting that one test covers it

pegainfer-frontend/tests/common/mod.rs
```diff
@@ -0,0 +1,199 @@
...
+fn chat_role(name: &str) -> ChatRole {
+    match name {
+        "system" => ChatRole::System,
+        "user" => ChatRole::User,
+        "assistant" => ChatRole::Assistant,
+        other => panic!("unexpected role {other}"),
+    }
+}
+
+fn reasoning_effort(name: &str) -> ReasoningEffort {
```
FeathBow: Only `xhigh`, `medium` and `low` appear in any golden? I'd map just those and let anything else hit the panic, which already says to extend the match.

pegainfer-frontend/tests/common/mod.rs
```diff
@@ -0,0 +1,199 @@
...
+            template_kwargs: template_kwargs.into_iter().collect(),
+            ..ChatOptions::default()
+        },
+        ..ChatRequest::for_test()
+    }
+}
+
+/// Render every case in `golden["chat_templates"]` and collect the disagreements
+/// so one run reports every divergence, not just the first.
+pub(crate) fn render_mismatches<R>(golden: &Value, render: R) -> Vec<String>
```
FeathBow: `render_mismatches` has one caller, so the closure parameter isn't buying anything yet. It could be folded into `string_form_mismatches`.

pegainfer-frontend/tests/qwen38_chat_template_parity.rs
```diff
@@ -0,0 +1,50 @@
...
+    common::golden(
+        "PEGAINFER_QWEN38_CHAT_GOLDEN",
+        concat!(
+            env!("CARGO_MANIFEST_DIR"),
+            "/../test_data/qwen38-chat-golden.json"
+        ),
+    )
+}
+
+fn model_path() -> String {
```
FeathBow: `model_path()` is the same as the one in `gemma4_tokenizer_parity.rs`. It could move to `common` with the rest.

pegainfer-qwen35/src/tp_executor.rs
```diff
@@ -718,14 +396,20 @@ impl Qwen35TpExecutor {
...
         let mut models = Vec::with_capacity(world_size);
         for (rank, &device_ordinal) in device_ordinals.iter().enumerate() {
-            models.push(Qwen35Model::from_safetensors_with_runtime(
+            // Load at the capacity the caller asked for. `…_with_runtime` defaults
+            // to `MAX_BATCH` decode slots, so going through it here reserved for
+            // 64 slots no matter what `--max-batch` said — the rank-local
+            // recurrent-state pool is sized at load time, so a later admission
+            // shrink could not give that memory back and 27B TP2 could not fit on
+            // a card that could not spare 64 slots' worth of it.
+            models.push(Qwen35Model::from_safetensors_with_runtime_and_capacity(
```
FeathBow: The fix is right. Once the defaulting constructor is gone this comment can be one line, or none, since passing `max_batch` explains itself.

pegainfer-qwen35/src/weights/layers.rs
```diff
@@ -462,4 +505,68 @@ mod tests {
...
+    "eos_token_id": 248044
+  }
+}"#,
+        )
+        .unwrap();
+        Config35::from_file(dir.path().to_str().unwrap()).expect("27b fixture validates")
+    }
+
+    #[test]
+    fn unsharded_ranges_cover_the_full_checkpoint_tensor_shapes() {
```
FeathBow: I'd drop this test. It asserts hardcoded 27B shapes against range helpers the golden gates already exercise, and it stays green if the shape checks are deleted, so it isn't guarding them.

tools/accuracy/dump_qwen35_hf_golden.py
```diff
@@ -93,11 +115,18 @@ def infer_revision(path: Path) -> str:
     return "unknown"
 
 
-def load_model(model_path: str, dtype: str, device_map: str):
+def load_model(model_path: str, dtype: str, device_map: str, max_memory_gib: int | None):
     kwargs = {"trust_remote_code": True, "torch_dtype": DTYPES[dtype]}
     if device_map == "none":
         model = AutoModelForCausalLM.from_pretrained(model_path, **kwargs).to("cuda")
     else:
+        if max_memory_gib is not None:
```
FeathBow: `--max-memory-gib` is there for one shared machine and isn't needed for Qwen3.8. I'd keep it out of this PR.

tools/accuracy/dump_qwen38_chat_golden.py
```diff
@@ -0,0 +1,128 @@
...
+    ("plain_user_turn", MESSAGES, True, {}),
+    ("plain_user_no_generation_prompt", MESSAGES, False, {}),
+    ("multi_turn", MULTI_TURN, True, {}),
+    ("unicode_content", UNICODE, True, {}),
+    ("reasoning_effort_xhigh", MESSAGES, True, {"reasoning_effort": "xhigh"}),
+    ("reasoning_effort_medium", MESSAGES, True, {"reasoning_effort": "medium"}),
+    ("reasoning_effort_low", MESSAGES, True, {"reasoning_effort": "low"}),
+    ("thinking_disabled", MESSAGES, True, {"enable_thinking": False}),
+    ("thinking_disabled_with_effort", MESSAGES, True, {"enable_thinking": False, "reasoning_effort": "low"}),
+    ("preserve_thinking", MULTI_TURN, True, {"preserve_thinking": True}),
```
FeathBow: This passes the template default against `MULTI_TURN`, whose assistant turn has no reasoning, so it renders exactly like `multi_turn`. I'd drop it rather than extend it. The richer template paths (tool-argument `tojson`, carried reasoning) run through the vendored renderer, which has its own tests for the HF-compatible `tojson`, so more cases here would mostly be testing upstream.

FeathBow (commented):
Some small things.

sysuTobo:
All six items are addressed in the commits on top of the branch:

1. **Shape checks in the loader** — `load_tensor_2d` now takes the expected `(rows, cols)` and rejects a mismatching header before upload. All callers migrated: qwen35 `layers.rs`/`weights.rs` (`check_matrix` and the embed/lm_head `ensure!`s deleted), deepseek-v2-lite (`load_dense_mlp` now threads hidden/intermediate through its three call sites), qwen3 dflash/eagle3, glm52 dspark and k3 dspark (`ensure_matrix`/`check2d` deleted).
2. **No parallel widen API** — `tensor_f32_cow` accepts 1D bf16 and widens it; `load_tensor_1d_f32_widened`, `load_tensor_1d_f32_shard_widened` and `WeightSource::gdn_scalar_f32*` are removed. Other dtypes and ranks stay rejected, pinned by the renamed cow tests.
3. **Fixture lookup by `config_sha256`** — the gate hashes the local `config.json` and keeps the committed fixture whose safetensors metadata records the same hash (exactly one match required); `output_gate_type` probing, the geometry table and `COMMITTED_FIXTURES` are deleted, and the dumper's `SIZE_NAMES` only names its default output. `check_fixture_metadata` still re-asserts hash + `model_revision` before comparing logits.
4. **One generic chat dumper** — `tools/accuracy/dump_chat_template_golden.py` with per-model case tables (`gemma4`, `qwen38`) replaces the qwen38 copy; `template_kwargs` is recorded so both committed goldens regenerate byte-identical.
5. **No `MAX_BATCH`-defaulting constructor** — `from_safetensors_with_runtime` is gone; `from_safetensors_with_options` now takes `max_batch`, so capacity is explicit at every constructor.
6. **Trimmed** — the support doc is cut to the essentials and updated for the above; `CLAUDE.md` keeps only the model-table row; the `.qoder` gitignore entry is reverted.

Verified on the cluster since the last round (details and per-head table in `docs/models/qwen35/support-qwen38.md`): Qwen3.5 single-GPU `hf_golden_gate` rows for 0.8B and 2B (tied) and 4B (untied) on the merged tree, all inside the line's tolerances; `pegainfer-qwen35 --lib` 102 passed on an L20 with the Triton AOT build; `pegainfer-core --lib` 38 passed; chat parity 9/9; clippy/fmt clean. The branch also carries a merge of `main` (vllm rust crates → 15ed1262) with the frontend test helpers following the new `EffortValue` API, and a signoff repair so the DCO job sees every commit signed.

sysuTobo:
All ten inline items addressed:

1. **`support-qwen38.md` verification table** — every row now names the head and device it ran at, and the table gains single-GPU Qwen3.5 `hf_golden_gate` rows on the merged tree (1× A40, sm_80 build): 0.8B and 2B (tied) and 4B (untied), all inside the existing tolerances — 4B short sequential mean 0.0238 / p99 0.0813, 2B long mean 0.0238 / p99 0.0778, 0.8B short mean 0.0298 / p99 0.1137. Those are what "Qwen3.5 behaviour is unchanged" points to now that `load_tensor_2d` asserts every tensor's shape on every load.
2. **Repeated bf16 note** — `gdn_scalar_f32` is gone (widening folded into `tensor_f32_cow`); the field docs keep a single "(widened from bf16 storage)" mention.
3. **Widen tests** — both dropped. `tensor_f32_cow_rejects_wrong_dtype_and_rank` now asserts the flip: 1D bf16 is accepted, f16/i64/2D stay rejected.
4. **`reasoning_effort` match** — maps only `xhigh`/`medium`/`low`; everything else hits the panic that says to extend it. (After the merge to the new vllm-chat API the mapping returns `EffortValue::String`; same three names, same panic.)
5. **`render_mismatches`** — folded into its only caller, `string_form_mismatches`.
6. **`model_path()`** — moved to `tests/common` (checkpoint family as a parameter); both parity tests use it.
7. **TP executor comment** — deleted; passing `max_batch` explains itself.
8. **Hardcoded 27B shape test** — deleted with its `config_27b` helper.
9. **`--max-memory-gib`** — removed from the dumper and the doc commands.
10. **`preserve_thinking` case** — confirmed byte-identical to `multi_turn` in the committed golden; dropped from the case table and the golden (10 → 9 cases).

Verified after the changes: `qwen38_chat_template_parity` 1/1 (9 cases) against the pinned checkpoint tokenizer; the exact CPU-Clippy job command clean locally; `pegainfer-core --lib` 38/38; `pegainfer-qwen35 --lib` 102/102 on an L20 (Triton AOT build); the single-GPU gate rows above.

FeathBow (commented):
`qwen38_chat_template_parity.rs`: if this stays, `reasoning_effort_xhigh` renders the same as `plain_user_turn` (xhigh is the default), and `thinking_disabled_with_effort` the same as `thinking_disabled`. I'd drop one from each pair.

`qwen38_chat_template_parity.rs` / `index.md`: both mention `preserve_thinking`, but the golden has no case for it.

`dump_chat_template_golden.py`: `ALWAYS_RECORD_KWARGS` preserves a fixture this PR introduces, so it can go.

`common/mod.rs`: `#![allow(dead_code)]` isn't needed, since both binaries use every helper.

`layers.rs`: `geometry_for` was split out for the dropped 27B test and only has one caller now.

`weight_loader.rs`: `tensor_f32_cow_rejects_*` now starts by asserting bf16 is accepted, so it's worth a rename. `I64` and `F16` hit the same check, so one is enough.

`hf_golden_gate.rs`: a missing committed fixture used to panic, but now it skips and still reports ok. The same happens when `config.json` differs by a byte. I'd suggest panicking on zero matches.

